PR #181, opened 2020-07-15
feature (#169): add migration badges

### LINKED ISSUE #169: Add appropriate badges to migration sections
Migration sections now have the ability to show badges in their headings which we can use as metadata. This issue is to add badges to all of the sections for consistency

bencodezen:
[Netlify Preview](https://deploy-preview-181--vue-next.netlify.app/guide/migration/v-model.html)

Still not sure about the full list of what we want to include and how to tag things appropriately. 

Maybe the badges should only be used next to the bulleted items in each overview section? That way it doesn't seem like the entire thing is broken or that an entire feature is removed?

Also, styles are totally up for feedback / changes. Just iterating along the way!

Closes #169


phanan:
The Netlify preview is gone for some reason. @bencodezen maybe a rebuild?

sdras:
Looks like there are some merge conflicts and then we're good to go, @bencodezen

PR #214, opened 2020-07-18
fix(reactivity): typo

yyyang1996:
Fix a typo

typo2


yyyang1996:
> Thanks for the catch on the `key`, `prop` typo! However, we want to keep the spread operator. Can you please remove those changes? Then we can merge.

Sure. It's done.

sdras:
Thank you!

PR #224, opened 2020-07-18
fix: use current directives API

visualfanatic:
Directives API was changed a while ago: the function was renamed, it no longer requires to provide a current instance and directives should be passed in an array as opposed to being passed as function arguments.


bencodezen:
@NataliaTepluhina Do we have this function documented somewhere in the API? Can't seem to find it, but if not, happy to create an issue for it.

NataliaTepluhina:
@bencodezen yes, we should add it to the API reference too

PR #235, opened 2020-07-18
Update plugins section to guide style with example

marina-mosti:
Update the `plugins` section to a step-by-step guide approach to writing plugins, and added some tips and extra information for `using`

## Note

### DIFF
--- a/src/guide/plugins.md
+++ b/src/guide/plugins.md
@@ -12,41 +12,70 @@ Plugins usually add global-level functionality to Vue. There is no strictly defi
 
 5. A library that provides an API of its own, while at the same time injecting some combination of the above (e.g. [vue-router](https://github.com/vuejs/vue-router)).
 
-## Using a Plugin
-
-After a Vue app has been initialized with `createApp()`, you can add a plugin to your application by calling the `use()` method:
-
-```js
-// calls `MyPlugin.install(Vue)`
-const app = Vue.createApp({})
-app.use(MyPlugin)
-```
-
-If you need to define configuration for a plugin, you can pass a config object as the second argument:
-
-```js
-app.use(MyPlugin, { someOption: true })
-```
-
-`app.use()` automatically prevents you from using the same plugin more than once, so calling it multiple times on the same plugin will install the plugin only once.
-
-Checkout [awesome-vue](https://github.com/vuejs/awesome-vue#components--libraries) for a huge collection of community-contributed plugins and libraries.
-
 ## Writing a Plugin
 
-A Vue.js plugin should expose an `install()` method. The method will be called with the application instance as the first argument, along with possible options:
+In order to better understand how to create your own Vue.js plugins, we will create a very simplified version of a plugin that displays `i18n` ready strings. 
+
+A Vue.js plugin is either an `object` that should expose an `install()` method, or a `function`. 
+
+Whenever this plugin is added to an application, the `install` method will be called if it is an object. If it is a `function`, the function itself will be called. In both cases, it will receive two parameters - the `app` object resulting from Vue's `createApp`, and the options passed in by the user.
+
+Let's begin by setting up the plugin object.
 
 ```js
-MyPlugin.install = (app, options) => {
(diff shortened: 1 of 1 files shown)


src/guide/plugins.md
```diff
@@ -12,49 +12,77 @@ Plugins usually add global-level functionality to Vue. There is no strictly defi
 
 5. A library that provides an API of its own, while at the same time injecting some combination of the above (e.g. [vue-router](https://github.com/vuejs/vue-router)).
 
-## Using a Plugin
+## Writing a Plugin
+
+In order to better understand how to create your own Vue.js plugins, we will create a very simplified version of a plugin that displays `i18n` ready strings. 
 
-After a Vue app has been initialized with `createApp()`, you can add a plugin to your application by calling the `use()` method:
+A Vue.js plugin is either an `object` that should expose an `install()` method, or a `function`.
```
bencodezen, line 19:
```suggestion
```

I could recommend moving this to the top where we define what a Plugin is.

```
A Vue.js plugin usually adds global-level functionality to Vue. It is either an `object` that should expose an `install()` method, or a `function`. There is no strictly defined scope for a plugin, but common scenarios where plugins are useful include:

File: src/guide/plugins.md [hunk 2]
```diff
@@ -12,49 +12,77 @@ Plugins usually add global-level functionality to Vue. There is no strictly defi
...
+In order to better understand how to create your own Vue.js plugins, we will create a very simplified version of a plugin that displays `i18n` ready strings. 
 
-After a Vue app has been initialized with `createApp()`, you can add a plugin to your application by calling the `use()` method:
+A Vue.js plugin is either an `object` that should expose an `install()` method, or a `function`. 
+
+Whenever this plugin is added to an application, the `install` method will be called if it is an object. If it is a `function`, the function itself will be called. In both cases, it will receive two parameters - the `app` object resulting from Vue's `createApp`, and the options passed in by the user.
+
+Let's begin by setting up the plugin object.
 
 ```js
```
bencodezen, line 25: Where would this hypothetical file live?

```suggestion
```js
// /User/i18nPlugin/src/index.js
```

src/guide/plugins.md
```diff
@@ -63,3 +91,36 @@ MyPlugin.install = (app, options) => {
...
+  greetings: {
+    hi: 'Hallo!'
+  }
+}
+
+app.use(i18nPlugin, i18nStrings)
+app.mount('#app')
+```
+
+`app.use()` automatically prevents you from using the same plugin more than once, so calling it multiple times on the same plugin will install the plugin only once.
```
bencodezen, line 124:
```suggestion
```

This could probably go higher in this section since it's a generic piece of advice prior to getting into specifics.

src/guide/plugins.md
```diff
@@ -63,3 +91,36 @@ MyPlugin.install = (app, options) => {
   })
 }
 ```
+
+## Using a Plugin
+
+After a Vue app has been initialized with `createApp()`, you can add a plugin to your application by calling the `use()` method. 
+
```
bencodezen, line 98:
```suggestion
::: info
`app.use()` automatically prevents you from using the same plugin more than once, so calling it multiple times on the same plugin will install the plugin only once.
:::
```

src/guide/plugins.md
marina-mosti: Agreed, going to add a bit more to the description though

src/guide/plugins.md
marina-mosti: The way the documentation was previously written implied that the code was being written directly on top of what looks like a `main.js` file. Do we want to incorporate some instructions on how to make it into its own file and import it? I can add that

marina-mosti:
Thanks for the review @bencodezen , I've applied the requested changes. It does look better with the file living in its own separate space, I was just afraid that it would introduce a foreign concept due to the way it was phrased earlier. But I also agree that probably it would be a best practice to place plugins in a separate folder/file anyway

src/guide/plugins.md
bencodezen: @marina-mosti Definitely. I think it would be super helpful for those who have no context as to where to start.

PR #236, opened 2020-07-18
Create a wider container for the codepens

sdras:
Working on the responsive styles for the explainer codepens but I think they will be more legible with a wider container.


sdras:
Yep exactly, otherwise it looks broken in the meantime. Thanks!

Issue #240, opened 2020-07-18
Wrong link on Ref page (localhost:8080)


edimitchel:
Going on https://v3.vuejs.org/api/refs-api.html#customref, the last link point to a localhost domain.
May be, elsewhere this typo exists..

Here the fix : https://github.com/edimitchel/docs-next/commit/885282f202238dcf79131b4588aa3233ecc996df

bencodezen:
Good catch! Do you mind submitting a PR @edimitchel? Would love to give you credit for finding the issue and solving it!

edimitchel:
Yes, I can do it now ! I create this issue because the PR template invites to do it ;)

NataliaTepluhina:
Closing an issue as related PR was merged

PR #241, opened 2020-07-19
Fix anchors which refers to localhost:8080 domain

edimitchel:
@bencodezen

#240


edimitchel:
Links are not working as expected in Ref. I fix it !

sdras:
Nice catch, thank you!

PR #251, opened 2020-07-19
chore: allow discarding beta banner

phanan:
As discussed, this PR adds the ability for the user to discard the "docs current in beta" banner. I've also played a bit with the HTML to make the message fit on devices with smaller screens.

### DIFF
--- a/src/.vuepress/components/BetaBanner.vue
+++ b/src/.vuepress/components/BetaBanner.vue
@@ -1,31 +1,34 @@
 <template>
-  <aside class="beta-banner" v-if="shouldShow">
-    <p>
-      <span class="hide-sm">⚠️ Beta Version:</span>
-      Docs
-      <span class="hide-sm">are</span> in development and subject to change.
-    </p>
-    <button ref="closeButton" @click="close">Close</button>
-  </aside>
+  <ClientOnly>
+    <aside class="beta-banner" v-if="shouldShow">
+      <p>
+        <span class="hide-sm">⚠️ Beta Version:</span>
+        Docs
+        <span class="hide-sm">are</span> in development and subject to change.
+      </p>
+      <button ref="closeButton" @click="close">Close</button>
+    </aside>
+  </ClientOnly>
 </template>
 
 <script>
 export default {
   computed: {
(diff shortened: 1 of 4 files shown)


NataliaTepluhina:
@phanan Netlify build produces an error:

```
ReferenceError: localStorage is not defined
```

phanan:
Oh! I forgot about the client/server part!

phanan:
@NataliaTepluhina Maybe another round?

phanan:
Yeah, I requested reviews from Ben & Sarah as well.

phanan:
@CyberAP Since this is a temporary future, I'd like to keep the component as self-contained as possible.

CyberAP:
> @CyberAP Since this is a temporary feature, I'd like to keep the component as self-contained as possible.

You can achieve the same thing if you put `v-if` on a root element of `<BetaBanner>` component.

phanan:
@CyberAP I think I know I can achieve the same thing with other approaches, thank you :) This is the approach I've chosen though, and I don't think it's worth it to do otherwise.

src/.vuepress/components/BetaBanner.vue
```diff
@@ -1,9 +1,41 @@
...
+      document.querySelector('html').classList.remove('with-beta-banner')
+    }
+  },
+
+  beforeMount() {
+    if (this.shouldShow) {
+      document.querySelector('html').classList.add('with-beta-banner')
+    }
+  }
+}
```
CyberAP, line 36:
```suggestion
  data() {
    return {
      shouldShow: typeof window === 'undefined' ?
        true :
        window.localStorage.getItem('beta-banner-discarded') === null
    }
  },

  methods: {
    close() {
      window.localStorage.setItem('beta-banner-discarded', true)
      this.shouldShow = false
      document.querySelector('html').classList.remove('with-beta-banner')
    }
  },

  beforeMount() {
    if (this.shouldShow) {
      document.querySelector('html').classList.add('with-beta-banner')
    }
  }
}
```

sdras:
Thank you An! Looks great!

PR #255, opened 2020-07-19
add resolveComponent, resolveDirective, createRenderer and withDirectives to global API

marina-mosti:


src/api/global-api.md
```diff
@@ -193,6 +193,100 @@ const AsyncComp = defineAsyncComponent({
...
+})
+```
+
+```js
+import { resolveComponent } from 'vue'
+setup() {
+  const MyComponent = resolveComponent('MyComponent')
+}
+```
+
```
NataliaTepluhina, line 219: **question**: could we add arguments here if possible just like we did on directives?

```suggestion
#### Arguments

- `{ String } name
```

src/api/global-api.md
```diff
@@ -193,6 +193,100 @@ const AsyncComp = defineAsyncComponent({
...
+  [foo, this.x],
+  [bar, this.y]
+])
+```
+
+#### Arguments
+
+Accepts two arguments: `vnode` and `directives`.
+
+##### vnode
```
NataliaTepluhina, line 269: **nitpick**: I'm afraid it will be _too_ nested 😢  Could we follow the same format as for previous arguments?

```
- `{ VNode } vnode` - a virtual node, usually created with `h()`
- `{ Array  }directives ` - an array of directives. Each directive itself is an array, which allows for up to 4 indexes to be defined as follows
...
```

marina-mosti: I made a mistake with the indentation actually, thanks for pointing it out.

marina-mosti:
Thanks for the review @NataliaTepluhina 
I opted for adding a few more code examples to the `withDirectives` arg `directives` since it didn't seem to be too clear now that I took a second look at it.

Indentation is looking homogenous:

image

src/api/global-api.md
```diff
@@ -193,6 +193,195 @@ const AsyncComp = defineAsyncComponent({
...
+- **Type:** `String`
+
+- **Details:**
+
+  The name of a loaded directive.
+
+## withDirectives
+
+:::warning
+`withDirectives` can only be used within render functions.
```
bencodezen, line 269: Love these warnings!

PR #261, opened 2020-07-19
Update attribute coercion with respect to new behaviour

CyberAP:
Only `null` and `undefined` will result in no attribute in the final DOM, according to [RFC #24](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0024-attribute-coercion-behavior.md#detailed-design)

## Note

### DIFF
--- a/src/guide/template-syntax.md
+++ b/src/guide/template-syntax.md
@@ -60,7 +60,7 @@ In the case of boolean attributes, where their mere existence implies `true`, `v
 <button v-bind:disabled="isButtonDisabled">Button</button>
 ```
 
-If `isButtonDisabled` has the value of `null` or `undefined` attribute will not even be included in the rendered `<button>` element.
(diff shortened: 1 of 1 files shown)


src/guide/template-syntax.md
```diff
@@ -60,7 +60,7 @@ In the case of boolean attributes, where their mere existence implies `true`, `v
 <button v-bind:disabled="isButtonDisabled">Button</button>
 ```
 
-If `isButtonDisabled` has the value of `null`, `undefined`, or `false`, the `disabled` attribute will not even be included in the rendered `<button>` element.
+If `isButtonDisabled` has the value of `null` or `undefined` attribute will not even be included in the rendered `<button>` element.
```
NataliaTepluhina, line 63: **suggestion**: it still has to have an attribute name
```suggestion
If `isButtonDisabled` has the value of `null` or `undefined`, the `disabled` attribute will not even be included in the rendered `<button>` element.
```

NataliaTepluhina:
Thanks! Approved now 👍🏻

PR #264, opened 2020-07-19
Link to Vue reactivity on Instance page

CyberAP:
This is the first mention of Reactivity in Vue Guide, so it might be worthwhile to link to the explanation page.

## Note


sdras:
Looks good, thank you!

PR #265, opened 2020-07-19
Link to Reactivity in depth on Computed page

CyberAP:
It is not yet clear why `Date.now()` is not reactive at the current stage of learning progression, a sidenote with a link to Reactivity in Depth should resolve that issue.

## Note

### DIFF
--- a/src/guide/computed.md
+++ b/src/guide/computed.md
@@ -93,6 +93,10 @@ computed: {
 }
 ```
 
+::: tip 
+To learn why `Date.now()` is not reactive please refer to [**Reactivity in Depth**](reactivity.html).
+:::
+
 In comparison, a method invocation will **always** run the function whenever a re-render happens.
 
 Why do we need caching? Imagine we have an expensive computed property `list`, which requires looping through a huge array and doing a lot of computations. Then we may have other computed properties that in turn depend on `list`. Without caching, we would be executing `list`’s getter many more times than necessary! In cases where you do not want caching, use a `method` instead.


sdras:
Sorry, we don't cover absolutely all use cases in the Guide, it breaks up the flow of reading. Tips are meant to be used sparingly as well, you can see the Writing Guide for more detail as to why.

CyberAP:
I understand the best intent here, but for a novice Vue learner it is unclear why `Date.now()` is not reactive, since it's not mentioned before how a value becomes reactive. It's a very common issue among developers at start to not being able to distinguish between reactive and non-reactive values. I think it is beneficial to answer crucial questions for learners about reactivity, they might not immediately dive into reactivity but at least it would be clear that this is explained later.
What do you think about changing the form in which it is presented to a reader? An info tip might be an overkill here and I agree in that, maybe a short message would suffice instead?

CyberAP:
@sdras could you please reconsider your decision regarding this pull request?

I've found that current documentation actually does have an example of solving a similar problem: https://v3.vuejs.org/guide/custom-directive.html#hook-functions

Here we are introduced with an unknown term `VNode`, a tip explains that it'll be uncovered later on in the docs.

Could we add the same thing here as well? A `reactive dependency` is not a known term at this stage of docs progression and I think it is important to resolve a raising question of ’what is a reactive dependency’?

src/guide/computed.md
```diff
@@ -93,6 +93,10 @@ computed: {
 }
 ```
 
+::: tip 
+To learn why `Date.now()` is not reactive please refer to [**Reactivity in Depth**](reactivity.html).
+:::
+
```
CyberAP, line 99:
```suggestion
::: tip
We'll cover Reactivity and reactive dependencies in more detail [later](reactivity.html).
:::
```

PR #268, opened 2020-07-19
fix: sidebar's top padding on mobile devices

phanan:
Not sure if I'm missing anything here, but currently the sidebar has a weird top padding on mobile:

This PR removes it:


sdras:
Thank you An! This was bugging me 😅

PR #272, opened 2020-07-20
Clarifying SFC and other Simplifications

sdras:
This PR makes some opinionated choices, so feel free to close it. The main change I thought was meaningful is to let people know upfront that we typically use SFC where we use a template string. I moved the two way binding up because I think it's compelling, but not attached to it.


src/guide/introduction.md
```diff
@@ -87,138 +87,138 @@ Vue.createApp(AttributeBindingApp).mount('#bind-attribute')
```
sdras, line 87: From here down, it's just that I swapped two sections around, but it's similar enough that weird diffing is happening

PR #285, opened 2020-07-20
Improve Mixins page

CyberAP:
State that:

1. Mixins are a __simple__ way to reuse component logic, rather than a flexible way to do it. The term __flexible__ goes into conflict with the last chapter of the page, which describes that mixins are in fact inflexible.
2. Add a section title for Precautions.

## Note

### DIFF
--- a/src/guide/mixins.md
+++ b/src/guide/mixins.md
@@ -2,7 +2,7 @@
 
 ## Basics
 
-Mixins are a simple way to distribute reusable functionalities for Vue components. A mixin object can contain any component options. When a component uses a mixin, all options in the mixin will be "mixed" into the component's own options.
+Mixins distribute reusable functionalities for Vue components. A mixin object can contain any component options. When a component uses a mixin, all options in the mixin will be "mixed" into the component's own options.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


src/guide/mixins.md
```diff
@@ -2,7 +2,7 @@
 
 ## Basics
 
-Mixins are a flexible way to distribute reusable functionalities for Vue components. A mixin object can contain any component options. When a component uses a mixin, all options in the mixin will be "mixed" into the component's own options.
+Mixins are a simple way to distribute reusable functionalities for Vue components. A mixin object can contain any component options. When a component uses a mixin, all options in the mixin will be "mixed" into the component's own options.
```
sdras, line 5: This is a good point, however we avoid the use of words like "simple" as they can be hostile to people learning (see our writing guide for more info). It might be worth just taking out the qualifier altogether here.

CyberAP:
```suggestion
Mixins distribute reusable functionalities for Vue components. A mixin object can contain any component options. When a component uses a mixin, all options in the mixin will be "mixed" into the component's own options.
```

What do you think about this change?

PR #286, opened 2020-07-20
Adjust migration section grouping title to better reflect its contents

CyberAP:
**Motivation**

* Pages under this section group reflect changes that happened only between Vue 2 and Vue 3. It does not help if you're migrating from Vue 1 for example.
* Since it is already a Vue 3 documentation calling migration section `Migration to Vue 3` is a bit of a tautology.
* This grouping does not reflect migration from other frameworks like React or Svelte.

## Note

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -108,7 +108,7 @@ const sidebar = {
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


phanan:
LGTM, thanks for the quick catch!

PR #289, opened 2020-07-20
Update computed properties example

### LINKED ISSUE #38: Refactor computed properties chapter
From https://github.com/vuejs/docs-next/pull/31#discussion_r368278101 and https://github.com/vuejs/docs-next/pull/31#discussion_r368278483

1. An example with `reversedMessage` on Computed properties chapter is introducing to complex handling of strings which is out of the scope here. We can add a simpler example something like the sum of two numbers as a computed property.

2. The example about in-template expressions makes sense, but we  introduce computed properties as a mechanism to get derived values. The introduction as a solution to prevent complex expressions in templates does not sound that impressive.

NataliaTepluhina:
Changed an example of `computed` properties to be less cryptic and more 'real-life'

Close #38

### DIFF
--- a/src/guide/computed.md
+++ b/src/guide/computed.md
@@ -13,7 +13,7 @@ Vue.createApp({
         books: [
           'Vue 2 - Advanced Guide',
           'Vue 3 - Basic Guide',
-          'Vue 4 - the Mystery'
+          'Vue 4 - The Mystery'
         ]
       }
     }
--- a/src/guide/computed.md
+++ b/src/guide/computed.md
@@ -52,7 +52,7 @@ Vue.createApp({
         books: [
           'Vue 2 - Advanced Guide',
           'Vue 3 - Basic Guide',
-          'Vue 4 - the Mystery'
+          'Vue 4 - The Mystery'
         ]
       }
     }
--- a/src/guide/computed.md
+++ b/src/guide/computed.md
@@ -12,7 +12,7 @@ Vue.createApp({
         name: 'John Doe',
         books: [
           'Vue 2 - Advanced Guide',
-          'Vue 3 Basic Guide',
+          'Vue 3 - Basic Guide',
           'Vue 4 - the Mystery'
         ]
       }
--- a/src/guide/computed.md
+++ b/src/guide/computed.md
@@ -50,7 +50,7 @@ Vue.createApp({
       author: {
         name: 'John Doe',
         books: [
-          'Vue 2 Advanced guide',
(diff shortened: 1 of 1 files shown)


src/guide/computed.md
```diff
@@ -2,45 +2,69 @@
...
+In-template expressions are very convenient, but they are meant for simple operations. Putting too much logic in your templates can make them bloated and hard to maintain. For example, if we have an object with a nested array:
+
+```js
+Vue.createApp({
+  data() {
+    return {
+      author: {
+        name: 'John Doe',
+        books: [
+          'Vue 2 Advanced guide',
```
bencodezen, line 14:
```suggestion
          'Vue 2 - Advanced Guide',
```

src/guide/computed.md
```diff
@@ -2,45 +2,69 @@
...
+          'Vue 3 Basic Guide',
```
bencodezen, line 15:
```suggestion
          'Vue 3 - Basic Guide',
```

src/guide/computed.md
```diff
@@ -2,45 +2,69 @@
...
+          'Vue 4 - the Mystery'
```
bencodezen, line 16:
```suggestion
          'Vue 4 - The Mystery'
```

src/guide/computed.md
```diff
@@ -2,45 +2,69 @@
...
 ```js
-const vm = Vue.createApp({
+Vue.createApp({
   data() {
     return {
-      message: 'Hello'
+      author: {
+        name: 'John Doe',
+        books: [
+          'Vue 2 Advanced guide',
```
bencodezen, line 53:
```suggestion
          'Vue 2 - Advanced Guide',
```

src/guide/computed.md
```diff
@@ -2,45 +2,69 @@
...
+          'Vue 3 Basic Guide',
```
bencodezen, line 54:
```suggestion
          'Vue 3 - Basic Guide',
```

src/guide/computed.md
```diff
@@ -2,45 +2,69 @@
...
+          'Vue 4 - the Mystery'
```
bencodezen, line 55:
```suggestion
          'Vue 4 - The Mystery'
```

CyberAP:
Would it be reasonable to add an **Add book** functionality to highlight that computed properties are in fact reactive? That could potentially help to understand why book list was put in `data` rather than in a variable outside app.

PR #290, opened 2020-07-20
fix: broken urls due to refactor

bencodezen:


bencodezen:
Gonna merge without approval after verifying redirects work so it doesn't break for other people.

PR #291, opened 2020-07-20
Increase accessibility on SFC page

CyberAP:
**Motivation**

* The code in an image can not copied.
* The code can not be indexed by search engines.
* The code is not accessible to screen readers.
* The code can not be changed without generating a new image.
* Link to codepen and image are tightly coupled.
* These image usage examples are not consistent with the rest of the docs.

## Note


phanan:
Thanks for the PR. The whole point of having the screenshots is to demonstrate how the code would look like in a real editor, with full syntax highlighting for different processors. Your change just defeats that purpose (notice how `template` and `style` lose the highlights).

Nevertheless, as clearly mentioned in the PR template, at this time, while the team continues to iterate on the docs, please provide any feedback via a new issue rather than a PR.

PR #296, opened 2020-07-20
Add examples for v-model API section

CyberAP:
## Note


bencodezen:
Thanks for the PR @CyberAP. At this time, I think the API examples would require more explanation if someone doesn't have the pre-requisite knowledge for what is going on; so for now, I'm going to close this one in favor of pushing people towards the guides which have more in-depth explanations.

PR #297, opened 2020-07-20
Reflect current ref API in special attributes section

### LINKED ISSUE #292: Ref could used with a function
Ref docs : https://v3.vuejs.org/api/special-attributes.html#ref

The attribute is a string in the docs, but we can use a function with as parameter the element ref.

```html
<input 
  :ref="(ref) => inputRef = ref"
/>
```
This is feasable ! May be, this could be explicited in the attribute type ?

CyberAP:
Fix #292

## Note

### DIFF
--- a/src/api/special-attributes.md
+++ b/src/api/special-attributes.md
@@ -33,7 +33,7 @@
 
 ## ref
 
-- **Expects:** `string`
+- **Expects:** `String | Function | Inline statement`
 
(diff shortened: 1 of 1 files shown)


edimitchel:
Thanks @CyberAP  !

NataliaTepluhina:
Closing this PR as it's already fixed on `master`

PR #298, opened 2020-07-20
Add comma separated list of expressions to v-on API section, fix dated example description

CyberAP:
## Note

### DIFF
--- a/src/api/directives.md
+++ b/src/api/directives.md
@@ -155,7 +155,7 @@
 
 - **Shorthand:** `@`
 
-- **Expects:** `Function | Inline Statement | Object`
+- **Expects:** `Function | Inline Statement | Comma separated expressions | Object`
 
 - **Argument:** `event`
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
I think this is already implicit from the existing documentation.

The comma in this case is just a JavaScript comma operator being used to form a complex expression from smaller expressions. You could equally write the same code using any left-to-right operator that doesn't use short-circuiting. e.g. any of the following would work:

```text
@click="a(), b(), c()"
@click="a() + b() + c()"
@click="a() | b() | c()"
@click="a() & b() & c()"
@click="a() > b() > c()"
```

You could get a similar effect using an array:

```text
@click="[a(), b(), c()]"
```

While some of these might look like special Vue syntax for events, they're all just exploiting JavaScript language features to construct complex expressions.

In reality you can even use multiple statements:

```text
@click="a(); b(); c()"
```

I don't know if that's something we would want to encourage in the documentation though.

skirtles-code:
The 'dated example description' has now been fixed on `master`.

Given my previous comment about the commas, I'm going to close this. It can be re-opened if necessary but currently I don't see any reason to document this.

PR #300, opened 2020-07-20
Update built-in meta component API section

CyberAP:
* Add more examples
* Clarify what `is` can actually accept as its value
* Remove redundant types (ComponentConstructor is not described anywhere)

## Note


CyberAP:
It seems that component `is` also accepts VNodes, will test it further.

PR #302, opened 2020-07-20
fix: use current event modifiers syntax in render function

visualfanatic:
To make `.passive`, `.capture` and `.once` event modifiers JSX-friendly, the syntax was recently changed to allow passing them in the camelCase notation right after the event name. I added the note about combining them and extended the example to include the `.passive` event modifier just in case.

### DIFF
--- a/src/guide/render-function.md
+++ b/src/guide/render-function.md
@@ -295,27 +295,32 @@ render() {
 
 #### Event Modifiers
 
-For the `.passive`, `.capture`, and `.once` event modifiers, Vue offers object syntax of the handler:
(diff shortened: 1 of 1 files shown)


skirtles-code:
@visualfanatic It looks like the changes you were proposing are quite similar to #552, which has now been merged.

Given the current conflicts, I think it would be best to close this PR. Do you agree?

I can't speak for anyone else but I think it would be remiss of me to suggest closing this PR without also acknowledging the long delay in giving you any feedback. I dare say that's quite frustrating for you. I don't know why that happened but my guess would be that there was a good reason why it couldn't be merged immediately and then it got forgotten about. My own feeling is that this is more likely to happen when there are a lot of open PRs, so I'm trying to help out by reviewing all the old PRs.

PR #304, opened 2020-07-20
fix: use Vue component instead of `input` in v-model section

visualfanatic:
Using `v-model` with `input` is still the same as in Vue 2, this was changed only when using with Vue components.

### DIFF
--- a/src/guide/render-function.md
+++ b/src/guide/render-function.md
@@ -276,7 +276,7 @@ props: ['modelValue'],
 render() {
   return Vue.h(SomeComponent, {
     modelValue: this.modelValue,
-    'onUpdate:modelValue': value => this.$emit('onUpdate:modelValue', value)
+    'onUpdate:modelValue': value => this.$emit('update:modelValue', value)
   })
 }
 ```
--- a/src/guide/render-function.md
+++ b/src/guide/render-function.md
@@ -274,7 +274,7 @@ The `v-model` directive is expanded to `modelValue` and `onUpdate:modelValue` pr
 ```js
 props: ['modelValue'],
 render() {
-  return Vue.h('input', {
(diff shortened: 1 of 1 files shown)


src/guide/render-function.md
```diff
@@ -274,9 +274,9 @@ The `v-model` directive is expanded to `modelValue` and `onUpdate:modelValue` pr
 ```js
 props: ['modelValue'],
 render() {
-  return Vue.h('input', {
+  return Vue.h(SomeComponent, {
```
znck, line 277: I don't understand the need for this change.

src/guide/render-function.md
```diff
@@ -274,9 +274,9 @@ The `v-model` directive is expanded to `modelValue` and `onUpdate:modelValue` pr
...
     modelValue: this.modelValue,
-    'onUpdate:modelValue': value => this.$emit('onUpdate:modelValue', value)
+    'onUpdate:modelValue': value => this.$emit('update:modelValue', value)
```
znck, line 279: 😆 I stumbled on this yesterday, API docs saved me.

src/guide/render-function.md
znck: AFIK `modelValue` is handled in runtime and it is same for components and elements.

visualfanatic: I will try to expand a bit more on using the `v-model` directive in a render function. I changed `input` to `SomeComponent`, because `modelValue` simply will not work with HTML form elements. You have to either pass a regular `value` or use for example `vModelText` directive.

visualfanatic: My suggestion would be to merge this PR and expand on `v-model` usage in a render function in another PR, what do you think?

PR #312, opened 2020-07-21
docs: added video explanation links for Vue 3

Gregg:
We have a few lessons that should help with Vue 3 learning.  They are all free.

Ultimately these should be turned into play buttons like on the original documentation, but @bencodezen encouraged me to use Markdown for now.

image

Feel free to modify and move if needed.

### DIFF
--- a/src/guide/composition-api-introduction.md
+++ b/src/guide/composition-api-introduction.md
@@ -6,6 +6,8 @@
 Reaching this far in the documentation, you should already be familiar with both [the basics of Vue](introduction.md) and [creating components](component-basics.md).
 :::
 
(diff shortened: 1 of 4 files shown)


fatony788:
saya tidak mengerti ini aplikasi buat apa ? 
tolong saya diberi tahu teman

phanan:
@Gregg @bencodezen We've already used [markups](https://github.com/vuejs/docs-next/pull/94/files#diff-1e460c4c695fe3a67c59a05dbaabe866R8) for Vue Mastery video links anyway 😬, so I guess we can just go ahead and apply styles on them.

bencodezen (approved):

PR #314, opened 2020-07-21
feat: Add `watch` changes

yusufkandemir:
Added explanation about breaking changes to component `watch` option and `$watch` instance method. Updated the related explanation and added an example for clarification.

See:
- https://codepen.io/yusufkandemir/pen/KKVJKBQ

- https://github.com/vuejs/vue-next/issues/671

- Current source code:
https://github.com/vuejs/vue-next/blob/288b4eab9e10187eb14d4d6d54dc9f077343a2a5/packages/runtime-core/src/apiWatch.ts#L308-L320

### DIFF
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -38,8 +38,6 @@
 
       // function for watching a single nested property
       this.$watch(
-        // Only `this.c.d` will be watched
-        // `this.c.e` will not be watched
         () => this.c.d,
         (newVal, oldVal) => {
           // do something
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -14,7 +14,7 @@
 
 - **Usage:**
 
-  Watch an expression or a computed function on the Vue instance for changes. The callback gets called with the new value and the old value. The expression only accepts top-level data and component property names. For more complex expressions, use a function instead.
+  Watch a string expression or a computed function on the Vue instance for changes. The callback gets called with the new value and the old value. The string expression only accepts top-level data and component property names. For more complex expressions, use a function instead.
 
 - **Example:**
 
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -31,7 +31,7 @@
       }
     },
     created() {
-      // keypath
+      // top-level property name
       this.$watch('a', (newVal, oldVal) => {
         // do something
       })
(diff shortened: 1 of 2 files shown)


src/api/instance-methods.md
```diff
@@ -14,7 +14,7 @@
 
 - **Usage:**
 
-  Watch an expression or a computed function on the Vue instance for changes. The callback gets called with the new value and the old value. The expression only accepts dot-delimited paths. For more complex expressions, use a function instead.
+  Watch a string expression or a computed function on the Vue instance for changes. The callback gets called with the new value and the old value. The string expression only accepts top-level data and component property names. For more complex expressions, use a function instead.
```
privatenumber, line 17: I believe by "expression", they mean via function:`this.$watch(() => this.c.d, ...)`

And by "computed function on the Vue instance", they mean watch a computed property on the Vue instance via property name (string).

NataliaTepluhina, line 17: Let's rephrase it a bit to make an explanation less vague:

```suggestion
  Watch a reactive property or a computed function on the Vue instance for changes. The callback gets called with the new value and the old value for the given property. We can only pass top-level `data`, `prop`, or `computed` property name as a string. For more complex expressions or nested properties, use a function instead.
```

src/api/instance-methods.md
```diff
@@ -32,7 +36,17 @@
         // do something
       })
 
-      // function
+      // function for watching a deep path
```
NataliaTepluhina, line 39: **suggestion**: let's replace `deep path` with `nested property` to reduce the ambiguity with `deep` parameter
```suggestion
      // function for watching a single nested property
```

src/api/instance-methods.md
```diff
@@ -32,7 +36,17 @@
...
+      this.$watch(
+        // Only `this.c.d` will be watched
```
NataliaTepluhina, line 41: **nitpick**: I believe with the change of code comment above, we could simply remove this explanation

src/api/instance-methods.md
yusufkandemir: It sounds like that, but it is conflicting with `The expression only accepts dot-delimited paths`. I belive the suggestions below makes everything clearer.

src/api/instance-methods.md
yusufkandemir: 👍 I removed the redundant comments.

src/api/instance-methods.md
yusufkandemir: It should be clearer with the latest changes, thanks for the explanation. 👍

PR #319, opened 2020-07-21
Removed "internals" chapter with duplicate guides

kawamataryo:
Relation https://github.com/vuejs/docs-next/issues/318

Remove `internals` guid.  It has duplicate guides.

https://github.com/vuejs/docs-next/blob/master/src/.vuepress/config.js#L71
https://github.com/vuejs/docs-next/blob/master/src/.vuepress/config.js#L86-L87

It was probably degrade in https://github.com/vuejs/docs-next/pull/243

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -42,15 +42,6 @@ const sidebar = {
       ]
     },
     {
-      title: 'Internals',
-      collapsable: false,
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@kawamataryo thanks for fixing this! 👍🏻

PR #330, opened 2020-07-23
Reword preference of Vue Testing Library over Vue Test Utils

### LINKED ISSUE #326: Vue Testing Library Recommendation
From the [docs](https://v3.vuejs.org/guide/testing.html#vue-test-utils):

>Vue Test Utils is the low-level component testing library that was written to provide users access to Vue specific APIs. While it is officially maintained, we recommend most users use Vue Testing Library, which is an abstraction over Vue Test Utils.

This new statement recommending Vue Testing Library **over** Vue Test Utils as a norm for users seems surprising. Does the core-team indeed recommend the former over the latter?  If so, is this also an indication of changes coming to the `@vue/cli-plugin-unit-jest` plugin (which does not currently use Vue Testing Library), and/or of a planned phase out of Vue Test Utils? 

A documented preference by the Vue team away from direct use of its own official tool seems noteworthy, so I wanted to ask.

afontcu:
This PR aims to clarify the statement where the official docs recommend Vue Testing Library over Vue Test Utils (the official library).

When developing [the new version of VTU](https://github.com/vuejs/vue-test-utils-next/), we wanted to frame it as an “enabling” library. VTU 2 is both a **complete library to test out components**, and also a baseline to build other tools – We even developed a plugin system. This means that VTU is still a valid choice, and that there's no need to rewrite any codebase from VTU to Vue Testing Library.

During conversarions, we agreed that Vue Testing Library had a terser, useful API and that we should recommend it to newcomers.

Closes #326

### DIFF
--- a/src/guide/testing.md
+++ b/src/guide/testing.md
@@ -86,7 +86,7 @@ Its guiding principle is that the more tests resemble the way software is used, 
 
 #### Vue Test Utils
 
-Vue Test Utils is the low-level component testing library that was written to provide users access to Vue specific APIs. While it is officially maintained, we recommend most users use Vue Testing Library, which is an abstraction over Vue Test Utils. 
+Vue Test Utils is the official low-level component testing library that was written to provide users access to Vue specific APIs. If you are starting testing Vue applications, we would recommend using Vue Testing Library, which is an abstraction over Vue Test Utils. 
 
 **Resources**
 
--- a/src/guide/testing.md
+++ b/src/guide/testing.md
@@ -86,7 +86,7 @@ Its guiding principle is that the more tests resemble the way software is used, 
 
 #### Vue Test Utils
 
-Vue Test Utils is the official low-level component testing library that was written to provide users access to Vue specific APIs. If you are starting testing Vue applications, we would recommend using Vue Testing Library, which is an abstraction over Vue Test Utils. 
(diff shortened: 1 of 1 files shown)


src/guide/testing.md
```diff
@@ -86,7 +86,7 @@ Its guiding principle is that the more tests resemble the way software is used,
 
 #### Vue Test Utils
 
-Vue Test Utils is the low-level component testing library that was written to provide users access to Vue specific APIs. While it is officially maintained, we recommend most users use Vue Testing Library, which is an abstraction over Vue Test Utils. 
+Vue Test Utils is the official low-level component testing library that was written to provide users access to Vue specific APIs. If you are starting testing Vue applications, we would recommend using Vue Testing Library, which is an abstraction over Vue Test Utils.
```
lmiller1990, line 89: Should this be "If you are new to testing" or "If you are starting a new application"?

I like the testing library philosophy more, but I like the VTU API more 😱 that said I don't mind which library is used, as long as people are testing their apps 👍

afontcu: fixed! thanks. copy&paste went wrong 😇

NataliaTepluhina:
I believe we have a coon consensus here 😅  Thanks @afontcu for the contribution!

PR #339, opened 2020-07-25
TypeScript section

NataliaTepluhina:
This is a raw draft of a TypeScript section

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -117,7 +117,7 @@ import { defineComponent } from 'vue'
 const Component = defineComponent({
   data() {
     return {
-      message: 'Hello!' as string
+      message: 'Hello!'
     }
   },
   computed: {
@@ -125,6 +125,16 @@ const Component = defineComponent({
     greeting(): string {
       return this.message + '!'
     }
+
+    // in a computed with a setter, getter needs to be annotated
+    greetingUppercased: {
+      get(): number {
+        return this.greeting.length;
+      },
+      set(newValue) {
+        this.message = newValue.toUpperCase();
+      },
+    },
   }
 })
 ```
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -70,14 +70,13 @@ const Component = defineComponent({
 
 ## Using with Options API
 
-You can type your `data` properties using type assertions:
+TypeScript should be able to infer most of the types without defining types explicitly. For example, if you have a component with a number `count` property, you will have an error if you try to call a string-specific method on it:
 
-```js
+```ts
 const Component = defineComponent({
   data() {
     return {
-      message: 'Hello!' as string,
-      count: 0 as number
+      count: 0
     }
   },
   mounted() {
@@ -86,7 +85,7 @@ const Component = defineComponent({
 })
 ```
 
-If you need a complex type, you can create an interface:
+If you have a complex type or interface, you can cast it using [type assertion](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions):
 
 ```ts
 interface Book {
@@ -98,13 +97,11 @@ interface Book {
 const Component = defineComponent({
   data() {
     return {
-      message: 'Hello!' as string,
-      count: 0 as number,
-      book: <Book>{
+      book: {
         title: 'Vue 3 Guide',
         author: 'Vue Team',
         year: 2020
-      }
+      } as Book
     }
   }
 })
@@ -112,7 +109,7 @@ const Component = defineComponent({
 
 ### Annotating Return Types
 
-Because of the circular nature of Vue’s declaration files, TypeScript may have difficulties inferring the types of certain methods. For this reason, you may need to annotate the return type on methods like render and those in computed.
+Because of the circular nature of Vue’s declaration files, TypeScript may have difficulties inferring the types of computed. For this reason, you may need to annotate the return type computed properties.
 
 ```ts
 import { defineComponent } from 'vue'
@@ -123,14 +120,8 @@ const Component = defineComponent({
       message: 'Hello!' as string
     }
   },
-  methods: {
-    // need annotation due to `this` in return type
-    greet(): string {
-      return this.message + ' world'
-    }
-  },
   computed: {
-    // need annotation
+    // needs an annotation
     greeting(): string {
       return this.message + '!'
     }
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -207,7 +207,7 @@ const Component = defineComponent({
 })
 ```
 
-Sometimes we may need to specify complex types for a ref's inner value. We can do that succinctly by passing a generics argument when calling ref to override the default inference:
+Sometimes we may need to specify complex types for a ref's inner value. We can do that simply passing a generic argument when calling ref to override the default inference:
 
 ```ts
 const year = ref<string | number>('2020') // year's type: Ref<string | number>
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -236,6 +236,8 @@ interface Book {
 export default defineComponent({
   name: 'HelloWorld',
   setup() {
+    const book = reactive<Book>({ title: 'Vue 3 Guide' })
+    // or
     const book: Book = reactive({ title: 'Vue 3 Guide' })
     // or
     const book = reactive({ title: 'Vue 3 Guide' }) as Book
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -195,14 +195,16 @@ const Component = defineComponent({
 
 ### Typing `ref`s
 
-To type a ref, we need to specify a type when calling `ref()` method:
+Refs infer the type from the initial value:
 
 ```ts
 import { defineComponent, ref } from 'vue'
 
 const Component = defineComponent({
   setup() {
-    const year = ref<number>(2020)
+    const year = ref(2020)
+
+    const result = year.value.split('') // => Property 'filter' does not exist on type 'number'
   }
 })
 ```
@@ -215,18 +217,13 @@ const year = ref<string | number>('2020') // year's type: Ref<string | number>
 year.value = 2020 // ok!
 ```
 
-If the type of the generic is unknown, it's recommended to cast `ref` to `Ref<T>`:
-
-```ts
-function useState<State extends string>(initial: State) {
-  const state = ref(initial) as Ref<State> // state.value -> State extends string
-  return state
-}
-```
+::: tip Note
+If the type of the generic is unknown, it's recommended to cast `ref` to `Ref<T>`.
+:::
 
 ### Typing `reactive`
 
-Similarly to `ref`, we can define a `reactive` type on method call:
+When typing a `reactive` property, we can use use interfaces:
 
 ```ts
 import { defineComponent, reactive } from 'vue'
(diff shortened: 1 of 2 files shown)


src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+
+const Component = defineComponent({
+  data() {
+    return {
+      message: 'Hello!' as string
+    }
+  },
+  methods: {
+    // need annotation due to `this` in return type
+    greet(): string {
```
pikax, line 128: I don't think the annotation here is necessary, only `computed` is required, I added a dts check just to be sure 😄 
https://github.com/vuejs/vue-next/pull/1708

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+      return this.message + ' world'
+    }
+  },
+  computed: {
```
pikax, line 132: Should we add an example with `get` and `set`? 

just added to DTS tests: https://github.com/vuejs/vue-next/pull/1709

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+      return this.message + '!'
+    }
+  }
+})
+```
+
+### Annotating Props
+
+```ts
+{ defineComponent, PropType } from 'vue'
```
pikax, line 144:
```suggestion
import { defineComponent, PropType } from 'vue'
```

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+    // need annotation
+    greeting(): string {
+      return this.message + '!'
+    }
+  }
+})
+```
+
+### Annotating Props
+
```
pikax, line 142: We should give some explaination:

"Vue does runtime validation on `props` to provide typescript we need to cast the constructor with `PropType`." - as usual better working advised 😄

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+    }
+  }
+})
+```
+
+If you find validator not getting type inference or member completion isn’t working, annotating the argument with the expected type may help address these problems.
+
+## Using with Composition API
+
+On `setup()` function, you don't need ass a typing to `props` parameter as it will infer types from `props` component option.
```
pikax, line 173:
```suggestion
On `setup()` function, you don't need pass a typing to `props` parameter as it will infer types from `props` component option.
```

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+  props: {
+    message: {
+      type: String,
+      required: true
+    }
+  },
+
+  setup(props) {
+    /* props will be typed as
+    (parameter) props: Readonly<{
```
pikax, line 188: No sure about providing the full type here, probably just showing the `props.message.split('')` is valid because is string

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+
+```ts
+import { defineComponent, ref, rcomputed } from "vue";
+
+export default defineComponent({
+  name: "HelloWorld",
+  setup() {
+    let count = ref<number>(0);
+
+    // read-only, will have a type of Readonly<Ref<Readonly<number>>>
```
pikax, line 261: Although  technically correct, must users won't understand the type here.

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+    const double = computed<number>(() => count.value * 2);
+
+    // writable, will have a type of Ref<T>
+    const triple = computed<number>({
+      get: () => count.value * 3,
+      set: (value: number) => value,
```
pikax, line 267: the `value:number` is not required, even without `<number>` typescript is able to infer the type correct.

Ideally typescript should be able to infer 80% of the cases, we just need to give to provide some information when those types are not inferred automatically.

src/guide/typescript-support.md
NataliaTepluhina: Yup, I will add it to be consistent with Composition API part

src/guide/typescript-support.md
NataliaTepluhina: I think I will add this to the beginning of the TS section 👍🏻

NataliaTepluhina:
@pikax thank you for all the invaluable suggestions! I've applied fixes, could you please take a second look? 🙏🏻

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+### Typing `ref`s
+
+To type a ref, we need to specify a type when calling `ref()` method:
+
+```ts
+import { defineComponent, ref } from 'vue'
+
+const Component = defineComponent({
+  setup() {
+    const year = ref<number>(2020)
```
pikax, line 205:
```ts
    const year = ref(2020) // equivalent to ref<number>(2020)
```
In this cases `ref` should infer the type

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+  }
+})
+```
+
+Sometimes we may need to specify complex types for a ref's inner value. We can do that succinctly by passing a generics argument when calling ref to override the default inference:
```
pikax, line 210:
```suggestion
Sometimes we may need to specify complex types for a ref's inner value. We can do that simply passing a generic argument when calling ref to override the default inference:
```

Simplifying wording

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+
+```ts
+const year = ref<string | number>('2020') // year's type: Ref<string | number>
+
+year.value = 2020 // ok!
+```
+
+If the type of the generic is unknown, it's recommended to cast `ref` to `Ref<T>`:
```
pikax, line 218: Not sure about this paragraph, the following example `State` is not `unknown`, I think it will make it better by just having, when using with generic is recommended to cast ...

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+
+interface Book {
+  title: string
+  year?: number
+}
+
+export default defineComponent({
+  name: 'HelloWorld',
+  setup() {
+    const book = reactive<Book>({
```
pikax, line 242: Here we can actually show that the `reactive` can be casted to the `Book` because if the `Book` doesn't have any nested `ref`, it will be pretty much the same as `Book`.

In this case, this would be equivalent: 
```ts
const book: Book = reactive({ title: 'Vue 3 Guide' });
// or 
const book = reactive({ title: 'Vue 3 Guide' }) as Book;
```

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+
+For computed, we could specify a type with getter only, or with getter and setter:
+
+```ts
+import { defineComponent, ref, computed } from "vue";
+
+export default defineComponent({
+  name: "HelloWorld",
+  setup() {
+    let count = ref<number>(0);
```
pikax, line 259: should we remove the `<number>` from this examples? because typescript will be able to infer the correct types in this setup.

src/guide/typescript-support.md
```diff
@@ -0,0 +1,270 @@
...
+```ts
+function useState<State extends string>(initial: State) {
+  const state = ref(initial) as Ref<State> // state.value -> State extends string
+  return state
+}
+```
+
+### Typing `reactive`
+
+Similarly to `ref`, we can define a `reactive` type on method call:
```
pikax, line 229: Probably we can mention that `reactive` should have the same type as the argument, only if there's unwrapped ref, that's when the types will be different

src/guide/typescript-support.md
NataliaTepluhina: Removed an example 👍🏻

src/guide/typescript-support.md
NataliaTepluhina: @pikax if we assign this to `book`:

```ts
const book: Book = reactive({ title: "Vue 3 Guide", year: ref(2020) });
```

There is no error thrown. However, if we try to reassign the property, error is present:

```ts
const book: Book = reactive({ title: "Vue 3 Guide" });

book.year = ref(2020) // 'Ref<number>' is not assignable to type 'number'
```

Is there an unwrapping happening somewhere? 🤔 
```

File: src/guide/typescript-support.md (hunk 14 above)
Comment (NataliaTepluhina) [reply src/guide/typescript-support.md:229]: @pikax if we assign this to `book`:

```ts
const book: Book = reactive({ title: "Vue 3 Guide", year: ref(2020) });
```

There is no error thrown. However, if we try to reassign the property, error is present:

```ts
const book: Book = reactive({ title: "Vue 3 Guide" });

book.year = ref(2020) // 'Ref<number>' is not assignable to type 'number'
```

Is there an unwrapping happening somewhere? 🤔 
```

**Comment by NataliaTepluhina** [comment]
@pikax thank you! Applied more fixes and there is still one question remaining. Back to you ⚽

File: src/guide/typescript-support.md (hunk 14 above)
Comment (pikax) [reply src/guide/typescript-support.md]: Yes, there's unwrapping when you use `reactive`. 

Also when you assign a `ref` to a value in `reactive` it will `unwrap`, but unfortunate  we can't do it in typescript because it doesn't allow us to define a setter with a different type from getter https://github.com/microsoft/TypeScript/issues/2521

Comment (NataliaTepluhina) [reply src/guide/typescript-support.md]: Gotcha :( Should we leave this example as it is now or it's better to remove it?

File: src/guide/typescript-support.md [hunk 15]
```diff
@@ -0,0 +1,280 @@
+# TypeScript Support
+
+> [Vue CLI](https://cli.vuejs.org) provides built-in TypeScript tooling support.
```
sdras, line 3: Is this true as of a particular version? If so, we should say so- or even which version is best for TS

src/guide/typescript-support.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Official Declaration in NPM Packages
+
+A static type system can help prevent many potential runtime errors, especially as applications grow. That's why Vue is written in TypeScript. This means you don't need any additional tooling to use TypeScript with Vue - it has a first-class citizen support
```
sdras, line 7: Second sentence is a fragment, how about:

>A static type system can help prevent many potential runtime errors as applications grow, which is why Vue 3 is written in TypeScript.

sdras: Also missing a period at the end of the paragraph

src/guide/typescript-support.md
NataliaTepluhina: It's true since Vue CLI 3.x so we can leave it 🎉

src/guide/typescript-support.md
pikax: I would change the example to: 
```ts
const book: Book = reactive({ title: "Vue 3 Guide", year: ref(20)});
```

NataliaTepluhina: @pikax but it won't throw an error 🤔 

I don't mind changing it anyways 👍🏻

pikax: Do you want to throw error?

src/guide/typescript-support.md
NataliaTepluhina: Fixed! Thank you @sdras 🙇🏻‍♀️

src/guide/typescript-support.md
```diff
@@ -0,0 +1,280 @@
...
+For computed, we could specify a type with getter only, or with getter and setter:
+
+```ts
+import { defineComponent, ref, computed } from "vue";
+
+export default defineComponent({
+  name: "HelloWorld",
+  setup() {
+    let count = ref(0);
+
```
pikax, line 257: can we have an example with `computed` automatically infer the return type? 

The `read-only` bellow can become override computed type

NataliaTepluhina: @pikax done!

src/guide/typescript-support.md
```diff
@@ -0,0 +1,264 @@
...
+interface Book {
+  title: string
+  year?: number
+}
+
+export default defineComponent({
+  name: 'HelloWorld',
+  setup() {
+    const book: Book = reactive({ title: 'Vue 3 Guide' })
```
pikax, line 239:
```suggestion
    const book = reactive<Book>({ title: 'Vue 3 Guide' })
    // or
    const book: Book = reactive({ title: 'Vue 3 Guide' })
```

Just adding the other possible example

src/guide/typescript-support.md
```diff
@@ -0,0 +1,266 @@
...
+```ts
+import { defineComponent, ref, computed } from 'vue'
+
+export default defineComponent({
+  name: 'HelloWorld',
+  setup() {
+    let count = ref(0)
+
+    // read-only
+    const double = computed(() => count.value)
```
edimitchel, line 261: Compute a value without any change is IMHO a little too simply..

edimitchel:
@NataliaTepluhina did you see my comment ?

PR #343, opened 2020-07-26
Replaced first example in Introduction

NataliaTepluhina:
Following the reasoning in https://github.com/vuejs/docs-next/issues/253, I've replaced the first example with Vue component with a reactive property changing over time. I'm still not sure if it's the best decision so would be open to hearing your opinions

Close https://github.com/vuejs/docs-next/issues/253

### DIFF
--- a/src/.vuepress/components/FirstExample.vue
+++ b/src/.vuepress/components/FirstExample.vue
@@ -1,0 +1,31 @@
+<template>
+  <div class="demo">Counter: {{ counter }}</div>
+</template>
+
+<script>
+export default {
+  data() {
+    return {
+      counter: 0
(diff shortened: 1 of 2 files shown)


bencodezen:
Nice job on this @NataliaTepluhina! Do you think it would be worth adding a pause / resume button on the example since it technically run infinitely? And to be clear, I don't think we should add it to the code snippet itself, but just on the rendered version as a UX enhancement.

NataliaTepluhina:
Thanks @bencodezen! Makes sense to me, will add a button 👍🏻

NataliaTepluhina:
@bencodezen fix applied!

Issue #342, opened 2020-07-26
V-model Modifiers Confusion


joezimdelta:
https://v3.vuejs.org/guide/component-custom-events.html#handling-v-model-modifiers says 

> In 2.x, we have hard-coded support for modifiers like .trim on component v-model. However, it would be more useful if the component can support custom modifiers. In 3.x, modifiers added to a component v-model will be provided to the component via the modelModifiers prop:

It then shows a few examples of how modifiers get passed down and such, but sometimes they implement the capitalize modifier and sometimes they don't. Is capitalize a built-in modifier? Are there any built-in modifiers still? The language and examples give no way to determine one way or the other.

NataliaTepluhina:
@joezimdelta thank you for reporting this! Makes sense to me, we will add a note about existing built-in modifiers as well as specify that you can extend this list with custom modifiers 👍🏻

NataliaTepluhina:
Closing an issue as related PR was merged

PR #344, opened 2020-07-26
vmodel modifiers clarify example and diff 2.x v 3.x

marina-mosti:
Issue #342

### DIFF
--- a/src/guide/component-custom-events.md
+++ b/src/guide/component-custom-events.md
@@ -141,7 +141,13 @@ app.component('user-name', {
 
 ## Handling `v-model` modifiers
 
-In 2.x, we have hard-coded support for modifiers like `.trim` on component `v-model`. However, it would be more useful if the component can support custom modifiers. In 3.x, modifiers added to a component `v-model` will be provided to the component via the modelModifiers prop:
+In Vue 2.x, we had hard-coded support for modifiers like `.trim` or `.native` on component `v-model` bindings. And even though Vue 3.x also supports [out of the box modifiers](https://v3.vuejs.org/guide/forms.html#modifiers), it would be more useful if our components could support custom modifiers of our own. 
+
+Let's create an example custom modifier, `capitalize`, that capitalizes the first letter of the string provided by the `v-model` binding.
+
+In 3.x, modifiers added to a component `v-model` will be provided to the component via the `modelModifiers` prop. In the below example, we have created a component that contains a `modelModifiers` prop and defaults to an empty object.
+
+Notice that when the component's `created` lifecycle hook triggers, the `modelModifiers` prop contains `capitalize` and it's value is `true` - due to it being set on the `v-model` binding `v-model.capitalize="bar"`.
 
 ```html
 <my-component v-model.capitalize="bar"></my-component>
@@ -166,7 +172,7 @@ app.component('my-component', {
 })
 ```
 
-We can check `modelModifiers` object keys and write a handler to change the emitted value. In the code below we will capitalize the string:
+Now that we have our prop set up, we can check the `modelModifiers` object keys and write a handler to change the emitted value. In the code below we will capitalize the string whenever the `<input />` element fires an `input` event.
 
 ```html
 <div id="app">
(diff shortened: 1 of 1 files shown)


src/guide/component-custom-events.md
```diff
@@ -141,7 +141,13 @@ app.component('user-name', {
 
 ## Handling `v-model` modifiers
 
-In 2.x, we have hard-coded support for modifiers like `.trim` on component `v-model`. However, it would be more useful if the component can support custom modifiers. In 3.x, modifiers added to a component `v-model` will be provided to the component via the modelModifiers prop:
+In Vue 2.x, we had hard-coded support for modifiers like `.trim` or `.native` on component `v-model` bindings. And even though Vue 3.x also supports [out of the box modifiers](/guide/forms.html#modifiers), it would be more useful if our components could support custom modifiers of our own.
```
NataliaTepluhina, line 144: **suggestion**: I think we should remove 2.x mention here completely and just mention we have built-in modifiers with the link as nothing changes except `.native` modifier.

Also, let's remove the statement about custom modifiers being **more** useful (it really depends on the use-case). What about just `useful`?

```suggestion
As we already mentioned, `v-model` has [built-in modifiers](/guide/forms.html#modifiers) like `.trim` or `.native`. In some cases, you might also want to add your own custom modifier.
```

src/guide/component-custom-events.md
```diff
@@ -141,7 +141,13 @@ app.component('user-name', {
 
 ## Handling `v-model` modifiers
 
-In 2.x, we have hard-coded support for modifiers like `.trim` on component `v-model`. However, it would be more useful if the component can support custom modifiers. In 3.x, modifiers added to a component `v-model` will be provided to the component via the modelModifiers prop:
+In Vue 2.x, we had hard-coded support for modifiers like `.trim` or `.native` on component `v-model` bindings. And even though Vue 3.x also supports [out of the box modifiers](/guide/forms.html#modifiers), it would be more useful if our components could support custom modifiers of our own. 
+
+Let's create an example custom modifier, `capitalize`, that capitalizes the first letter of the string provided by the `v-model` binding.
+
+In 3.x, modifiers added to a component `v-model` will be provided to the component via the `modelModifiers` prop. In the below example, we have created a component that contains a `modelModifiers` prop and defaults to an empty object.
```
NataliaTepluhina, line 148: **suggestion**: we are on 3.x docs already so we could remove version mention 😅 

```suggestion
Modifiers added to a component `v-model` will be provided to the component via the `modelModifiers` prop. In the below example, we have created a component that contains a `modelModifiers` prop that defaults to an empty object.
```

src/guide/component-custom-events.md
```diff
@@ -141,7 +141,13 @@ app.component('user-name', {
...
+
+Notice that when the component's `created` lifecycle hook triggers, the `modelModifiers` prop contains `capitalize` and it's value is `true` - due to it being set on the `v-model` binding `v-model.capitalize="bar"`.
```
NataliaTepluhina, line 150:
```suggestion
Notice that when the component's `created` lifecycle hook triggers, the `modelModifiers` prop contains `capitalize` and its value is `true` - due to it being set on the `v-model` binding `v-model.capitalize="bar"`.
```

marina-mosti: oops 😅

marina-mosti:
Thanks for the feedback @NataliaTepluhina 
I've gone ahead and made the changes, with a little rewording of the intro paragraph

src/guide/component-custom-events.md
```diff
@@ -141,7 +141,13 @@ app.component('user-name', {
 
 ## Handling `v-model` modifiers
 
-In 2.x, we have hard-coded support for modifiers like `.trim` on component `v-model`. However, it would be more useful if the component can support custom modifiers. In 3.x, modifiers added to a component `v-model` will be provided to the component via the modelModifiers prop:
+When we were learning about form input bindings, we saw that `v-model` has [built-in modifiers](/guide/forms.html#modifiers) like `.trim` or `.native`. In some cases, however, you might also want to add your own custom modifiers.
```
NataliaTepluhina, line 144: My bad, I haven't caught it on the first round of review: `.native` does not exist anymore, so we should remove it

marina-mosti: Oops, yeah. Removed.

Issue #352, opened 2020-07-28
Search page


NoelDeMartin:
In v2 there was a separate search page at `/search` in order to have a url to run searches. For example: https://vuejs.org/v2/search/?q=v%2Dmodel

I implemented that in [this PR](https://github.com/vuejs/vuejs.org/pull/1488), so you can take a look there for the motivations and discussion.

Would it make sense to do the same here? Would it make more sense to implement this feature in VuePress instead?

NataliaTepluhina:
@NoelDeMartin it would be great to have the same page for new docs! Regarding 'where to implement' question: I believe it would be beneficial for VuePress as well but if you think it would be faster/easier to add it to docs website only - let's implement it here first.

You also reminded me that I need to apply for Algolia DocSearch token 😅

bencodezen:
@NoelDeMartin VuePress comes with a built-in search functionality, so I wonder if it makes sense to see if there's a way you can leverage it to build the search page. That said, if VuePress search functionality could use some improvement, then I think we could use your help on updating the VuePress library itself so a lot more people can benefit from your ideas!

NoelDeMartin:
Hi @bencodezen. I've been looking into the [@vuepress/plugin-search](https://vuepress.vuejs.org/plugin/official/plugin-search.html#install) plugin and I'll probably make a PR to add it there afterwards.

But the docs at the moment are build using VuePress `1.3`, and the latest version is `1.5`. So I think it's still a good idea to add it here first. I'm not sure if the maintainers would accept a PR upgrading the VuePress dependency, and I'd also have to wait until a version with the search page is released.

If I create a PR here first, it's more likely that it gets merged before the official release of Vue 3 :).

PR #363, opened 2020-07-31
Implement search page

NoelDeMartin:
Added search page as discussed in #352.

Given that the Algolia credentials are not available yet, I've used the ones from v2 for development. If you want to do the same, add the following to `src/.vuepress/config.js` under `themeConfig`:

```js
algolia: {
  appId: 'BH4D9OD16A',
  apiKey: '85cc3221c9f23bfbaa4e3913dd7625ea',
  indexName: 'vuejs',
  algoliaOptions: { facetFilters: ["version:2.6"] },
}
```

I've implemented this looking at the [current implementation in v2 docs](https://github.com/vuejs/vuejs.org/pull/1488).

There are some improvements that can be done given the new architecture, for example falling back to VuePress-powered search if Algolia is not configured or unavailable. However, as we've discussed in #352, that would be a nice addition to VuePress itself. So I think it's best to have this for now that is very similar to what we had in v2 and work on the improvements in VuePress.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "axios": "^0.19.1",
+    "intersection-observer": "^0.11.0",
     "showdown": "^1.9.1"
   }
 }
--- a/src/.vuepress/components/search/index.vue
+++ b/src/.vuepress/components/search/index.vue
@@ -1,0 +1,264 @@
+<template>
+  <main id="search-page">
+
+    <p v-if="!isAlgoliaConfigured">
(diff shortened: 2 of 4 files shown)


NataliaTepluhina:
@NoelDeMartin thank you for working on this! I see we have this error on build: `ReferenceError: location is not defined  at VueComponent.data (src/.vuepress/components/search/index.vue:57:0)`. Could you please look into it and fix it?

NoelDeMartin:
@NataliaTepluhina Thanks for the heads-up, I forgot about SSR 😅 It should be fixed now.

PR #370, opened 2020-08-02
Added `outerHTML` -> `innerHTML` el change to docs

### LINKED ISSUE #335: Migration guide: document changes in mounting point
In Vue 3.0, there's a subtle breaking change in app mounting ([see RFC](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0009-global-api-change.md#mounting-behavior-difference-from-2x)).

In Vue 2.x, app element's `outerHTML` is replaced (and eventually compiled to a template, if root component has no `template`/`render`). Vue 3.x now uses app element's `innerHTML`, instead.

This change usually is harmless, but code that expects previous behavior may break (e.g. https://github.com/vuejs/vue-next/issues/1642 and https://github.com/vuejs/vue-next/issues/1696).

I couldn't find any mention to this behavior change in migration guide. I'm sorry if this is not the case.

NataliaTepluhina:
As described in #335, we need to highlight the change regarding the application container on the migration guide: now on mounting an app, we use container `innerHTML` instead of `outerHTML` as it was on 2.x. I've also changed the lifecycle hooks diagram to reflect this change

Close #335


NataliaTepluhina:
Thanks @phanan! Let me try to optimize it 😅

PR #372, opened 2020-08-02
Fixed Simple State management section to use reactive

### LINKED ISSUE #368: Simple State Management code doesn't work with the new reactivity system
I know the docs are still being updated from Vue 2 so I hope this issue is relevant.

In the [Simple State Management from Scratch](https://v3.vuejs.org/guide/state-management.html#simple-state-management-from-scratch) section of the Guide, the example code relies on the old reactivity system which implicitly modified the source object to detect mutations. The new system leaves the underlying object alone and detects mutations through a separate proxy object.

After the first snippet with `sourceOfTruth` being used as the data for two components, the guide explains that mutations to `sourceOfTruth` will update both components, but actually a direct mutation to `sourceOfTruth` will have no effect. With the simple store example, the same is true for the `state` property. In fact, the store snippet won't work as none of the store mutation methods will have an effect.

One fix is to wrap each of the objects with `reactive()`:
```typescript
const sourceOfTruth = reactive({
    message: 'Hello'
})
```
```typescript
  // ...
  state: reactive({
    message: 'Hello!'
  }),
  // ...
```
The snippets using `reactive()` are less magical than the old version so the introductory paragraph in this section should probably explain more about how the reactivity system enables the technique.

NataliaTepluhina:
In the `Simple State Management` section, we had a non-working example with sharing a plain object without making it reactive. This PR fixes the code snippets to use `reactive`

Close #368

### DIFF
--- a/src/guide/state-management.md
+++ b/src/guide/state-management.md
@@ -10,39 +10,56 @@ If you're coming from React, you may be wondering how vuex compares to [redux](h
 
 ## Simple State Management from Scratch
 
-It is often overlooked that the source of truth in Vue applications is the raw `data` object - a component instance only proxies access to it. Therefore, if you have a piece of state that should be shared by multiple instances, you can share it by identity:
+It is often overlooked that the source of truth in Vue applications is the reactive `data` object - a component instance only proxies access to it. Therefore, if you have a piece of state that should be shared by multiple instances, you can use a [reactive](http://localhost:8080/guide/reactivity-fundamentals.html#declaring-reactive-state) method to make an object reactive:
 
-``` js
-const sourceOfTruth = {
-    message: 'Hello'
-}
+```js
+const sourceOfTruth = Vue.reactive({
+  message: 'Hello'
+})
 
 const appA = Vue.createApp({
-  data () {
+  data() {
     return sourceOfTruth
   }
 }).mount('#app-a')
 
 const appB = Vue.createApp({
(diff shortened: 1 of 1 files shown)


src/guide/state-management.md
```diff
@@ -10,51 +10,68 @@ If you're coming from React, you may be wondering how vuex compares to [redux](h
 
 ## Simple State Management from Scratch
 
-It is often overlooked that the source of truth in Vue applications is the raw `data` object - a component instance only proxies access to it. Therefore, if you have a piece of state that should be shared by multiple instances, you can share it by identity:
+It is often overlooked that the source of truth in Vue applications is the reactive `data` object - a component instance only proxies access to it. Therefore, if you have a piece of state that should be shared by multiple instances, you can use a [reactive](http://localhost:8080/guide/reactivity-fundamentals.html#declaring-reactive-state) method to make an object reactive:
```
phanan, line 13:
```suggestion
It is often overlooked that the source of truth in Vue applications is the reactive `data` object - a component instance only proxies access to it. Therefore, if you have a piece of state that should be shared by multiple instances, you can use a [reactive](/guide/reactivity-fundamentals.html#declaring-reactive-state) method to make an object reactive:
```

NataliaTepluhina: @phanan my bad 😅  thank you for fixing this!

PR #396, opened 2020-08-10
docs (#391): update provide inject in composition api with enhanced reactivity

### LINKED ISSUE #391: Provide / Inject: Use readonly
In https://v3.vuejs.org/guide/composition-api-provide-inject.html

This is the current example for allowing components in which the object is injected to change the object.

```js
  const book = reactive({
    title: 'Vue 3 Guide',
    author: 'Vue Team'
  })

  function changeBookName() {
    book.title = 'Vue 3 Advanced Guide'
  }

  provide('book', book)
  provide('changeBookName', changeBookName)
```

I believe that best practice would be to mark the book as readonly and force the use of the setter that way:

```js
  const book = reactive({
    title: 'Vue 3 Guide',
    author: 'Vue Team'
  })

  function changeBookName() {
    book.title = 'Vue 3 Advanced Guide'
  }

  // NEW
  // readonly prevents direct mutation of book  and forces the use of the changeBookName function.
  provide('book', readonly(book))
  provide('changeBookName', changeBookName)
```

(and btw, this warning should best be placed in the normal [provide/inject](https://v3.vuejs.org/guide/component-provide-inject.html#working-with-reactivity) guide)

bencodezen:
The PR was originally supposed to just be a small change regarding `readonly`, but then I came up with this scenario to explain provide / inject that I thought would be helpful. 

[Netlify Preview](https://deploy-preview-396--vue-docs-next-preview.netlify.app/guide/composition-api-provide-inject.html#scenario-background)

Feedback welcome as always!

Closes #391

### DIFF
--- a/src/guide/composition-api-provide-inject.md
+++ b/src/guide/composition-api-provide-inject.md
@@ -107,7 +107,9 @@ export default {
 </script>
 ```
 
-## Adding Reactivity
+## Reactivity
+
+### Adding Reactivity
 
 To add reactivity between provided and injected values, we can use a [ref](reactivity-fundamentals.html#creating-standalone-reactive-values-as-refs) or [reactive](reactivity-fundamentals.html#declaring-reactive-state) when providing a value.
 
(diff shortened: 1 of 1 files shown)


src/guide/composition-api-provide-inject.md
```diff
@@ -1,120 +1,261 @@
...
+
+The `provide` function allows you to define the property through two parameters:
+
+1. The property's name (`<String>` type)
+2. The property's value
 
-const RootComponent = {
+Using our `MyMap` component, our provided values can be refactored as the following:
+
+```vue{7,14-20}
```
phanan, line 55: Nice! I didn't know we could do this. But is it necessary here though, if `vue` is meant for the whole SFC syntax? It might also create more maintenance burden (one would need to update these numbers whenever the code is changed).

src/guide/composition-api-provide-inject.md
```diff
@@ -1,120 +1,261 @@
...
-  provide('changeBookName', changeBookName)
 }
+</script>
+```
 
-// in consumer
-setup() {
-  const book = inject('book')
-  const changeBookName = inject('changeBookName')
+:::tip
```
phanan, line 247: suggestion: To remove the `:::tip` and blend this content into the current flow. WDYT?

bencodezen: I was debating this too. With your recommendation, let me blend it in better

src/guide/composition-api-provide-inject.md
bencodezen: I'm a little confused. Are you saying we shouldn't use line highlighting? Or are you recommending splitting it up into HTML and JS for easier maintenance?

phanan: Oops sorry, I misread the code. Ignore me, all good :)

bencodezen:
Thanks for the review and feedback @phanan!

Issue #18, opened 2020-01-08
Migrate Tooling content


bencodezen:
# Requirements

- [x]  Migrate the following pages:
    - [x]  Single File Components #118 
    - [x]  Unit Testing
    - [x]  TypeScript Support
    - [x]  Production Deployment

NataliaTepluhina:
Closing this issue as all the mentioned items were migrated

PR #412, opened 2020-08-17
docs (#18): add deployment page

bencodezen:
This is mostly a copy and paste from the v2 docs. As far as I'm aware, I don't think anything major has changed CLI wise from v2 to v3. Let me know if I missed anything @sodatea!

P.S. Feel free to add any other reviewers who might know more about this section.

[Netlify Preview](https://deploy-preview-412--vue-docs-next-preview.netlify.app/guide/tooling/deployment.html)

### DIFF
--- a/src/guide/tooling/deployment.md
+++ b/src/guide/tooling/deployment.md
@@ -1,0 +1,132 @@
+# Production Deployment
+
+::: info
+Most of the tips below are enabled by default if you are using [Vue CLI](https://cli.vuejs.org). This section is only relevant if you are using a custom build setup.
+:::
+
+## Turn on Production Mode
+
+During development, Vue provides a lot of warnings to help you with common errors and pitfalls. However, these warning strings become useless in production and bloat your app's payload size. In addition, some of these warning checks have small runtime costs that can be avoided in production mode.
+
+### Without Build Tools
+
+If you are using the full build, i.e. directly including Vue via a script tag without a build tool, make sure to use the minified version (`vue.min.js`) for production. Both versions can be found in the [Installation guide](installation.html#Direct-lt-script-gt-Include).
+
+### With Build Tools
+
+When using a build tool like Webpack or Browserify, the production mode will be determined by `process.env.NODE_ENV` inside Vue's source code, and it will be in development mode by default. Both build tools provide ways to overwrite this variable to enable Vue's production mode, and warnings will be stripped by minifiers during the build. All `vue-cli` templates have these pre-configured for you, but it would be beneficial to know how it is done:
+
+#### Webpack
+
+In Webpack 4+, you can use the `mode` option:
+
+```js
+module.exports = {
+  mode: 'production'
+}
+```
+
+But in Webpack 3 and earlier, you'll need to use [DefinePlugin](https://webpack.js.org/plugins/define-plugin/):
+
+```js
+var webpack = require('webpack')
+
+module.exports = {
+  // ...
+  plugins: [
+    // ...
+    new webpack.DefinePlugin({
+      'process.env.NODE_ENV': JSON.stringify('production')
+    })
+  ]
+}
+```
+
+#### Browserify
+
+- Run your bundling command with the actual `NODE_ENV` environment variable set to `"production"`. This tells `vueify` to avoid including hot-reload and development related code.
+
+- Apply a global [envify](https://github.com/hughsk/envify) transform to your bundle. This allows the minifier to strip out all the warnings in Vue's source code wrapped in env variable conditional blocks. For example:
+
+  ```bash
+  NODE_ENV=production browserify -g envify -e main.js | uglifyjs -c -m > build.js
+  ```
+
+- Or, using [envify](https://github.com/hughsk/envify) with Gulp:
+
+  ```js
+  // Use the envify custom module to specify environment variables
+  var envify = require('envify/custom')
+
+  browserify(browserifyOptions)
+    .transform(vueify)
+    .transform(
+      // Required in order to process node_modules files
+      { global: true },
+      envify({ NODE_ENV: 'production' })
+    )
+    .bundle()
+  ```
+
+- Or, using [envify](https://github.com/hughsk/envify) with Grunt and [grunt-browserify](https://github.com/jmreidy/grunt-browserify):
+
+  ```js
+  // Use the envify custom module to specify environment variables
+  var envify = require('envify/custom')
+
+  browserify: {
+    dist: {
+      options: {
+        // Function to deviate from grunt-browserify's default order
+        configure: b =>
+          b
+            .transform('vueify')
+            .transform(
+              // Required in order to process node_modules files
+              { global: true },
+              envify({ NODE_ENV: 'production' })
+            )
+            .bundle()
+      }
+    }
+  }
+  ```
+
+#### Rollup
+
+Use [@rollup/plugin-replace](https://github.com/rollup/plugins/tree/master/packages/replace):
+
+```js
+const replace = require('@rollup/plugin-replace')
+
+rollup({
+  // ...
+  plugins: [
+    replace({
+      'process.env.NODE_ENV': JSON.stringify( 'production' )
+    })
+  ]
+}).then(...)
+```
+
+## Pre-Compiling Templates
+
+When using in-DOM templates or in-JavaScript template strings, the template-to-render-function compilation is performed on the fly. This is usually fast enough in most cases, but is best avoided if your application is performance-sensitive.
+
+The easiest way to pre-compile templates is using [Single-File Components](single-file-components.html) - the associated build setups automatically performs pre-compilation for you, so the built code contains the already compiled render functions instead of raw template strings.
+
+If you are using Webpack, and prefer separating JavaScript and template files, you can use [vue-template-loader](https://github.com/ktsn/vue-template-loader), which also transforms the template files into JavaScript render functions during the build step.
+
+## Extracting Component CSS
+
+When using Single-File Components, the CSS inside components are injected dynamically as `<style>` tags via JavaScript. This has a small runtime cost, and if you are using server-side rendering it will cause a "flash of unstyled content". Extracting the CSS across all components into the same file will avoid these issues, and also result in better CSS minification and caching.
+
+Refer to the respective build tool documentations to see how it's done:
+
+- [Webpack + vue-loader](https://vue-loader.vuejs.org/en/configurations/extract-css.html) (the `vue-cli` webpack template has this pre-configured)
+- [Browserify + vueify](https://github.com/vuejs/vueify#css-extraction)
+- [Rollup + rollup-plugin-vue](https://vuejs.github.io/rollup-plugin-vue/#/en/2.3/?id=custom-handler)
+
+## Tracking Runtime Errors
+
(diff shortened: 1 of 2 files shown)


edison1105 (approved):

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+Most of the tips below are enabled by default if you are using [Vue CLI](https://cli.vuejs.org). This section is only relevant if you are using a custom build setup.
+:::
+
+## Turn on Production Mode
+
+During development, Vue provides a lot of warnings to help you with common errors and pitfalls. However, these warning strings become useless in production and bloat your app's payload size. In addition, some of these warning checks have small runtime costs that can be avoided in production mode.
+
+### Without Build Tools
+
+If you are using the full build, i.e. directly including Vue via a script tag without a build tool, make sure to use the minified version (`vue.min.js`) for production. Both versions can be found in the [Installation guide](installation.html#Direct-lt-script-gt-Include).
```
bencodezen, line 13:
```suggestion
If you are using the full build, i.e. directly including Vue via a script tag without a build tool, make sure to use the minified version for production. Both versions can be found in the [Installation guide](installation.html#Direct-lt-script-gt-Include).
```

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+
+Refer to the respective build tool documentations to see how it's done:
+
+- [Webpack + vue-loader](https://vue-loader.vuejs.org/en/configurations/extract-css.html) (the `vue-cli` webpack template has this pre-configured)
+- [Browserify + vueify](https://github.com/vuejs/vueify#css-extraction)
+- [Rollup + rollup-plugin-vue](https://vuejs.github.io/rollup-plugin-vue/#/en/2.3/?id=custom-handler)
+
+## Tracking Runtime Errors
+
+If a runtime error occurs during a component's render, it will be passed to the global `Vue.config.errorHandler` config function if it has been set. It might be a good idea to leverage this hook together with an error-tracking service like [Sentry](https://sentry.io), which provides [an official integration](https://sentry.io/for/vue/) for Vue.
```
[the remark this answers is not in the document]
bencodezen:
```suggestion
If a runtime error occurs during a component's render, it will be passed to the global `app.config.errorHandler` config function if it has been set. It might be a good idea to leverage this hook together with an error-tracking service like [Sentry](https://sentry.io), which provides [an official integration](https://sentry.io/for/vue/) for Vue.
```

bencodezen:
@phanan @NataliaTepluhina - Would love your review when you have a moment!

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+
+During development, Vue provides a lot of warnings to help you with common errors and pitfalls. However, these warning strings become useless in production and bloat your app's payload size. In addition, some of these warning checks have small runtime costs that can be avoided in production mode.
+
+### Without Build Tools
+
+If you are using the full build, i.e. directly including Vue via a script tag without a build tool, make sure to use the minified version for production. Both versions can be found in the [Installation guide](installation.html#Direct-lt-script-gt-Include).
+
+### With Build Tools
+
+When using a build tool like Webpack or Browserify, the production mode will be determined by `process.env.NODE_ENV` inside Vue's source code, and it will be in development mode by default. Both build tools provide ways to overwrite this variable to enable Vue's production mode, and warnings will be stripped by minifiers during the build. All `vue-cli` templates have these pre-configured for you, but it would be beneficial to know how it is done:
```
skirtles-code, line 17: I'm well out of my depth here but is the terminology *'`vue-cli` templates'* still correct?

Is this referring to the templates concept of Vue CLI 2 (i.e. `vue init`), which no longer exist?

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+      envify({ NODE_ENV: 'production' })
+    )
+    .bundle()
+  ```
+
+- Or, using [envify](https://github.com/hughsk/envify) with Grunt and [grunt-browserify](https://github.com/jmreidy/grunt-browserify):
+
+  ```js
+  // Use the envify custom module to specify environment variables
+  var envify = require('envify/custom')
```
skirtles-code, line 75: A few of these examples are using `var` instead of `const`. Unless there's a specific compatibility issue with `const` I think these should be updated to reflect the conventions used elsewhere in the docs.

skirtles-code (changes requested):
This isn't a subject I know much about but I've provided what little feedback I can.

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
+# Production Deployment
+
+::: info
+Most of the tips below are enabled by default if you are using [Vue CLI](https://cli.vuejs.org). This section is only relevant if you are using a custom build setup.
+:::
+
+## Turn on Production Mode
+
+During development, Vue provides a lot of warnings to help you with common errors and pitfalls. However, these warning strings become useless in production and bloat your app's payload size. In addition, some of these warning checks have small runtime costs that can be avoided in production mode.
```
sdras, line 9: We may want to add a piece here that states what that command would be in the CLI, or if we don't want to keep both updated, provide a link to that section of the CLI docs

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Without Build Tools
+
+If you are using the full build, i.e. directly including Vue via a script tag without a build tool, make sure to use the minified version for production. Both versions can be found in the [Installation guide](installation.html#Direct-lt-script-gt-Include).
```
sdras, line 13:
```suggestion
If you are using the full build, i.e. directly including Vue via a script tag without a build tool, make sure to use the minified version for production. This can be found in the [Installation guide](installation.html#cdn).
```

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+
+In Webpack 4+, you can use the `mode` option:
+
+```js
+module.exports = {
+  mode: 'production'
+}
+```
+
+But in Webpack 3 and earlier, you'll need to use [DefinePlugin](https://webpack.js.org/plugins/define-plugin/):
```
sdras, line 29: Webpack is at v5, we can probably remove the version 3 section, and just double check that v5 works the same way (I think it does) and tighten this a bit.

src/guide/tooling/deployment.md
sdras: Yeah, good point- we used to not use ES6 in the docs but we changed it this version. Thanks!

sdras, line 75:
```suggestion
  const envify = require('envify/custom')
```

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+
+  ```bash
+  NODE_ENV=production browserify -g envify -e main.js | uglifyjs -c -m > build.js
+  ```
+
+- Or, using [envify](https://github.com/hughsk/envify) with Gulp:
+
+  ```js
+  // Use the envify custom module to specify environment variables
+  var envify = require('envify/custom')
```
sdras, line 59:
```suggestion
  const envify = require('envify/custom')
```

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+    })
+  ]
+}).then(...)
+```
+
+## Pre-Compiling Templates
+
+When using in-DOM templates or in-JavaScript template strings, the template-to-render-function compilation is performed on the fly. This is usually fast enough in most cases, but is best avoided if your application is performance-sensitive.
+
+The easiest way to pre-compile templates is using [Single-File Components](single-file-components.html) - the associated build setups automatically performs pre-compilation for you, so the built code contains the already compiled render functions instead of raw template strings.
```
sdras, line 116:
```suggestion
The easiest way to pre-compile templates is using [Single-File Components](single-file-component.html) - the associated build setups automatically performs pre-compilation for you, so the built code contains the already compiled render functions instead of raw template strings.
```

sdras (changes requested):
Minor changes here and there, but really close! @skirtles-code gave a great foundation of the review here

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+
+If you are using Webpack, and prefer separating JavaScript and template files, you can use [vue-template-loader](https://github.com/ktsn/vue-template-loader), which also transforms the template files into JavaScript render functions during the build step.
+
+## Extracting Component CSS
+
+When using Single-File Components, the CSS inside components are injected dynamically as `<style>` tags via JavaScript. This has a small runtime cost, and if you are using server-side rendering it will cause a "flash of unstyled content". Extracting the CSS across all components into the same file will avoid these issues, and also result in better CSS minification and caching.
+
+Refer to the respective build tool documentations to see how it's done:
+
+- [Webpack + vue-loader](https://vue-loader.vuejs.org/en/configurations/extract-css.html) (the `vue-cli` webpack template has this pre-configured)
```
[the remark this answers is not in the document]
sdras: I don't think so actually, because it will default to the latest, so we want to be less specific here. Good eye, though.

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+- [Browserify + vueify](https://github.com/vuejs/vueify#css-extraction)
+- [Rollup + rollup-plugin-vue](https://vuejs.github.io/rollup-plugin-vue/#/en/2.3/?id=custom-handler)
```
sdras, line 128:
```suggestion
- [Rollup + rollup-plugin-vue](https://rollup-plugin-vue.vuejs.org/)
```

src/guide/tooling/deployment.md
```diff
@@ -0,0 +1,132 @@
...
+
+During development, Vue provides a lot of warnings to help you with common errors and pitfalls. However, these warning strings become useless in production and bloat your app's payload size. In addition, some of these warning checks have small runtime costs that can be avoided in production mode.
+
+### Without Build Tools
+
+If you are using the full build, i.e. directly including Vue via a script tag without a build tool, make sure to use the minified version for production. This can be found in the [Installation guide](installation.html#cdn).
+
+### With Build Tools
+
+When using a build tool like Webpack or Browserify, the production mode will be determined by `process.env.NODE_ENV` inside Vue's source code, and it will be in development mode by default. Both build tools provide ways to overwrite this variable to enable Vue's production mode, and warnings will be stripped by minifiers during the build. All `vue-cli` templates have these pre-configured for you, but it would be beneficial to know how it is done:
```
sdras, line 17:
```suggestion
When using a build tool like Webpack or Browserify, the production mode will be determined by `process.env.NODE_ENV` inside Vue's source code, and it will be in development mode by default. Both build tools provide ways to overwrite this variable to enable Vue's production mode, and warnings will be stripped by minifiers during the build. Vue CLI has this pre-configured for you, but it would be beneficial to know how it is done:
```

src/guide/tooling/deployment.md
sdras, line 9:
```suggestion
During development, Vue provides a lot of warnings to help you with common errors and pitfalls. However, these warning strings become useless in production and bloat your app's payload size. In addition, some of these warning checks have small runtime costs that can be avoided in [production mode](https://cli.vuejs.org/guide/mode-and-env.html#modes).
```

src/guide/tooling/deployment.md
sdras: Just double-checked, this is the same in v5

sdras (approved):

sdras:
Great job @bencodezen and @skirtles-code!

PR #416, opened 2020-08-19
fix: MigrationBadges validator

skirtles-code:
The `validator` for `MigrationBadges` is failing, resulting in console warnings in local builds of the docs.

The `value` is an array, not a string, so it needs to use `every` to test the array entries.


NataliaTepluhina (approved):
Thanks for the fix again 😅

PR #417, opened 2020-08-19
fix: computed example should not mutate ref value

skirtles-code:
There are two problems with the existing example:

1. Using `count.value++` changes the value of `count.value`.
2. Despite what it says on the next line, the logged value of `plusOne.value` will be 1, not 2. Note `count.value++` is being used, not `++count.value`.

Changing it to `count.value + 1` fixes both problems.


NataliaTepluhina (approved):
Thanks for the fix! 👍🏻

PR #427, opened 2020-08-24
Add return values for Application API methods

### LINKED ISSUE #423: Return values for Application API methods
The methods in the Application API currently each have a section for `Arguments` but not a section for `Returns`. The returned values are sometimes mentioned in the `Usage` section or in the `Examples` but I think there should be a dedicated section immediately after `Arguments` for `Returns`.

There was something similar in various places in the Vue 2 documentation but most of those methods no longer exist. In the current Vue 3 documentation the only example I could find is for `$watch`.

The returned value is usually the `application`, allowing for chaining, but there are some variations:

* The methods `mixin`, `provide` and `use` just return the current application.
* The methods `component` and `directive` return the current application when two arguments are passed. If only one argument is passed they return the component/directive.
* `mount` returns the root component instance.

I'd be happy to put together a PR for this.

skirtles-code:
Fixes #423.

There didn't seem to be an established convention for what form these should take. I had a look at both the Vue 2 and Vue 3 docs before making some arbitrary choices:

1. No name or type. In other contexts I might have included the type but for these methods I didn't think `{Object}` really added anything.
2. The values are a sub-list within `Returns`, even though they typically only have a single list item.
3. For the complicated cases (`component` and `directive`) I went for two items in the list rather than trying to cram it all into a single item.
4. I used a capital letter at the start of each item but no full-stop at the end. My reasoning was that these aren't proper sentences, they're more like items on a shopping list or a to-do list.

I made a couple of other small changes, both of which are directly related to these return values.

### DIFF
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -19,6 +19,11 @@ In addition, since the `createApp` method returns the application instance itsel
   - `{string} name`
   - `{Function | Object} [definition]`
 
+- **Returns:**
+
+  - The application instance if a `definition` argument was passed
+  - The component definition if a `definition` argument was not passed 
+
 - **Usage:**
 
   Register or retrieve a global component. Registration also automatically sets the component's `name` with the given `name` parameter.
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
Thank you for this contribution @skirtles-code! Changes look good to me 👍🏻

PR #428, opened 2020-08-24
Add in transition migration guide docs

sdras:
This PR should round out the migration guide. Transition docs in, still haven't updated the diagram yet though I don't think it's necessary for this section.

### DIFF
--- a/src/guide/migration/transition.md
+++ b/src/guide/migration/transition.md
@@ -14,7 +14,7 @@ The `v-enter` transition class has been renamed to `v-enter-from` and the `v-lea
 
 Before v2.1.8, we had two transition classes for each transition direction: initial and active states.
 
-In v2.1.8, we introduced `v-enter-to` to address the timing gap between enter/leave transitions, however, for backwards compatibility the v-enter name was untouched:
+In v2.1.8, we introduced `v-enter-to` to address the timing gap between enter/leave transitions. However, for backward compatibility, the `v-enter` name was untouched:
 
 ```css
 .v-enter,
--- a/src/guide/migration/transition.md
+++ b/src/guide/migration/transition.md
@@ -39,6 +39,7 @@ In order to be more explicit and legible, we have now renamed these initial stat
 .v-leave-to {
   opacity: 0;
 }
+
 .v-leave-from,
 .v-enter-to {
   opacity: 1;
--- a/src/guide/migration/transition.md
+++ b/src/guide/migration/transition.md
@@ -46,7 +46,7 @@ In order to be more explicit and legible, we have now renamed these initial stat
 }
 ```
 
-It's now much clearer what the difference between these states are.
+It's now much clearer what the difference between these states is.
 
 The `<transition>` component's related prop names are also changed:
 
--- a/src/guide/migration/transition.md
+++ b/src/guide/migration/transition.md
@@ -12,7 +12,7 @@ The `v-enter` transition class has been renamed to `v-enter-from` and the `v-lea
 
 ## 2.x Syntax
 
-Before v2.1.8, we had two transition classes each transition direction: the initial and active states.
+Before v2.1.8, we had two transition classes for each transition direction: initial and active states.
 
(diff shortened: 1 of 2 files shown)


src/guide/migration/transition.md
```diff
@@ -0,0 +1,59 @@
...
+
+# {{ $frontmatter.title }} <MigrationBadges :badges="$frontmatter.badges" />
+
+## Overview
+
+The `v-enter` transition class has been renamed to `v-enter-from` and the `v-leave` transition class is renamed to `v-leave-from`.
+
+## 2.x Syntax
+
+Before v2.1.8, we had two transition classes each transition direction: the initial and active states.
```
phanan, line 15:
```suggestion
Before v2.1.8, we had two transition classes for each transition direction: initial and active states.
```

src/guide/migration/transition.md
```diff
@@ -0,0 +1,59 @@
...
+
+In v2.1.8, we introduced `v-enter-to` to address the timing gap between enter/leave transitions, however, for backwards compatibility the v-enter name was untouched:
```
phanan, line 17:
```suggestion
In v2.1.8, we introduced `v-enter-to` to address the timing gap between enter/leave transitions. However, for backward compatibility, the `v-enter` name was untouched:
```

src/guide/migration/transition.md
```diff
@@ -0,0 +1,59 @@
...
+
+```css
+.v-enter-from,
+.v-leave-to {
+  opacity: 0;
+}
+.v-leave-from,
+.v-enter-to {
+  opacity: 1;
+}
```
phanan, line 45:
```suggestion
.v-enter-from,
.v-leave-to {
  opacity: 0;
}

.v-leave-from,
.v-enter-to {
  opacity: 1;
}
```

src/guide/migration/transition.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+It's now much clearer what the difference between these states are.
```
phanan, line 48:
```suggestion
It's now much clearer what the difference between these states is.
```

src/guide/migration/transition.md
```diff
@@ -0,0 +1,59 @@
...
+
+The `<transition>` component's related prop names are also changed:
+
+- `leave-class` is renamed to `leave-from-class` (in render functions or JSX, can be written as `leaveFromClass`)
+- `enter-class` is renamed to `enter-from-class` (in render functions or JSX, can be written as `enterFromClass`)
```
phanan, line 53:
```suggestion
- `leave-class` is renamed to `leave-from-class` (can be written as `leaveFromClass` in render functions or JSX)
- `enter-class` is renamed to `enter-from-class` (can be written as `enterFromClass` in render functions or JSX)
```

src/guide/migration/transition.md
```diff
@@ -0,0 +1,59 @@
...
+title: Transition Class Change
+badges:
+  - breaking
+---
+
+# {{ $frontmatter.title }} <MigrationBadges :badges="$frontmatter.badges" />
+
+## Overview
+
+The `v-enter` transition class has been renamed to `v-enter-from` and the `v-leave` transition class is renamed to `v-leave-from`.
```
phanan, line 11:
```suggestion
The `v-enter` transition class has been renamed to `v-enter-from` and the `v-leave` transition class to `v-leave-from`.
```

phanan (commented):
Nice! I've left a couple of suggestions, none of which should be blocking.

src/guide/migration/transition.md
bencodezen:
```suggestion
The `v-enter` transition class has been renamed to `v-enter-from` and the `v-leave` transition class has been renamed to `v-leave-from`.
```

bencodezen: This might be more consistent since without it "the v-leave transition class to v-leave-from" might be a little confusing for non-native English speakers.

sdras:
Thank you!

PR #432, opened 2020-08-26
isHydrate is optional

skirtles-code:
The `isHydrate` argument of `mount` is optional so I've added square brackets.

### DIFF
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -181,7 +181,7 @@ Apart from `el`, you should treat these arguments as read-only and never modify 
 - **Arguments:**
 
   - `{Element | string} rootContainer`
-  - `{boolean} isHydrate`
+  - `{boolean} [isHydrate]`
 
 - **Returns:**
 


src/api/application-api.md
```diff
@@ -181,7 +181,7 @@ Apart from `el`, you should treat these arguments as read-only and never modify
 - **Arguments:**
 
   - `{Element | string} rootContainer`
-  - `{boolean} isHydrate`
+  - `{boolean} [isHydrate]`
```
NataliaTepluhina, line 184: **suggestion**: could we replace square brackets here with the note about optional component? Like we have on Directives

```
**Argument:** `attrOrProp (optional)`
```

skirtles-code: That would be inconsistent with the other optional parameters on this page, so I have created #442 to address that.

If you want to proceed with the `(optional)` format then this PR can be closed in deference to #442.

PR #434, opened 2020-08-27
fix: remove unneeded `toRefs` imports

yannbertrand:
Some imports are unused in examples


NataliaTepluhina (approved):
@yannbertrand thank you for fixing this! 👍🏻

PR #441, opened 2020-08-31
docs (#386): add mobile section

### LINKED ISSUE #386: Add Mobile section to the docs
Given that mobile solutions are a popular request, it would be good to provide some guidance in this regard.

Things that come to mind include:

- NativeScript
- Capacitor
- Vue Native

bencodezen:
This was created as a way to allow users to discover great solutions being provided and actively supported by the community. Let me know what you think!

@mhartington @rigor789 Please feel free to let me know if you would like to add or change anything about your project's description and/or resources.

Netlify Preview: https://5f4d0ec6a42f5400070c3564--vue-docs-next-preview.netlify.app/guide/mobile.html

Closes #386

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -1,10 +1,13 @@
 const sidebar = {
-  cookbook: [{
-    title: 'Cookbook',
-    collapsable: false,
-    children: ['/cookbook/', '/cookbook/editable-svg-icons']
-  }],
-  guide: [{
+  cookbook: [
+    {
+      title: 'Cookbook',
+      collapsable: false,
+      children: ['/cookbook/', '/cookbook/editable-svg-icons']
+    }
+  ],
+  guide: [
+    {
       title: 'Essentials',
       collapsable: false,
       children: [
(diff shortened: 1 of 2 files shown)


phanan:
Generally, I like the idea 👍 but maybe we want to make sure it doesn't overlap with the "Comparison with other frameworks" page (not yet published). For reference, this is how it currently looks [in v2](https://vuejs.org/v2/guide/comparison.html#Native-Rendering):

> On this front, Vue has an official collaboration with Weex, a cross-platform UI framework created by Alibaba Group and being incubated by the Apache Software Foundation (ASF). Weex allows you to use the same Vue component syntax to author components that can not only be rendered in the browser, but also natively on iOS and Android!

> At this moment, Weex is still in active development and is not as mature and battle-tested as React Native, but its development is driven by the production needs of the largest e-commerce business in the world, and the Vue team will also actively collaborate with the Weex team to ensure a smooth experience for Vue developers.

> Another option is NativeScript-Vue, a NativeScript plugin for building truly native applications using Vue.js.

bencodezen:
@phanan Agreed. Since it's something that is separate from a framework comparison point, I think it'd be good to strip that out of that section and let it be its own section since I imagine we'll get more dedicated resources on this when Vue 3 releases.

bencodezen:
Pinging @phanan just to check on what your thoughts are as far as merging this for now.

phanan:
LGTM 👍 Thanks for working on this!

PR #442, opened 2020-08-31
Indicate optional arguments with the word 'optional'

skirtles-code:
This is in response to feedback on #432.

In that PR I attempted to put square brackets around the `isHydrate` argument of `mount` to mark it as optional. It was suggested that we should use the word `(optional)` instead, but that would have been inconsistent with other parts of the same page.

This PR changes all uses of square brackets to the `(optional)` format. It includes `isHydrate`, so #432 can be closed if this is merged.

### DIFF
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -17,7 +17,7 @@ In addition, since the `createApp` method returns the application instance itsel
 - **Arguments:**
 
   - `{string} name`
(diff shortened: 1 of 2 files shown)


NataliaTepluhina (approved):
@skirtles-code thank you so much for fixing it! 🙇🏻‍♀️

PR #443, opened 2020-09-01
fix: watchEffect api links

euaaaio:


NataliaTepluhina (approved):
@euaaaio thank you for the fixes!

PR #450, opened 2020-09-04
docs: fix argument name

edison1105:
the first argument name is `type` not `tag` of function `h`.

### DIFF
--- a/src/api/global-api.md
+++ b/src/api/global-api.md
@@ -71,9 +71,9 @@ render() {
 
 ### Arguments
 
-Accepts three arguments: `tag`, `props` and `children`
+Accepts three arguments: `type`, `props` and `children`
 
-#### tag
+#### type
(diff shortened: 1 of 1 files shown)


bencodezen:
Hi @edison1105! Is there a particular reason you believe it should be named `type` rather than `tag`?

bencodezen (changes requested):
Hi @edison1105! Is there a particular reason you believe it should be named type rather than tag?

edison1105:
> Hi @edison1105! Is there a particular reason you believe it should be named type rather than tag?

because the parameter name is “type” in the source code of vue-next.In vue2，its name is “tag”.so，I think we should rename it to “type”.

bencodezen (approved):
Ah found it! Thank you for the PR. Appreciate it!

PR #455, opened 2020-09-05
add usage that pass setup function to defineComponent

unbyte:


src/api/global-api.md
```diff
@@ -130,6 +130,17 @@ const MyComponent = defineComponent({
 })
 ```
 
+Or a `setup` function, function name will be used as component name
+
+```js
+import { defineComponent, ref } from 'vue'
+
+const HelloWorld = defineComponent(function HelloWorld() {
```
NataliaTepluhina, line 138: **nitpick**: could we change the syntax here to ES6?

```suggestion
const HelloWorld = defineComponent(HelloWorld() => {
```

unbyte: However, it's not a valid js(ts) syntax 😂. `function` is used to keep the document concise, because if use arrow function, code in document should be like 
```js
const HelloWorld = () => {
    ....
}

// since `HelloWorld` is already defined in the scope, use `HelloWorldComponent` instead
const HelloWorldComponent = defineComponent(HelloWorld)
```

NataliaTepluhina: Sorry, I didn't pay enough attention when was writing the suggestion 🤦🏻 You're right

PR #458, opened 2020-09-05
fix: Change the options object - defineAsyncComponent

zhenzhenChange:
fix: Change the options object: remove the `retryWhen`, `maxRetries` attributes, add the `onError` event, and attach a comment


NataliaTepluhina (approved):
@zhenzhenChange thank you for the fix!

PR #459, opened 2020-09-07
docs: add transitions diagram

sdras:
Redid the transitions diagram for 3.x

- Created a new illustrator file and committed to assets
- Updated the diagram in the transitions-enterelave section
- Added the diagram to the migration section to be more clear


sdras:
I think it's failing because of the .AI file, I'll compress it and see if that helps

sdras:
better :)

bencodezen (approved):
Thanks so much for doing this! Glad to have the raw design file committed now so future changes are easier!

PR #462, opened 2020-09-09
Update instance-methods.md

pasBone:
One more character in the submenu  "\".

### DIFF
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -1,6 +1,6 @@
 # Instance Methods
 
-## \$watch
+## $watch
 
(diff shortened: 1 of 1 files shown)


bencodezen:
This is necessary to escape the $ character, so we will need to leave it in at this time. Thank you for the PR though!

bencodezen:
@pasBone Also, just saw the issue as far as the sidebar. Would you be interested in [creating a VuePress issue](https://github.com/vuejs/vuepress/issues/new?template=bug_report.md)? Wanted to give you credit for finding this, but no worries if you'd rather not.

PR #464, opened 2020-09-10
fix: render-function, transitions-enterleave.md

ajycc20:
fix the hyperlink of `this.$slots`
fix the hyperlink of 'dynamic component'


sdras (approved):
Thank you!

PR #475, opened 2020-09-15
fix: fix the `onMounted` hooks

ajycc20:


src/guide/reactivity-computed-watchers.md
```diff
@@ -122,7 +122,7 @@ In this example:
 - The count will be logged synchronously on initial run.
 - When `count` is mutated, the callback will be called **after** the component has updated.
 
-Note the first run is executed before the component is mounted. So if you wish to access the DOM (or template refs) in a watched effect, do it in the mounted hook:
+Note the first run is executed before the component is `mounted`. So if you wish to access the DOM (or template refs) in a watched effect, do it in the `onMounted` hook:
```
NataliaTepluhina, line 125: **nitpick**: Thank you for the fix! Looks great to me but I would remove the first backticks here:

```suggestion
Note the first run is executed before the component is mounted. So if you wish to access the DOM (or template refs) in a watched effect, do it in the `onMounted` hook:
```

NataliaTepluhina (changes requested):

src/guide/reactivity-computed-watchers.md
ajycc20: the first backticks has removed

NataliaTepluhina (approved):

PR #480, opened 2020-09-15
change document version from char to npm svg file, keep be consistent…

awkj:
… with vue-next repo


NataliaTepluhina (approved):
Thank you @awkj, I really like this change! 👍🏻

PR #481, opened 2020-09-16
Migration guide section on key attribute changes

### LINKED ISSUE #379: Migration guide section about v-if key
**Edit**: It seems like the behavior has changed on the master branch since I posted this yesterday, so the following description is out of date. I think it's still a breaking change, so the migration guide might still need a section on this. However, it might be good to wait a little bit before working on that, since it seems possible that the behavior could change again before the next release.

***

I can't find anything about this in the migration guide, but it seems to be a breaking change.

### Vue 2
In v2, it was common to use `key`s on `v-if`, `v-else`, and `v-else-if`. For example:
```vue
<h1 v-if="cond" key="a">A</h1>
<h2 v-else key="b">B</h2>
```

### Vue 3
In v3, this produces an compiler error because `key`s are **automatically generated** on `v-if`, `v-else`, and `v-else-if` nodes. These keys can be removed:
```vue
<h1 v-if="cond">A</h1>
<h2 v-else>B</h2>
```

### Vue 2
In v2, a `<template>` could not have a `key`, so when using a `<template v-for="...">` with a child `v-if`/`v-else`, you would place the `key` on both of the children:
```vue
<template v-for="thing in list">
  <h1 v-if="cond" :key="thing.id">A</h1>
  <h2 v-else :key="thing.id">B</h2>
</template>
```

### Vue 3
In v3, the previous example causes the `v-if` to not work properly, and it produces an error message. Because of Fragments, we can now put the `key` on the `<template>` directly:
```vue
<template v-for="thing in list" :key="thing.id">
  <h1 v-if="cond">A</h1>
  <h2 v-else>B</h2>
</template>
```

## Error message
This is the [new error message](https://github.com/vuejs/vue-next/blob/11ed2103919af46b5d9f573d3fad749ca7eb73f5/packages/compiler-core/src/errors.ts#L139-L142) for this problem in the master branch of vue-next. (This new message isn't released into a new version yet.)
> `v-if branches must use compiler generated keys. In many cases, you can simply remove this key. If this tag is inside of a <template v-for="...">, then you can move the key up to the parent <template>.`

## References
* https://github.com/vuejs/vue-next/issues/1734#issuecomment-665745971
* https://github.com/vuejs/vue-next/issues/1712#issuecomment-665206103

Aurelius333:
## Description of Problem

This is my attempt to fix #379 (a couple of undocumented breaking changes in Vue 3 related to the `key` attribute).

## Proposed Solution

I think this is technically two separate breaking changes:

* If you manually provide `key`s on `v-if` branches, then each branch must use a unique `key`.
* `<template v-for>` `key` should be placed on the `<template>` tag (rather than on its children).

However, I think it's probably best to keep them on the same page because they're related and often appear at the same time. For example, they appear together in this relatively common situation:

 ```html
<!-- Vue 2.x -->
<template v-for="item in list">
  <div v-if="item.isVisible" :key="item.id">...</div>
  <span v-else :key="item.id">...</span>
</template>

<!-- Vue 3.x -->
<template v-for="item in list" :key="item.id">
  <div v-if="item.isVisible">...</div>
  <span v-else>...</span>
</template>
``` 

## Additional Information

In my opinion, `key`s are one of the more confusing parts of Vue. That's why I included a "Background" section near the top that explains what the `key` attribute is. I'm not really sure whether this section should be there or not, so I'm totally fine with removing the "Background" if people think it's unnecessary.

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -1,13 +1,10 @@
 const sidebar = {
-  cookbook: [
-    {
-      title: 'Cookbook',
-      collapsable: false,
-      children: ['/cookbook/', '/cookbook/editable-svg-icons']
-    }
-  ],
-  guide: [
-    {
+  cookbook: [{
+    title: 'Cookbook',
+    collapsable: false,
+    children: ['/cookbook/', '/cookbook/editable-svg-icons']
+  }],
+  guide: [{
       title: 'Essentials',
       collapsable: false,
       children: [
@@ -63,8 +60,7 @@ const sidebar = {
     {
       title: 'Advanced Guides',
       collapsable: false,
-      children: [
-        {
+      children: [{
           title: 'Reactivity',
           children: [
             '/guide/reactivity',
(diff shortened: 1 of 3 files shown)


NataliaTepluhina (approved):
Thank you @Aurelius333! This is a great addition to migration guide 👍🏻

PR #483, opened 2020-09-16
fix: sample grid-component is error

57code:
## Description of Problem
the sample grid-component is Elastic Header Example


NataliaTepluhina:
Thank you for your contribution! We prefer to keep all the codepens on Vue CodePen account, please let me know what exact changes would you prefer to add to existing codepen?

skirtles-code:
I believe this has been fixed by #551.

PR #487, opened 2020-09-16
Note that props can be accessed in template directly

garyo:
In "Usage with Templates" the doc shows how props of the object returned from `setup` can be used directly in the template, but I didn't find anywhere else in the composition-api doc where it says that props can also be used directly. This seemed like as good a place as any to add that, since that's where I was looking for it first :-).

## Description of Problem
Composition API doc doesn't mention that props can be used directly in templates

## Proposed Solution
I added a note and changed the example to use props.

### DIFF
--- a/src/guide/composition-api-setup.md
+++ b/src/guide/composition-api-setup.md
@@ -99,7 +99,7 @@ In other words, you **will not have access** to the following component options:
 
 ## Usage with Templates
 
-If `setup` returns an object, the properties on the object can be accessed in the component's template, as can the properties of the `props` passed in to `setup`:
+If `setup` returns an object, the properties on the object can be accessed in the component's template, as well as the properties of the `props` passed into `setup`:
 
 ```vue-html
 <!-- MyBook.vue -->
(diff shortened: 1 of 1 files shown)


src/guide/composition-api-setup.md
```diff
@@ -99,19 +99,19 @@ In other words, you **will not have access** to the following component options:
 
 ## Usage with Templates
 
-If `setup` returns an object, the properties on the object can be accessed in the component's template:
+If `setup` returns an object, the properties on the object can be accessed in the component's template, as can the properties of the `props` passed in to `setup`:
```
NataliaTepluhina, line 102: **nitpick**: small wording change
```suggestion
If `setup` returns an object, the properties on the object can be accessed in the component's template, as well as the properties of the `props` passed into `setup`:
```

NataliaTepluhina (commented):
@garyo thank you for the contribution! I've added a small suggestion there

garyo:
> @garyo thank you for the contribution! I've added a small suggestion there

It's OK with me; I struggled with that for a bit because I don't want to imply that you get access to props in the template _only if_ setup returns an object. But I agree that "as can" is a bit archaic.

NataliaTepluhina (approved):
@garyo thank you! Happy to merge the change 🎉

PR #492, opened 2020-09-18
fix: introduce the wrapper object to the props default example

skirtles-code:
## Description of Problem

The example of using `default` for `props` in the migration guide appears to be missing a wrapper object and `default` property.

I don't think the code as currently written will actually work. Testing with rc.13 seemed to need the extra object.

## Proposed Solution

Added the wrapper object.


NataliaTepluhina (approved):
@skirtles-code thanks!

PR #503, opened 2020-09-19
Fixes "Hook inside `setup`" grammar

cathrinevaage:
## Description of Problem
The column header for life-cycle hook equivalents in the setup context was called "Hook inside inside `setup`"

## Proposed Solution
One "inside" is enough.


NataliaTepluhina (approved):
@cathrinevaage thank you for the fix!

PR #504, opened 2020-09-19
fix: component-slots, typescript-support.md

ajycc20:
close `<span>` tags
fix 'Typing `ref` '

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -194,7 +194,7 @@ const Component = defineComponent({
 })
 ```
 
-### Typing `ref`
+### Typing `ref`s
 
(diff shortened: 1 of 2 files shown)


src/guide/typescript-support.md
```diff
@@ -194,7 +194,7 @@ const Component = defineComponent({
 })
 ```
 
-### Typing `ref`s
```
NataliaTepluhina, line 197: **note**: this is an intention, we're speaking about multiple refs here

NataliaTepluhina (changes requested):
@ajycc20 thank you for closing span tags! For refs, this was an intentional multiple form

PR #507, opened 2020-09-19
Resolve props template usage composition api

xi-osman:
## Description of Problem
Was not able to run the provided composition template code snippet and realized its because the props object was not returned.
## Proposed Solution
Added the props property and also returned props so the template can reference it.

### DIFF
--- a/src/guide/composition-api-setup.md
+++ b/src/guide/composition-api-setup.md
@@ -111,6 +111,9 @@ If `setup` returns an object, the properties on the object can be accessed in th
   import { ref, reactive } from 'vue'
 
   export default {
+    props: {
+      collectionName: String
+    },
     setup(props) {
       const readersNumber = ref(0)
       const book = reactive({ title: 'Vue 3 Guide' })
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@xi-osman thank you for the contribution! However, the snippet works without returning `props` explicitly: https://codesandbox.io/s/setup-with-props-phsye?file=/src/Book.vue

xi-osman:
@NataliaTepluhina Awesome that worked! Would we want to remove the 'props.' from the template and just leave 'collectionName' then?  

```
<!-- MyBook.vue -->
<template>
  <div>{{ props.collectionName }}: {{ readersNumber }} {{ book.title }}</div>
</template>
```

NataliaTepluhina:
@xi-osman yes, that is a mistake I guess 😅  Would you mind doing this change on your pull request?

xi-osman:
@NataliaTepluhina Sure! already submitted a new commit 😎

NataliaTepluhina (approved):
Thank you for fixing it!

PR #508, opened 2020-09-19
fix: change to modern bundler

kazupon:
## Description of Problem
Browserify is currently almost underused in the Vue ecosystem.
In fact, it doesn't even work with [vueify](https://github.com/vuejs/vueify) in Vue 3.
vueify is arcvhied.

It has been arcvhied. 

## Proposed Solution
We recommend switching to the current mainstream bundler rollup
As Vue official, we offer a rollup-plugin-vue.
rollup-plugin-vue is available for Rollup as Vue official.
https://github.com/vuejs/rollup-plugin-vue

### DIFF
--- a/src/guide/installation.md
+++ b/src/guide/installation.md
@@ -30,7 +30,7 @@ For production, we recommend linking to a specific version number and build to a
 
 ## NPM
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
@kazupon thank you! I agree, mentioning Rollup makes way more sense than Browserify

PR #514, opened 2020-09-20
Rewrite of instance.md, introducing data-methods.md

skirtles-code:
This could take a while...

## Introduction

This is intended as a discussion starter, not necessarily something that should be merged. I could have just opened an issue but I don't think I could have properly conveyed what I wanted to say. Rather than just criticising the current version of `instance.md` I've tried to fix the problems myself. Opinions may vary as to whether I've succeeded.

~To be clear, this definitely shouldn't be merged in its present form. At the very least I'll need to go through other pages to update any links. I won't bother doing that unless there's a realistic chance of it being merged.~ *Update: I'm now happy for it to be merged.*

I've gone through and added inline review comments to this PR. I think they're necessary but they do make reading the finished files a bit difficult. I suggest reading the files without my comments first to get an overall feel for the finished result.

The diff isn't necessarily super helpful. While my version does borrow some small passages from the current version it's mostly a rewrite. Some of the structure and narrative are similar to the previous version but the wording is mostly new. The exceptions are the two sections about the lifecycle, which are almost unchanged.

The relevant pages are available at:

* https://deploy-preview-514--vue-docs-next-preview.netlify.app/guide/instance.html
* https://deploy-preview-514--vue-docs-next-preview.netlify.app/guide/data-methods.html

## The Problem

`instance.md` has evolved over many years. But what is it actually about? What are we trying to explain and why?

I feel like it's been tweaked and patched so many times that the original narrative has become somewhat blurred.

Vue 3 then introduced new problems. The concept of the 'Vue instance' doesn't really exist anymore. The closest match is a 'component instance', which behaves much the same as a `vm` in Vue 2.

On its own that wouldn't be a problem but the introduction of application instances makes it much, much more difficult to explain what's going on. Trying to write an explanation that navigates through all the potential misunderstandings without getting lost in the weeds is not at all easy.

In my opinion, the current version doesn't make a clear enough distinction between application instances and component instances. The term 'component instance' isn't even used but I think it's unavoidable to qualify the word 'instance' in some way. I don't think there's any reason to avoid the word 'component', the basic concept having been introduced in the guide's **Introduction**.

## My Attempted Solution

I've had a go at writing my own explanation of these topics. I've also moved some material out to a new page, called `data-methods.md`.

Whether this is an improvement over the current documentation is subjective. I prefer it (as a totally impartial judge). Whether you agree or not, hopefully it'll help as a discussion starter to find the right balance.

Please feel free to focus any feedback on the bits you don't like. No sugar coating required.


NataliaTepluhina:
@skirtles-code thank you for bringing this! From a quick glance, it makes a lot of sense to me, will look in detail today

src/.vuepress/config.js
```diff
@@ -11,6 +11,7 @@ const sidebar = {
         '/guide/installation',
         '/guide/introduction',
         '/guide/instance',
+        '/guide/data-methods',
```
NataliaTepluhina, line 14: **nitpick**: Could we move data-methods after template syntax so it will go before computed and watchers section?

src/guide/data-methods.md
```diff
@@ -0,0 +1,137 @@
...
+// so the `this` value inside the function will still
+// refer to the correct component instance
+inc()
+
+console.log(vm.count) // => 5
+```
+
+Just like all other properties of the component instance, the `methods` are accessible from within the component's template. If a method is called by a component's template, any reactive data it uses will be tracked and added as a rendering dependency.
+
+### Debouncing and Throttling
```
NataliaTepluhina, line 98: **suggestion**: I would recommend moving this section to `Examples`. We try to reduce the cognitive load (especially on **Essentials** part and here we
1. Introduce a new concept (debouncing)
2. Introduce a new library (lodash)
3. Neither of these is directly connected to Vue

src/guide/instance.md
```diff
@@ -31,109 +57,75 @@ Root Instance
...
-  }
-}).mount('#example')
+vm.increment()
 
-vm.$data.a // => 1
+console.log(vm.count) // => 5
 ```
 
-In the future, you can consult the [API reference](../api/instance-properties.html) for a full list of instance properties and methods.
+Inside the methods of a component, the `this` value will be bound to the component instance.
```
NataliaTepluhina, line 109: **opinion**: this feels like we're repeating `Data and methods` part. Should we just make a list of possible component instance properties here without diving in explanation, or should we exclude this section altogether?

NataliaTepluhina (commented):
First of all, thank you for the huge effort that you put into this rewrite, @skirtles-code! It's a very thoughtful addition and I believe we should merge it.

I've added a few notes and nitpicks for your consideration, would be happy to discuss!

skirtles-code (commented):
@NataliaTepluhina Sorry, I've made a right mess of this review process.

When I first created this PR I went through and carefully annotated my changes to explain exactly what I'd done and why. I didn't want anyone to have to try to review it without a proper explanation from me.

However, I now realise that I'm the only person who could see those comments. Oops.

I'm going to click the big green button and hopefully that'll make my original comments visible to everybody. I stress that these were all written prior to your feedback, which I will now go through and address individually.

Sorry about that.

src/guide/instance.md
skirtles-code: Good point.

I'll need to mull this over. I don't think I'd want to remove it altogether but it might be possible to cut it back significantly to the bits that are most relevant to the topic at hand.

I need to check the surrounding pages to ensure the overall narrative still holds together, especially if we're going to move `data-methods.md` to be after `template-syntax.md`.

src/.vuepress/config.js
skirtles-code: Done.

I don't feel strongly either way about this. Neither page leans heavily on the other so they could go in either order.

I've made some small tweaks to `data-methods.md` and `template.md` that seemed appropriate given the new order.

src/guide/instance.md
skirtles-code: I've cut a significant portion from that section. I think it's much better.

src/guide/data-methods.md
skirtles-code: Yeah, I had expected this to be controversial and tried to explain my reasoning when I first created the PR. Shame I failed to make the comments visible but the relevant one is here:

https://github.com/vuejs/docs-next/pull/514/files/952f8cba80365431108960ee9c957dca55a36f47#r491641224

In addition to what I wrote there:

1. Having it in the examples section wouldn't achieve everything I was trying to achieve. (See my earlier comment)
2. There is already something in the examples section that uses lodash for debouncing. https://v3.vuejs.org/examples/markdown.html. Unfortunately, because it's such a simple example it uses the 'cheap and cheerful' approach to debouncing that won't actually work with reusable components. Constructing a full example that 'does it properly' is actually quite tricky but that current example does somewhat lead people astray without anything to warn them.
3. The Vue 2 docs did have another debouncing example: https://vuejs.org/v2/guide/computed.html#Watchers. It seems that's been tweaked to remove the debouncing aspect in Vue 3.
4. I suspect that for most intermediate JS developers, debouncing would not be a new concept. One reason I also included the term 'throttling' was to jog the memories of those who had forgotten what debouncing was. For those who still don't know what debouncing is, I tried to word the opening sentence in such a way that it implied it could be skipped.

Specifically on the topic of debouncing itself, it is something that gets asked about in the context of Vue on a very regular basis. Just today someone was asking about it on SO: https://stackoverflow.com/questions/64026366. A quick search for questions mentioning Vue and debouncing throws up loads of other examples. One of the most popular Vue questions is https://stackoverflow.com/questions/42199956, which does have all the information lurking in the answers but you really have to dig through a lot of weak suggestions to find the gold.

All my waffling aside, the TLDR is that I'd like to keep that example where I currently have it. If you really want it gone then I will, of course, remove it.

src/guide/instance.md
```diff
@@ -31,109 +57,75 @@ Root Instance
       └─ TodoListStatistics
 ```
 
-We'll talk about [the component system](component-basics.html) in detail later. For now, just know that all Vue components are also instances, and so accept the same options object.
+Each component will have its own component instance, `vm`. For some components, such as `TodoItem`, there will likely be multiple instances rendered at any one time. All of the component instances in this application will share the same application instance.
 
-## Data and Methods
```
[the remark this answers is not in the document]
skirtles-code: In response to feedback I've trimmed out the parts about methods. I think it gets to the point much better now.

src/guide/instance.md
```diff
@@ -31,109 +57,75 @@ Root Instance
...
-    newTodoText: '',
-    visitCount: 0,
-    hideCompletedTodos: false,
-    todos: [],
-    error: null
-  }
-}
-```
-
-The only exception to this being the use of `Object.freeze()`, which prevents existing properties from being changed, which also means the reactivity system can't _track_ changes.
```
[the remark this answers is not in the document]
skirtles-code: I've looked into this a bit more and with hindsight I don't think there's any real need to mention `Object.freeze` in the documentation for Vue 3.

The performance benefits of `Object.freeze` are less clear cut in Vue 3. The new `markRaw` seems a more official way to achieve a similar effect.

`Object.freeze` is not alone in its special treatment. `Object.seal` and `Object.preventExtensions` are handled the same way. I don't think it's anything worth shouting about in the docs.

src/guide/instance.md
```diff
@@ -31,109 +57,75 @@ Root Instance
...
-
-const vm = Vue.createApp({
+const app = Vue.createApp({
   data() {
-    return obj
+    return { count: 4 }
+  },
+  methods: {
+    increment() {
+      // `this` refers to the same object as vm
```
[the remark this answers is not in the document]
skirtles-code: This example been lost in the latest changes but there is a similar point made in the next section about lifecycle hooks. It also gets revisited in `data-methods.md`.

src/guide/template-syntax.md
```diff
@@ -16,7 +16,7 @@ The most basic form of data binding is text interpolation using the "Mustache" s
 <span>Message: {{ msg }}</span>
 ```
 
-The mustache tag will be replaced with the value of the `msg` property on the corresponding data object. It will also be updated whenever the data object's `msg` property changes.
+The mustache tag will be replaced with the value of the `msg` property from the corresponding component instance. It will also be updated whenever the `msg` property changes.
```
skirtles-code, line 19: I tweaked this because:

1. We haven't gone into detail about `data` by this point in the guide.
2. It could just as easily be a `computed` property or some other form of property. Given what's in `instance.md` I think this wording is fine.

skirtles-code:
@NataliaTepluhina As far as I'm concerned this is now good to be merged.

I haven't marked the conversations as Resolved as I suspect that'll make them harder to read but I can go through and mark all mine as resolved if you want.

Let me know if you'd prefer further changes.

src/.vuepress/config.js
NataliaTepluhina: Thank you! 👍🏻 Looks great now

src/guide/data-methods.md
```diff
@@ -0,0 +1,137 @@
+# Data Properties and Methods
+
+## Data Properties
+
+The `data` function for a component is called as part of creating a new component instance. It should return an object, which Vue wraps in its reactivity system and stores on the instance as `$data`. For convenience, any top-level properties of that object are also exposed directly via the component instance:
+
+```js
+const app = Vue.createApp({
+  data() {
+    return { count: 4 }
```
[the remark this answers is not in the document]
NataliaTepluhina: Sounds like valid points to me. The only issue might be that we plan to add a linter to code snippets and this one may be automatically formatted to 'standard' view (which is 3 lines)

src/guide/data-methods.md
```diff
@@ -0,0 +1,137 @@
...
+
+You might be tempted to put functions in your `data` so that you can use them as methods:
+
+```js
+const app = Vue.createApp({
+  data() {
+    return {
+      count: 4,
+
+      // Don't do this
```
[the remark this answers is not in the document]
NataliaTepluhina: Agreed. People tend to jump from once code snippet to another so it's good to have code comments repeating the logic of the explanation

src/guide/data-methods.md
```diff
@@ -0,0 +1,137 @@
...
+inc()
+
+console.log(vm.count) // => 5
+```
+
+Just like all other properties of the component instance, the `methods` are accessible from within the component's template. If a method is called by a component's template, any reactive data it uses will be tracked and added as a rendering dependency.
+
+### Debouncing and Throttling
+
+Vue doesn't include built-in support for debouncing or throttling but it can be implemented using libraries such as lodash.
```
NataliaTepluhina, line 100:
```suggestion
Vue doesn't include built-in support for debouncing or throttling but it can be implemented using libraries such as [lodash](https://lodash.com/).
```

src/guide/data-methods.md
```diff
@@ -0,0 +1,137 @@
...
+
+In cases where a component is only used once, the debouncing can be applied directly within `methods`:
+
+```js
+const app = Vue.createApp({
```
NataliaTepluhina, line 105: We need to show importing `_` from lodash in this code snippet as it's not completely clear to people without lodash experience what is an underscore here

NataliaTepluhina (commented):
Incredible work, thank you so much @skirtles-code for all the detailed explanations you added to your changes! I have a few additional minor tweaks but in general changes look good to me 👍🏻

src/guide/data-methods.md
```diff
@@ -0,0 +1,137 @@
...
+
+These instance properties are only added when the instance is first created, so you need to ensure they are all present in the object returned by the `data` function. Where necessary, use `null`, `undefined` or some other placeholder value for properties where the desired value isn't yet available.
+
+It is possible to add a new property directly to the component instance without including it in `data`. However, because this property isn't backed by the reactive `$data` object it won't automatically be tracked by [Vue's reactivity system](reactivity.html).
+
+Vue uses a `$` prefix when exposing its own built-in APIs via the component instance, e.g. `$emit`. It also reserves the prefix `_` for internal properties. You should avoid using names for top-level `data` properties that start with either of these characters.
+
+## Methods
+
+You might be tempted to put functions in your `data` so that you can use them as methods:
```
phanan, line 36: I've never encountered anyone who does this (putting functions in `data` as methods), no matter how much of a beginner they are so IMHO this might not be the best approach to introduce `methods`.

skirtles-code: Yeah, I made a similar observation here: https://github.com/vuejs/docs-next/pull/514/files/932839b6f612d79f299ab7e45604e0188d327e96#r491639058

I'll give it some more thought. I probably just need to delete most of that introduction and get to the point.

src/guide/data-methods.md
```diff
@@ -0,0 +1,137 @@
+# Data Properties and Methods
+
+## Data Properties
+
+The `data` function for a component is called as part of creating a new component instance. It should return an object, which Vue wraps in its reactivity system and stores on the instance as `$data`. For convenience, any top-level properties of that object are also exposed directly via the component instance:
```
bencodezen, line 5:
```suggestion
The `data` property for a component is called as part of creating a new component instance. It is a function that should return an object, which Vue will then wrap in its reactivity system and store on the component instance as `$data`. For convenience, any top-level properties of that object are also exposed directly via the component instance:
```

bencodezen: Suggested edits are motivated by people not as familiar with JS, so calling it a property would be easier to follow initially, and then defining it technically should allow them that gradual ease into the concepts for data.

src/guide/data-methods.md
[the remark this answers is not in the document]
bencodezen: Agreed. I'm all for removing single letter or nonsense words as variable names in documentation.

src/guide/data-methods.md
bencodezen: While this is theoretically possible, I'm think this would introduce an anti-pattern and may be more harmful than helpful. However, I do think this would make an interesting blog post!

bencodezen (changes requested):
Thanks for the detailed work on this @skirtles-code! Looking forward to giving this another look once the changes and discussion have been resolved!

src/guide/data-methods.md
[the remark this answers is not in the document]
skirtles-code: This has been completely rewritten.

src/guide/data-methods.md
skirtles-code: Done. I've also changed Lodash to a capital L, as that seems to be their preferred convention.

src/guide/data-methods.md
skirtles-code: I've updated this example to try to address this. It isn't quite as straightforward as it sounded because (I think) it needs to use a CDN version of Lodash for consistency with the other examples. That leaves the `_` not explicitly defined. I've added a comment in the code to try to clarify a bit more.

src/guide/data-methods.md
skirtles-code: I've rewritten the **Methods** section. A few small pieces remain but it's mostly new.

As a result, the sections on **Data Properties** and **Methods** aren't really linked anymore. They could be split into two pages, though they would be quite short compared to the rest of the guide.

I've given up on trying to explain that the properties created by `methods` aren't reactive. There are plenty of more important points to hit.

src/guide/data-methods.md
skirtles-code: The later parts are now integrated.

I don't think we can open the paragraph with *'The `data` property'* because that would conflict with the more common meaning of 'data property'.

I've had a go at rewording the start of this paragraph to take things slower.

skirtles-code:
@NataliaTepluhina @phanan @bencodezen Thanks for all the feedback. I think I've addressed all the concerns raised so far.

The **Methods** section is almost completely rewritten since you last saw it.

NataliaTepluhina (approved):
@skirtles-code thanks for `Methods` rewrite! Looks good to me 👍🏻

NataliaTepluhina:
@phanan @bencodezen I am going to merge this one and if you want more changes to the doc, let's make it iteratively on follow-up PRs

PR #521, opened 2020-09-20
fix: change "an matching id" to "a matching id"

davidwarrington:
## Description of Problem
"setting the labels with an matching id" is incorrect.

## Proposed Solution
Change to "setting the labels with a matching id".


NataliaTepluhina (approved):
@davidwarrington thank you for fixing this!

Issue #522, opened 2020-09-20
adding examples and modifying parameter descriptions


zhenzhenChange:
document: src/api/global-api.md

Modified api:

- defineAsyncComponent

  Advanced Usage, Adding examples to the `onError` hook ([Reference rfc](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0026-async-component-api.md#retry-control)) :
  ```js
  ...
  onError(error, retry, fail, attempts) {
    if (error.message.match(/fetch/) && attempts <= 3) {
      // retry on fetch errors, 3 max attempts
      retry()
    } else {
      // Note that retry/fail are like resolve/reject of a promise:
      // one of them must be called for the error handling to continue.
      fail()
    }
  }
  ...
  ```

- resolveComponent

  Modify parameter descriptions: 
  ```md
  Accepts one argument: `component` --> Accepts one argument: `name`
  ```

skirtles-code:
I believe this has been fixed in #523.

PR #523, opened 2020-09-20
fix: adding examples and modifying parameter descriptions (#522)

zhenzhenChange:
See also: https://github.com/vuejs/docs-next/issues/522


NataliaTepluhina (approved):
@zhenzhenChange thank you for making these changes! 👍🏻

PR #529, opened 2020-09-21
Add the delimiters option

### LINKED ISSUE #429: Re-document delimiters
The `delimiters` setting has been removed from the documentation but it does seem to be alive and well:

Bug report: https://github.com/vuejs/vue-next/issues/1679

Example: https://jsfiddle.net/skirtle/hgntLx9s/11/

I believe there should be an entry in the API reference and in the Style Guide section 'Component/instance options order'.

skirtles-code:
## Description of Problem

The `delimiters` option was not included in early Vue 3 pre-releases, so it was removed from the documentation. However, it was reintroduced prior to 3.0.0 being released.

## Proposed Solution

I have added it back in the two places it was mentioned previously. I've made some changes to the wording so you may want to check you're happy with that.

## Additional Information

Closes #429.

If anyone knows a better way to escape the default value so that it'll render correctly, please let me know. What I've done does work but it isn't pretty.


NataliaTepluhina (approved):
@skirtles-code thank you for adding it back! 👍🏻

PR #530, opened 2020-09-21
fix: update the API reference for inheritAttrs

skirtles-code:
## Description of Problem

`inheritAttrs` has changed in Vue 3 and the current API reference entry is slightly incorrect.

1. The `class` and `style` are now included in `$attrs`, so they are affected by `inheritAttrs`.
2. The example given doesn't quite work. It emits the `input` event twice due to the listener in `$attrs`.

## Proposed Solution

1. Removed the note about `class` and `style`.
2. Added `emits` to the example, which fixes the double listeners.


NataliaTepluhina (approved):
Thank you for the fix! 👍🏻

PR #531, opened 2020-09-21
fix: remove the restriction for name, which no longer applies

skirtles-code:
## Description of Problem

For the `name` option, I don't think this line still applies in Vue 3:

> **Restriction:** only respected when used as a component option.

The distinction between component options and root instance options no longer exists. We now only have component options.

If you're curious, here's a demo showing `name` working fine on the root component: https://jsfiddle.net/skirtle/7ocdjz65/

## Proposed Solution

Remove that line.

### DIFF
--- a/src/api/options-misc.md
+++ b/src/api/options-misc.md
@@ -4,8 +4,6 @@
 
 - **Type:** `string`
 
-- **Restriction:** only respected when used as a component option.
-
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
@skirtles-code thank you for the fix!

PR #537, opened 2020-09-23
fix typo: instect -> inspect

ryo-gk:
## Description of Problem
The word `instecting` is typo.
`You can see the description by instecting Chrome DevTools:`

https://v3.vuejs.org/guide/a11y-semantics.html#instructions
Labels > aria-describedby
## Proposed Solution
change to `inspecting` from `instecting`.


NataliaTepluhina (approved):
@ryo-gk thank you for fixing it!

PR #539, opened 2020-09-23
Update component-basics.md

Meantub:
Change "does" to "do" in this context

## Description of Problem
"Do" is used when talking about plural nouns. Whereas we are using "does" which is supposed to be used for singular nouns.
## Proposed Solution
Change "does" to "do"


NataliaTepluhina (approved):
@Meantub thank you for fixing this! 👍🏻

PR #540, opened 2020-09-23
Fixed v-for/v-if priority and explanations

NataliaTepluhina:
## Description of Problem

In v3, the precedence of *v-for* and *v-if* when used on the same node is is now switched.

+ v2: *v-for* takes precedence over *v-if*
+ v3: *v-if* takes precedence over *v-for*

## Proposed Solution

Update respective sections of the guide

## Additional Information

Close https://github.com/vuejs/docs-next/issues/534

### DIFF
--- a/src/api/directives.md
+++ b/src/api/directives.md
@@ -62,7 +62,7 @@
 
   This directive triggers transitions when its condition changes.
 
-  When used together with `v-for`, `v-if` has a higher priority than v-if. We don't recommend using two these directives together on one element, see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
+  When used together with `v-for`, `v-if` has a higher priority than `v-for`. We don't recommend using these two directives together on one element, see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
 
 - **See also:** [Conditional Rendering - v-if](../guide/conditional.html#v-if)
 
--- a/src/api/directives.md
+++ b/src/api/directives.md
@@ -62,7 +62,7 @@
 
   This directive triggers transitions when its condition changes.
 
-  When used together with `v-for`, `v-if` has a higher priority than `v-for`. We don't recommend using these two directives together on one element, see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
+  When used together, `v-if` has a higher priority than `v-for`. We don't recommend using these two directives together on one element — see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
 
 - **See also:** [Conditional Rendering - v-if](../guide/conditional.html#v-if)
 
--- a/src/api/directives.md
+++ b/src/api/directives.md
@@ -62,7 +62,7 @@
 
   This directive triggers transitions when its condition changes.
 
-  When used together with `v-if`, `v-for` has a higher priority than v-if. See the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
+  When used together with `v-for`, `v-if` has a higher priority than v-if. We don't recommend using two these directives together on one element, see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
 
 - **See also:** [Conditional Rendering - v-if](../guide/conditional.html#v-if)
 
--- a/src/guide/conditional.md
+++ b/src/guide/conditional.md
@@ -91,4 +91,4 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 Using `v-if` and `v-for` together is **not recommended**. See the [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for further information.
 :::
 
-When used together with `v-if`, `v-for` has a higher priority than `v-if`. See the [list rendering guide](list#v-for-with-v-if) for details.
+When used together with `v-if`, `v-if` has a higher priority than `v-for`. See the [list rendering guide](list#v-for-with-v-if) for details.
--- a/src/guide/conditional.md
+++ b/src/guide/conditional.md
@@ -91,4 +91,4 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 Using `v-if` and `v-for` together is **not recommended**. See the [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for further information.
 :::
 
-When `v-if` and `v-for` are both used on the same element, the `v-if` will be evaluated first. See the [list rendering guide](list#v-for-with-v-if) for details.
+When `v-if` and `v-for` are both used on the same element, `v-if` will be evaluated first. See the [list rendering guide](list#v-for-with-v-if) for details.
--- a/src/guide/conditional.md
+++ b/src/guide/conditional.md
@@ -91,4 +91,4 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 Using `v-if` and `v-for` together is **not recommended**. See the [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for further information.
 :::
 
-When used together with `v-if`, `v-if` has a higher priority than `v-for`. See the [list rendering guide](list#v-for-with-v-if) for details.
+When `v-if` and `v-for` are both used on the same element, the `v-if` will be evaluated first. See the [list rendering guide](list#v-for-with-v-if) for details.
--- a/src/guide/list.md
+++ b/src/guide/list.md
@@ -268,25 +268,23 @@ Similar to template `v-if`, you can also use a `<template>` tag with `v-for` to 
 Note that it's **not** recommended to use `v-if` and `v-for` together. Refer to [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for details.
 :::
 
-When they exist on the same node, `v-for` has a higher priority than `v-if`. That means the `v-if` will be run on each iteration of the loop separately. This can be useful when you want to render nodes for only _some_ items, like below:
+When they exist on the same node, `v-if` has a higher priority than `v-for`. That means that if `v-if` condition depends on the property from `v-for` scope, we will have an error: 
 
 ```html
 <li v-for="todo in todos" v-if="!todo.isComplete">
+<!-- Property "todo" was accessed during render but is not defined on instance. -->
   {{ todo }}
 </li>
 ```
 
-The above only renders the todos that are not complete.
-
-If instead, your intent is to conditionally skip execution of the loop, you can place the `v-if` on a wrapper element (or [`<template>`](conditional#conditional-groups-with-v-if-on-lt-template-gt)). For example:
+This can be fixed with moving `v-for` to the wrapping <template> tag:
 
 ```html
-<ul v-if="todos.length">
-  <li v-for="todo in todos">
+<template> v-for="todo in todos"
+  <li v-if="!todo.isComplete">
     {{ todo }}
   </li>
-</ul>
-<p v-else>No todos left!</p>
+</template>
 ```
 
 ## `v-for` with a Component
--- a/src/guide/list.md
+++ b/src/guide/list.md
@@ -278,7 +278,7 @@ When they exist on the same node, `v-if` has a higher priority than `v-for`. Tha
 </li>
 ```
 
-This can be fixed with moving `v-for` to the wrapping <template> tag:
+This can be fixed by moving `v-for` to a wrapping `<template>` tag:
 
 ```html
 <template v-for="todo in todos">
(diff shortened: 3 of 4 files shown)


kazupon (approved):
LGTM!👍
Thanks quick fixes!

src/guide/list.md
```diff
@@ -268,25 +268,23 @@ Similar to template `v-if`, you can also use a `<template>` tag with `v-for` to
 Note that it's **not** recommended to use `v-if` and `v-for` together. Refer to [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for details.
 :::
 
-When they exist on the same node, `v-for` has a higher priority than `v-if`. That means the `v-if` will be run on each iteration of the loop separately. This can be useful when you want to render nodes for only _some_ items, like below:
+When they exist on the same node, `v-if` has a higher priority than `v-for`. That means that if `v-if` condition depends on the property from `v-for` scope, we will have an error:
```
skirtles-code, line 271: I think I'd avoid using the word 'property' to describe the iteration variable. The error message uses the word 'property' but that's because it's looking for a property on the instance.

Perhaps something like:

```text
That means the `v-if` condition will not have access to variables from the scope of the `v-for`.
```

src/guide/list.md
```diff
@@ -268,25 +268,23 @@ Similar to template `v-if`, you can also use a `<template>` tag with `v-for` to
 Note that it's **not** recommended to use `v-if` and `v-for` together. Refer to [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for details.
 :::
 
-When they exist on the same node, `v-for` has a higher priority than `v-if`. That means the `v-if` will be run on each iteration of the loop separately. This can be useful when you want to render nodes for only _some_ items, like below:
+When they exist on the same node, `v-if` has a higher priority than `v-for`. That means that if `v-if` condition depends on the property from `v-for` scope, we will have an error: 
 
 ```html
 <li v-for="todo in todos" v-if="!todo.isComplete">
+<!-- Property "todo" was accessed during render but is not defined on instance. -->
```
skirtles-code, line 275: I think this needs moving up a line. It's potentially confusing because the next line also mentions `todo`. Perhaps that line should be removed?

src/guide/list.md
```diff
@@ -268,25 +268,23 @@ Similar to template `v-if`, you can also use a `<template>` tag with `v-for` to
...
   {{ todo }}
 </li>
 ```
 
-The above only renders the todos that are not complete.
-
-If instead, your intent is to conditionally skip execution of the loop, you can place the `v-if` on a wrapper element (or [`<template>`](conditional#conditional-groups-with-v-if-on-lt-template-gt)). For example:
+This can be fixed with moving `v-for` to the wrapping <template> tag:
```
skirtles-code, line 280: The `<template>` needs backticks. I'd also make a few grammatical tweaks:

```text
This can be fixed by moving the `v-for` to a wrapping `<template>` tag:
                  ^^        ^^^            ^          ^^^^^^^^^^^^
```

src/style-guide/README.md
```diff
@@ -210,19 +210,9 @@ When Vue processes directives, `v-for` has a higher priority than `v-if`, so tha
 </ul>
 ```
 
-Will be evaluated similar to:
+Will throw an error, because `v-if` directive will be evaluated first and property `user` does not exist at this moment.
```
skirtles-code, line 213: I'd put a 'the' before `v-if`.

I'd also switch 'property' to 'the variable' or 'the iteration variable' or something like that.

src/style-guide/README.md
```diff
@@ -243,40 +233,18 @@ computed: {
 </ul>
 ```
 
-We get the following benefits:
+Alternatively, we can use a `<template>` tag with `v-for` to wrap `li` element:
```
skirtles-code, line 236: I think this needs an article:

```text
to wrap the `li` element:
        ^^^
```

Perhaps also angle brackets around the `<li>`?

src/style-guide/README.md
```diff
@@ -243,40 +233,18 @@ computed: {
...
+```html
 <ul>
-  <li
-    v-for="user in users"
-    v-if="shouldShowUsers"
-    :key="user.id"
-  >
-    {{ user.name }}
-  </li>
+  <template v-for="user in users" :key="user.id">
```
skirtles-code, line 240: Personally, I think this is fine.

However, this seems to clash with the style guide's own recommendation for elements with multiple attributes being split across multiple lines.

src/style-guide/README.md
```diff
@@ -321,14 +277,13 @@ By moving the `v-if` to a container element, we're no longer checking `shouldSho
...
-<ul v-if="shouldShowUsers">
-  <li
-    v-for="user in users"
-    :key="user.id"
-  >
-    {{ user.name }}
-  </li>
+```html
+<ul>
+  <template v-for="user in users" :key="user.id">
```
skirtles-code, line 282: As I noted previously, this is technically a breach of the style guide's own recommendations for multi-attribute elements.

Personally I think it's fine.

skirtles-code (changes requested):
@NataliaTepluhina Thanks for inviting me to participate in the review for this.

I have made some suggestions, I hope they're helpful.

src/api/directives.md
```diff
@@ -62,7 +62,7 @@
 
   This directive triggers transitions when its condition changes.
 
-  When used together with `v-if`, `v-for` has a higher priority than v-if. See the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
+  When used together with `v-for`, `v-if` has a higher priority than v-if. We don't recommend using two these directives together on one element, see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
```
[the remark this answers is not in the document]
NataliaTepluhina: Thanks, fixed!

src/guide/list.md
NataliaTepluhina: I don't think we should remove this line altogether because we need to show that this code snippet is incorrect. I changed a message to reflect this better

src/guide/list.md
NataliaTepluhina: I don't think we should remove this line altogether because we need to show that this code snippet is incorrect. I changed a message to reflect this better

src/guide/list.md
NataliaTepluhina: I don't think we should remove this line altogether because we need to show that this code snippet is incorrect. I changed a message to reflect this better

src/guide/list.md
NataliaTepluhina: Thank you, fixed 👍🏻

src/style-guide/README.md
NataliaTepluhina: I will leave it like this for now, I believe as soon as we add a linter this will be fixed automatically

src/style-guide/README.md
NataliaTepluhina: Fixed 👍🏻

src/guide/conditional.md
```diff
@@ -91,4 +91,4 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 Using `v-if` and `v-for` together is **not recommended**. See the [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for further information.
 :::
 
-When used together with `v-if`, `v-for` has a higher priority than `v-if`. See the [list rendering guide](list#v-for-with-v-if) for details.
+When used together with `v-if`, `v-if` has a higher priority than `v-for`. See the [list rendering guide](list#v-for-with-v-if) for details.
```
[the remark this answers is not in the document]
NataliaTepluhina: Fixed it here, for `directives` it's switched to `v-for` before the comma so I'll leave there

src/style-guide/README.md
NataliaTepluhina: Thank you, makes sense to me, fixed 🎉

NataliaTepluhina:
@skirtles-code thank you for the awesomely thorough review and suggestions! I've applied fixes, would you mind taking a second look?

src/api/directives.md
```diff
@@ -62,7 +62,7 @@
 
   This directive triggers transitions when its condition changes.
 
-  When used together with `v-if`, `v-for` has a higher priority than v-if. See the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
+  When used together with `v-for`, `v-if` has a higher priority than `v-for`. We don't recommend using these two directives together on one element, see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
```
phanan, line 65: Half joking: Maybe we do this to save ourselves a bit just in case the next change reverses the behavior again?

```suggestion
  When used together, `v-if` has a higher priority than `v-for`. We don't recommend using these two directives together on one element — see the [list rendering guide](../guide/list.html#v-for-with-v-if) for details.
```

src/guide/conditional.md
```diff
@@ -91,4 +91,4 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 Using `v-if` and `v-for` together is **not recommended**. See the [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for further information.
 :::
 
-When used together with `v-if`, `v-for` has a higher priority than `v-if`. See the [list rendering guide](list#v-for-with-v-if) for details.
+When `v-if` and `v-for` are both used on the same element, the `v-if` will be evaluated first. See the [list rendering guide](list#v-for-with-v-if) for details.
```
phanan, line 94:
```suggestion
When `v-if` and `v-for` are both used on the same element, `v-if` will be evaluated first. See the [list rendering guide](list#v-for-with-v-if) for details.
```

src/guide/list.md
```diff
@@ -268,25 +268,24 @@ Similar to template `v-if`, you can also use a `<template>` tag with `v-for` to
 Note that it's **not** recommended to use `v-if` and `v-for` together. Refer to [style guide](../style-guide/#avoid-v-if-with-v-for-essential) for details.
 :::
 
-When they exist on the same node, `v-for` has a higher priority than `v-if`. That means the `v-if` will be run on each iteration of the loop separately. This can be useful when you want to render nodes for only _some_ items, like below:
+When they exist on the same node, `v-if` has a higher priority than `v-for`. That means the `v-if` condition will not have access to variables from the scope of the `v-for`:
```
phanan, line 271:
```suggestion
When they exist on the same node, `v-if` has a higher priority than `v-for`. That means the `v-if` condition will not have access to variables from `v-for` scope:
```

phanan (commented):
Nice 👍 I'm being an annoying nitpicker again and left some non-blocking comments.

src/guide/list.md
skirtles-code: Ah, sorry, I wasn't clear.

I wasn't suggesting removing the comment line. I was suggesting removing the next line, `{{ todo }}`. The problem is that it mentions `todo`, so it wasn't entirely clear whether the error message was referring to the `v-if` or the `{{ todo }}`.

Moving the comment up has helped. I just wonder whether replacing the `{{ todo }}` with, e.g. `...` or a comment would remove any potential for someone to misunderstand? One of the earlier examples uses `<!-- content -->`, so perhaps that?

```html
<li v-for="todo in todos" v-if="!todo.isComplete">
  <!-- content -->
</li>
```

Now I look a bit closer there's actually another problem with the `{{ todo }}`. The `v-if="todo.isComplete"` implies that `todo` is an object with an `isComplete` property but the `{{ todo }}` seems to be treating `todo` like it's a string. It isn't technically wrong, I just worry about someone getting distracted by the `{{ todo }}` and thinking it has something to do with the error.

If you do decide to change it then the next example would need updating to match.

src/guide/list.md
skirtles-code: The wording 'the scope of the `v-for`' was my suggestion.

I don't think ending the sentence with '`v-for` scope' quite works in this context. To my ears, '`v-for` scope' refers to the concept in general whereas we're trying to refer to a specific `v-for` mentioned earlier in the paragraph.

phanan: Oh, that makes total sense for me 👍

src/api/directives.md
skirtles-code: I think this reads much better. The repetition of `v-for` didn't quite sound natural.

skirtles-code (approved):
@NataliaTepluhina I've made one more comment trying to clarify my earlier suggestion about one of the examples. I might just be worrying about nothing.

Happy for this to be merged if you don't think it's worth changing.

bencodezen (approved):
Great work on this @NataliaTepluhina! Other than the phrasing suggestions An added in the PR, it looks good to me!

PR #541, opened 2020-09-23
fix: improve component attrs section

kazupon:
## Description of Problem
The documentation says that `$attrs` can be accessed by setting `inheritAttrs: false`.
However, `$attrs` can be accessed with or without `inheritAttrs` setting.

The rfcs says the following:

>this.$attrs (and context.attrs for setup() and functional components) now contains all attributes passed to the component (as long as it is not declared as props). 

https://github.com/vuejs/rfcs/blob/master/active-rfcs/0031-attr-fallthrough.md#inheritattrs-false

## Proposed Solution
Remove the following sentence and add the new sentence of this PR, or we change to another explanation.

## Additional Information
I hope that you update my docs is wrote with my poor English! 🙇

### DIFF
--- a/src/guide/component-attrs.md
+++ b/src/guide/component-attrs.md
@@ -2,7 +2,7 @@
 
 > This page assumes you've already read the [Components Basics](component-basics.md). Read that first if you are new to components.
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
@kazupon thank you for the fix! Looks perfect to me 👍🏻

PR #545, opened 2020-09-24
docs: Add migration guide for `h` rendering registered component (#1927)

Fiv5:
## Description of Problem
In v3.x, `h` now no longer use a string ID (e.g. h('some-component')) to implicitly lookup globally registered components. It may explicitly show on the Migration Guide Docs.

## Proposed Solution
Update the Migration Guide.


NataliaTepluhina (commented):
@Fiv5 thank you for adding this contribution! I've left a few suggestions (mostly on wording/code formatting), rest looks good to me 👍🏻

Fiv5:
Thanks for the valuable suggestions!

src/guide/migration/render-function-api.md
```diff
@@ -128,6 +128,47 @@ In 3.x, the entire VNode props structure is flattened. Using the example from ab
...
+
+### 3.x Syntax
+
+In 3.x, with VNodes being context-free, we can no longer use a string ID to implicitly lookup registered components. Instead, we need to use an imported API:
+
+```js
+// 3.x
+import { h, resolveComponent } from 'vue'
+
+const ButtonCounter = resolveComponent('button-counter')
```
skirtles-code, line 165: I think you'd get a warning calling it like this:

> [Vue warn]: resolveComponent can only be used in render() or setup().

src/guide/render-function.md
```diff
@@ -394,8 +394,10 @@ render() {
 If we're writing a lot of `render` functions, it might feel painful to write something like this:
 
 ```js
+import { resolveComponent } from 'vue'
+
 Vue.h(
-  'anchored-heading',
+  resolveComponent('anchored-heading'),
```
skirtles-code, line 400: There seems to be a small inconsistency here. The rest of this example is using `Vue.h` but this line uses an imported `resolveComponent` rather than `Vue.resolveComponent`.

skirtles-code:
I don't believe `package-lock.json` should be included in this PR.

src/guide/migration/render-function-api.md
Fiv5: Sorry, I will double-check the examples provided.

skirtles-code (approved):

PR #552, opened 2020-09-25
correct misleading content about render function

unbyte:
see https://github.com/vuejs/vue-next/issues/2237

sample code and description for using event modifiers in render function is wrong


NataliaTepluhina (approved):
@unbyte thank you for the fix!

PR #557, opened 2020-09-26
docs: replace foo in v-model arguments

bencodezen:


NataliaTepluhina (approved):
@bencodezen thank you! Looks much better now 👍🏻

PR #558, opened 2020-09-27
Update custom-directives.md

DanielPe05:
## Description of Problem
Small typo

### DIFF
--- a/src/guide/migration/custom-directives.md
+++ b/src/guide/migration/custom-directives.md
@@ -85,7 +85,7 @@ Now that the custom directive lifecycle hooks mirror those of the components the
 
 ## Implementation Details
 
-In Vue 3, we're now supporting fragments, which allow us to return more than one DOM node per component. You can imagine how handy that is for something like a component with multiple lis or the children elements of a table:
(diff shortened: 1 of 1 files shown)


src/guide/migration/custom-directives.md
```diff
@@ -85,7 +85,7 @@ Now that the custom directive lifecycle hooks mirror those of the components the
 
 ## Implementation Details
 
-In Vue 3, we're now supporting fragments, which allow us to return more than one DOM node per component. You can imagine how handy that is for something like a component with multiple lis or the children elements of a table:
+In Vue 3, we're now supporting fragments, which allow us to return more than one DOM node per component. You can imagine how handy that is for something like a component with multiple lists or the children elements of a table:
```
NataliaTepluhina, line 88: I believe we meant multiple `<li>` tags here:

```suggestion
In Vue 3, we're now supporting fragments, which allow us to return more than one DOM node per component. You can imagine how handy that is for something like a component with multiple `<li>` elements or the children elements of a table:
```

NataliaTepluhina (changes requested):

skirtles-code (approved):

NataliaTepluhina (approved):
Thank you for the fix!

PR #562, opened 2020-09-28
Style Guide: Avoid toRefs with props

paulshen:
## Description of Problem

It may be tempting to standardize on destructuring props with `toRefs` but this can lead to subtle bugs.

## Proposed Solution

Add a style guide item to avoid `toRefs` with props when possible and read props using `props.foo` when possible.

## Additional Information

This is relevant when using the Composition API so this is probably an advanced issue. Feel free to ignore this or suggest alternatives. Thanks!

### DIFF
--- a/src/style-guide/README.md
+++ b/src/style-guide/README.md
@@ -1257,6 +1257,47 @@ While attribute values without any spaces are not required to have quotes in HTM
 ```
 </div>
 
+### Avoid toRefs with props <sup data-p="b">strongly recommended</sup>
+
+**Access props using props.propName inside `setup`.**
+
+When using the Composition API, you may consider using [`toRefs`](../guide/reactivity-fundamentals.html#destructuring-reactive-state) and destructuring `props` to have a consistent way to access reactive values.
+
+This can lead to subtle bugs when a prop is optional. When a prop is not provided, `toRefs` will not return a key for the prop. Destructuring will give you `undefined` and the following code will not be reactive.
+
+<div class="style-example style-example-bad">
+<h4>Bad</h4>
+
+``` js
+{
+  props: {
+    myNumber: Number
+  },
+  setup(props) {
+    const { myNumber } = toRefs(props);
+    const myNumberPlusOne = computed(() => (myNumber?.value ?? 0) + 1);
+    // ...
+  }
+}
+```
+</div>
+
+<div class="style-example style-example-good">
+<h4>Good</h4>
+
+``` js
+{
+  props: {
+    myNumber: Number
+  },
+  setup(props) {
+    const myNumberPlusOne = computed(() => (props.myNumber?.value ?? 0) + 1);
+    // ...
+  }
+}
+```
+</div>
+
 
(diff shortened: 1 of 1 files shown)


src/style-guide/README.md
```diff
@@ -1257,6 +1257,47 @@ While attribute values without any spaces are not required to have quotes in HTM
...
+<h4>Bad</h4>
+
+``` js
+{
+  props: {
+    myNumber: Number
+  },
+  setup(props) {
+    const { myNumber } = toRefs(props);
+    const myNumberPlusOne = computed(() => (myNumber?.value ?? 0) + 1);
```
skirtles-code, line 1278: I think the `?.` and `??` operators are still too new to be used in the documentation.

To some extent this example illustrates the dangers of using `?.`. It's being used to swallow an error, turning an unsubtle problem into a subtle one.

I wonder whether it's necessary for the example to show a subtle problem rather than an explosive one?

src/style-guide/README.md
```diff
@@ -1257,6 +1257,47 @@ While attribute values without any spaces are not required to have quotes in HTM
 ```
 </div>
 
+### Avoid toRefs with props <sup data-p="b">strongly recommended</sup>
+
+**Access props using props.propName inside `setup`.**
```
skirtles-code, line 1262: There's also [`toRef`](https://v3.vuejs.org/api/refs-api.html#toref). The documentation does recommend using that for grabbing individual `props` but it doesn't explicitly say 'because `toRefs` might not work'.

src/style-guide/README.md
```diff
@@ -1257,6 +1257,47 @@ While attribute values without any spaces are not required to have quotes in HTM
 ```
 </div>
 
+### Avoid toRefs with props <sup data-p="b">strongly recommended</sup>
```
skirtles-code, line 1260: I understand the desire to single out `props` for special attention, as it does seem likely to be the most common example of this problem. But I wonder whether the more general case, of any reactive object with optional properties, would be worth acknowledging?

skirtles-code (commented):
As far as I'm aware, this caveat of the Vue 3 reactivity system is not currently documented. I agree it should be mentioned somewhere.

I'm not sure the style guide is the right place for it. I don't see this as a 'style' issue. There are already a couple of 'non-style' recommendations in the style guide (about parent-child communication and state management) but personally I don't think they really belong in the style guide either.

There are a number of reactivity caveats with the new reactivity system but I can't find anywhere that they're currently documented. I think this would be a good fit for that section, if it existed. Writing something like that would be a significant undertaking, so it's understandable that it doesn't exist yet.

As an interim measure, I wonder whether this could be documented in the API reference for `toRefs`? Perhaps a general warning about missing properties, with a specific mention of `props` as the most common example?

skirtles-code:
I've opened #639 to try to address this.

paulshen:
@skirtles-code Oops I meant to reply but forgot! Your comments make a lot of sense and followup changes #639 look great. Thanks!

PR #567, opened 2020-09-29
Correct typo in "Single file components" reference

markjszy:
## Description of Problem
Current link to "Single file components" from the Typescript support page does not point to the correct location.

## Proposed Solution
Update the "Single file components" reference to point to the correct path.


NataliaTepluhina (approved):
@markjszy thank you for the fix!

PR #572, opened 2020-09-30
fix:fix a broken link

lichao0202:
## Description of Problem
In [Handling Edge Cases](https://vue3js.cn/docs/guide/component-edge-cases.html#controlling-updates) page, the link of "Components Basics" has broken and redirect to the 404 page.

The reason is that the link file name of "Components Basics" is not existed. The right name should be "component-basics.md"

## Proposed Solution
Change the link file name from "components.md" to "component-basics.md"

## Additional Information
In [Dynamic & Async Components](https://vue3js.cn/docs/guide/component-dynamic-async.html#dynamic-components-with-keep-alive) page, "Components Basics" also has a broken link, but I find nothing wrong with the code.

### DIFF
--- a/src/guide/component-edge-cases.md
+++ b/src/guide/component-edge-cases.md
@@ -1,6 +1,6 @@
 # Handling Edge Cases
 
-> This page assumes you've already read the [Components Basics](components.md). Read that first if you are new to components.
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
@lichao0202 thank you for the fix! 👍🏻

PR #576, opened 2020-09-30
fix, minor grammar. first-class citizen

bambalamm:
## Description of Problem
From typescript-support.md:
`This means you don't need any additional tooling to use TypeScript with Vue - it has a first-class citizen support.`

I'd think that 'It has a [...] support.' does not sound right here.

## Proposed Solution
So either; 1) *It has a first-class citizens' support*, or 2) *It has first-class citizen support* may fit better?
Option 2) technically means that first-class citizens support it. So I'd guess option 1) is the way to go.

## Additional Information
I'm not a native english speaker, so forgive me for wasting your time if I'm wrong altogether.
Cheers!

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -4,7 +4,7 @@
 
 ## Official Declaration in NPM Packages
 
-A static type system can help prevent many potential runtime errors as applications grow, which is why Vue 3 is written in TypeScript. This means you don't need any additional tooling to use TypeScript with Vue - it has a first-class citizens' support.
+A static type system can help prevent many potential runtime errors as applications grow, which is why Vue 3 is written in TypeScript. This means you don't need any additional tooling to use TypeScript with Vue - it has first-class citizen support.
 
 ## Recommended Configuration
 
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -4,7 +4,7 @@
 
 ## Official Declaration in NPM Packages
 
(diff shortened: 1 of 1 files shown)


src/guide/typescript-support.md
```diff
@@ -4,7 +4,7 @@
 
 ## Official Declaration in NPM Packages
 
-A static type system can help prevent many potential runtime errors as applications grow, which is why Vue 3 is written in TypeScript. This means you don't need any additional tooling to use TypeScript with Vue - it has a first-class citizen support.
+A static type system can help prevent many potential runtime errors as applications grow, which is why Vue 3 is written in TypeScript. This means you don't need any additional tooling to use TypeScript with Vue - it has a first-class citizens' support.
```
skirtles-code, line 7: I agree that the current wording is incorrect but only option 2 sounds right to me:

> it has first-class citizen support.

Either that or a total rewording.

bambalamm: Option 2 sounds alright to me.

skirtles-code (approved):

bambalamm:
Should I now do anything else or will the cosmos take it from here?

skirtles-code:
@bambalamm The change looks good to me but you'll need to wait for someone with merge rights to review it. Nothing needed from you until then.

NataliaTepluhina (approved):
@bambalamm thank you for the fix! Happy to be the cosmos moving this PR to merge 🎉

PR #595, opened 2020-10-04
feat(migration-guide): add global filters

LinusBorg:
## Description of Problem

The Migration strategy for filters was a little short and specifically did not help users with migrating their global filters.

## Proposed Solution

Give users guidance how to migrate globally registered filters with `app.config.globalProperties`


NataliaTepluhina (approved):
Thank you @LinusBorg, this is a great addition! 👍🏻

PR #600, opened 2020-10-05
docs: fix typo in options-data.md

Alex-Sokolov:
I think it was `Vue instance` and now need to be `component instance`


bencodezen (approved):
Thanks for the PR @Alex-Sokolov!

PR #603, opened 2020-10-06
make pinPadding value change reflect in the template

Timibadass:
## Description of Problem

The *200px* in the paragraph is supposed to change depending on the value of `pinPadding`.

## Proposed Solution
Replace *200*  in the text with `pinPadding` so it updates as the value changes.


bencodezen (approved):
Great catch. Thanks @Timibadass!

PR #605, opened 2020-10-07
Add the flush option for $watch

### LINKED ISSUE #457: Watch - flush
The `flush` option for `watch` is documented elsewhere but it isn't mentioned in *API Reference / Options / Data / watch*.

e.g.:

```js
watch: {
  a: {
    flush: 'sync',
    handler () {/**/}
  }
}
```

In addition, I think it would be preferable to have `handler`, `immediate`, `deep` and `flush` listed explicitly as options rather than relying on the example. `deep` and `immediate` are explained in much greater detail under `$watch`, which I find a little odd as that is a less common way to use watchers and I doubt anyone using `watch` will think to look under `$watch`.

`$watch` also needs `flush` documenting.

skirtles-code:
## Description of Problem

Vue 3 introduced the `flush` option for watchers. While it is documented in a few places, it isn't mentioned in the API reference for the `$watch` method or the `watch` option.

## Proposed Solution

I've added a complete description under `$watch`, similar to the `deep` and `immediate` options.

For the `watch` option I've only noted its existence and provided a link to `$watch`.

## Additional Information

Closes #457.

### DIFF
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -9,6 +9,7 @@
   - `{Object} options (optional)`
     - `{boolean} deep`
     - `{boolean} immediate`
+    - `{string} flush`
 
 - **Returns:** `{Function} unwatch`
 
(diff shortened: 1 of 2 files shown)


NataliaTepluhina (approved):
Thank you @skirtles-code!

PR #610, opened 2020-10-07
Minor edit to paragraph in Custom Events section

ElMassimo:
## Description of Problem

Starting the paragraph with _In the event_ might cause confusion, since the paragraph is discussing component events vs native events.

## Proposed Solution

Avoid using _In the event_, and use _When_ instead.

## Additional Information

We can also emphasize that the native event listener will not be used.


NataliaTepluhina (approved):
@ElMassimo thank you!

Issue #564, opened 2020-09-29
Figure caption on `css-vs-js-ease.svg`


u-sho:
This is at https://v3.vuejs.org/guide/transitions-overview.html#easing

The `CSS - limited handles` caption is broken.

<img width="1062" src="https://user-images.githubusercontent.com/26079835/94508009-fe77d100-024b-11eb-9503-df6ccb3d0146.png">

u-sho:
I think that these `<tspan>` tags make it wrong.

<img width="1263" src="https://user-images.githubusercontent.com/26079835/94508636-56fb9e00-024d-11eb-8e13-b284f396e93d.png">

sdras:
This seems like it's fixed. I can also give it another cleanup pass in another PR. Thanks!

PR #612, opened 2020-10-08
Fix broken text css-vs-js-ease.svg (#564)

MSib:
## Description of Problem
#564 The text crawls over the text

## Proposed Solution
Offset text  coordinates


NataliaTepluhina (approved):
@MSib thank you for the fix! 👍🏻

PR #614, opened 2020-10-09
fix: update broken links

skirtles-code:
This PR fixes various internal links.

In all cases the current link is pointing at something that doesn't exist. Usually this is because a section has been renamed or moved. In a few cases there was a problem with uppercase letters being used in the link hash.


bencodezen (approved):
Thanks for taking the time to diligently find and update these @skirtles-code!

PR #615, opened 2020-10-09
docs: fix typos in global-api.md

kungege:
## Description of Problem
fix typos in global-api.md


bencodezen:
Thanks for the PR @kungege! While I can see how it seems like a typo, this is an intentional separation between the two words. Appreciate your time!

PR #616, opened 2020-10-09
Docs/provide better intro to installation

bencodezen:
## Description of Problem

Got a question from a user that missed the CDN installation method. 

## Proposed Solution

Though it does exist and is searchable, I think users would benefit with a short summary at the beginning to funnel users correctly.

## Additional Information

I think it may be worth including links to online IDE's like CodePen, CodeSandbox, etc. in the future; but that is probably a discussion for another time and this iteration should be useful for users as it stands.


phanan (approved):
Nice! I especially like how you changed NPM to npm 😄

PR #619, opened 2020-10-12
fix: correct link casing

dargmuesli:
## Description of Problem
There is wrong casing in a link.
## Proposed Solution
Fixed casing.
## Additional Information
None.


skirtles-code:
Thank you for the PR but I believe this has already been fixed in #614.

Issue #556, opened 2020-09-26
Cosmetic issues on mobile


Alexn117:
I’m viewing on iOS Safari, for reference.

1. Scroll down from the very top of the homepage till you get the rubber band effect. There will be hidden object printed on screen.

![87C6D368-7039-43DA-9F12-0155903282D7](https://user-images.githubusercontent.com/29024614/94346023-dee86a80-fff7-11ea-9abc-9fc18e7c7bc7.jpeg)

2. The “K” button in the top right search box is not vertically aligned with the text of its sibling button.

![30A09B19-B50A-4E27-B904-DCA0A0FB3070](https://user-images.githubusercontent.com/29024614/94346053-10f9cc80-fff8-11ea-8526-b64b32f408c6.jpeg)

3. The “Subscribe” button in the footer does not seem to have full mobile styling.

![2554EFA1-AAA1-457B-9370-79A889DD8901](https://user-images.githubusercontent.com/29024614/94346091-6209c080-fff8-11ea-9662-309d41484de1.jpeg)

Just a few cosmetic issues I saw while browsing and thought I’d share them to better the site.

anilsahindev:
@Alexn117 @skirtles-code I created a PR for image 2 https://github.com/vuejs/docs-next/pull/625

skirtles-code:
I believe these have all been fixed now.

PR #543, opened 2020-09-24
Added PWA support

NataliaTepluhina:
## Description of Problem

Currently, the website lacks manifest and icons to have a PWA support. This MR adds icons, manifest, and PWA plugin to fix this

PR #622, opened 2020-10-13
fix: remove parentheses from the video player configuration

skirtles-code:
## Description of Problem

Every page of the docs has `<[object Object]>` at the top. Most of the time you can't see it. It's easiest to see on the 404 page, e.g. https://v3.vuejs.org/404

---

404 screenshot

---

As noted in #556, it can also be seen on other pages using some mobile browsers.

## Proposed Solution

Example: https://deploy-preview-622--vue-docs-next-preview.netlify.app/404

It is caused by a pair of parentheses in `config.js`:

```js
[
  ('script',
  {
    src: 'https://player.vimeo.com/api/player.js'
  })
],
```

In this PR I have removed those parentheses.

I am unclear whether this is the correct fix. Whatever that configuration setting was supposed to do, it doesn't seem to matter that it wasn't working. It may be that it can be removed instead.

## Additional Information

That section of the configuration was originally introduced by @phanan in #71. At that point it didn't have the parentheses. Effectively I am restoring the configuration to how it was at that point.

The parentheses were added in #543. I can't see anything in that PR that indicates that this was an intentional change.

Does anyone have a clear idea why that section is needed at all?

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -254,10 +254,10 @@ module.exports = {
     ],
     ['meta', { name: 'msapplication-TileColor', content: '#000000' }],
     [
-      ('script',
+      'script',
       {
         src: 'https://player.vimeo.com/api/player.js'
-      })
+      }
     ],
     [
       'script',


NataliaTepluhina (approved):
@skirtles-code thank you for the fix! I think they were introduced accidentally when changing the vuepress config

PR #625, opened 2020-10-13
Fix newsletter button on safari

anilsahindev:
## Description of Problem
IMG_3150

## Proposed Solution
IMG_3151

## Additional Information

https://www.w3.org/TR/CSS2/visudet.html#min-max-heights


NataliaTepluhina (approved):
@anilsahindev thanks! Looks much better

PR #638, opened 2020-10-20
Reduce heading levels for Template Refs to match other pages

skirtles-code:
## Description of Problem

The headings in `composition-api-template-refs.md` are `<h2>` and `<h3>`, rather than the `<h1>` and `<h2>` used by every other page.

## Proposed Solution

Remove the extra `#` signs from the headings.


bencodezen (approved):
Nice refactor. Thanks @skirtles-code!

PR #640, opened 2020-10-20
Transition needs children to use v-else-if

skirtles-code:
## Description of Problem

One of the `<transition>` examples uses multiple child elements, each with a `v-if`. While it will only have one child at runtime it triggers a template compiler warning about multiple children in a `<transition>`.

## Proposed Solution

Switching to `v-else-if` allows the template compiler to verify the children, removing the error.


bencodezen (approved):
Thanks for finding this @skirtles-code!

PR #646, opened 2020-10-23
Fix a description for custom directives used on components

NataliaTepluhina:
## Description of Problem

The current description says that directives are included to `$attrs` and can be bound to an element with `v-bind="$attrs"`

## Proposed Solution

We need to change the description to reflect the fact directives are not included in `$attrs`


bencodezen (approved):
Definitely an important change. Thanks for writing this so quickly!

PR #658, opened 2020-10-29
Update plugins.md

kaeyon:
## Description of Problem

Encountered undefined problem when experimenting with the sample code of plugins.

## Proposed Solution

According to the usage of the plugin and avoid undefined errors when experimenting with the sample code, `i18n` as the initialValue of reduce() might be `options` or give an explanation of variable `i18n`


NataliaTepluhina (approved):
@kaeyon thank you, looks good to me!

PR #665, opened 2020-11-01
Gradual Composition API Migration Cookbook recipe

sdras:
~Still a bit of a work in progress...~ this is close to being done now.


sdras:
The second is example is done now, it's ready for review whenever.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,255 @@
+# Gradual Composition API Migration
+
+The [Composition API](/guide/composition-api-introduction.html) is an advanced feature introduced in Vue 3. It is purely additive, and the Options API is not legacy. You may find, however, that the Composition API can be a useful feature for more flexible large-scale architectures. So how would we go about introducing it in a larger system, gradually? What follows is merely a recommendation, there are many potential ways forward.
+
+Due to the flexibility of the Composition API, you may need to be more explicit about organization so that future maintainers can quickly understand intended purpose of pieces of the application.
+
+## Refactor your Mixins
```
bencodezen, line 7:
```suggestion
## Refactor Your Mixins
```

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,255 @@
...
+<script>
+  import { toggle } from '@/mixins/toggle';
+
+  export default {
+    mixins: [toggle]
+  }
+</script>
+```
+
+We could refactor this by creating a `composables` folder and create `composables/useToggle.js` instead. We suggest using a directory named `composables` so that you can communicate that this is being used slightly differently from a component, it's reusable logic that you can consume.
```
bencodezen, line 60: This is a larger thought that I've been thinking on for some time, using this line as a way to open discussion.

I know this is starting to become more regularly used, but I'm thinking that `features` is a better name than `composables` due to the fact that these files are not required to be used in the Composition API context. Because of how Vue 3 has been rewritten internally, it would actually be more flexible to allow users the mental model of combining `features` together outside the component context as well. So in other word, `composables` seems to imply that these things are Vue specific, when other than some Vue magic, they're generally not and probably should not be given that impression.

The closest analogy to this is how `utilities` is a common folder that anyone who understand English and some programming would catch on quickly. And similarly, `features` would lower the barrier of entry. 

And in the case there are functions that require to be used in a Vue context (i.e., utilizes into lifecycle methods), then we could recommend a prefix like `setupToggle` which would be very explicit as far to its usage sine that would correlate with the `setup` method where it needs to be used.

Similarly, still need to test this out, but I wonder if `use` as a prefix will ultimately tie the concept too closely to React Hooks when the Composition API is in fact simpler and quite different from a usage / implementation perspective. And since as a verb it's fairly generic, like we would never name the lodash utilities function file "useLodash" even though we do "use" it, perhaps we should be setting precedent for more declarative verbs. So like if we're creating a new instance of toggle, "createToggle" would be more intuitive from a naming perspective.

What do you all think? @NataliaTepluhina @sdras @phanan

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,255 @@
...
+
+You may notice we used `reactive` here instead of `refs`. This is intentional- if you're refactoring a codebase, potentially full of many many values, `reactive` translates faster and more directly as the API is closer to Options, you're still using that same object notation.
+
+## In Place of Vuex
+
+It is possible to use the Composition API in place of Vuex, and save yourself a dependency. That said, it's not exactly necessary, and like many approaches, there are tradeoffs.
+
+If you're using Vuex, it's very clear exactly what centralized state is being used across the application. Composition API is very flexible, but you may lose that implicit declaration in communication to fellow maintainers. Our suggestion that if you do use it as a centralized state management store, that you place it in a `stateManagement` folder, or something similarly named, so that responsibilities are clear.
```
bencodezen, line 131:
```suggestion
If you're using Vuex, it's very clear exactly what centralized state is being used across the application. In addition, you gain the benefit of using a standardized pattern that comes with officially supported tooling in the DevTools which can make a big difference when it comes to debugging code.

Composition API is very flexible, but you may lose that implicit declaration in communication to fellow maintainers. Our suggestion that if you do use it as a centralized state management store, that you place it in a `stateManagement` folder, or something similarly named, so that responsibilities are clear.
```

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,255 @@
...
+     sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+   ></iframe>
+
+The second example is reusable throughout the application across multiple components, not just for the gitHub API as the first example shows.
+
+[VueUse](https://vueuse.js.org/) is a great resource to explore that covers many of these encapsulated use cases, ready to use in your application.
+
+## Alternative Patterns
+
+Not everything needs to be refactored, though. Use judgement to guide you and not hype. The Options API can support quite a lot of use cases, and it's absolutely fine to keep your components in the Options API, and move reusable logic and/or state management to Composition API, depending on your needs.
```
bencodezen, line 247:
```suggestion
Not everything needs to be refactored, though. Use data driven refactoring to guide you and avoid pre-optimizing when it's not necessary. The Options API is still a valid approach that will still allow you to build powerful applications. And in the event you find there's a lot of reusable logics and/or state amongst your application, that might be a good time to consider the Composition API to enhance your application!
```

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,255 @@
...
+
+Remember: the strength of the Composition API is in:
+
+- Reuse
```
bencodezen, line 251:
```suggestion
- Reuse
- Organization
```

bencodezen (approved):
Love the first iteration of this @sdras! 

I've left some discussion points that aren't blocking, so I've gone ahead and approved.

src/cookbook/gradual-compositionapi-migration.md
sdras: I completely agree because for me it's actually tougher to scan so easily- `composables` and `components` are close enough in spelling that it takes me one more second to understand which one is which. I like `features` way better myself.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+<iframe src="https://codesandbox.io/embed/refactor-mixin-to-composition-api-hnloh?fontsize=14&hidenavigation=1&theme=dark"
+     style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;"
+     title="Refactor Mixin to Composition API"
+     allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+     sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+   ></iframe>
+
+Note that the template stays the same, but we've extracted the logic to use in our script section differently. If, from here, we wanted to use `isShowing` in the Options API, we could access it with `this.isShowing` now, just as we normally do with a data property, and similarly, we can access `this.toggleShow` like we would a method.
+
+You may notice we used `reactive` here instead of `refs`. This is intentional- if you're refactoring a codebase, potentially full of many many values, `reactive` translates faster and more directly as the API is closer to Options, you're still using that same object notation.
```
NataliaTepluhina, line 125: **question**: should we mention that using `toRefs` only for destructuring? Because we're saying that we prefer `reactive` but then we convert it to standalone `refs`

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+      loading: false
+    }
+  },
+  methods: {
+    async callGitHub() {
+      const { hasErrors, isLoading, response, apiCall } = useFetchAPI(
+        `https://api.github.com/users/sdras/repos?page=1&per_page=100`,
+        {}
+      )
+      apiCall()
```
NataliaTepluhina, line 227: **question**: shouldn't this be
```suggestion
      await apiCall()
```

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+}
+```
+
+And now we can refactor our earlier component like this:
+
+```js
+import { useFetchAPI } from '@/composables/useFetchAPI.js'
+
+export default {
+  data() {
```
NataliaTepluhina, line 214: **question (probably dumb one)**: why do we prefer not to use `setup()` here for returned reactive properties and then combine it with Options API method?

NataliaTepluhina (commented):
Awesome work @sdras! I've left a few questions here and there

src/cookbook/gradual-compositionapi-migration.md
sdras: Extremely good point, I'll clear that up.

src/cookbook/gradual-compositionapi-migration.md
sdras: yes! whoops

src/cookbook/gradual-compositionapi-migration.md
sdras: I thought it might be nice to show how it would be used in options for those trying to merge, but am open to other ways of demonstrating

src/cookbook/gradual-compositionapi-migration.md
skirtles-code, line 125: Is the word 'many' repeated intentionally here?

It might be worth considering moving this paragraph up to be straight after the code for `useToggle.js`. I got distracted because the earlier example was using `reactive` instead of `ref`.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+Note that the template stays the same, but we've extracted the logic to use in our script section differently. If, from here, we wanted to use `isShowing` in the Options API, we could access it with `this.isShowing` now, just as we normally do with a data property, and similarly, we can access `this.toggleShow` like we would a method.
+
+You may notice we used `reactive` here instead of `refs`. This is intentional- if you're refactoring a codebase, potentially full of many many values, `reactive` translates faster and more directly as the API is closer to Options, you're still using that same object notation.
+
+## In Place of Vuex
+
+It is possible to use the Composition API in place of Vuex, and save yourself a dependency. That said, it's not exactly necessary, and like many approaches, there are tradeoffs.
+
+If you're using Vuex, it's very clear exactly what centralized state is being used across the application. In addition, you gain the benefit of using a standardized pattern that comes with officially supported tooling in the DevTools which can make a big difference when it comes to debugging code.
```
skirtles-code, line 131: Comma before *'which'*?

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+Composition API is very flexible, but you may lose that implicit declaration in communication to fellow maintainers. Our suggestion that if you do use it as a centralized state management store, that you place it in a `stateManagement` folder, or something similarly named, so that responsibilities are clear.
```
skirtles-code, line 133: I'd start this paragraph with *'The Composition API'*. It doesn't sound right to me without the *'The'*

Also, I think it needs *'is'* after the word *'suggestion'*.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Components
+
+It is possible at this point to start to move your components to the Composition API, however, for larger code bases, you may not see immediate need or good reason to, as the Options API works nicely.
```
skirtles-code, line 137: I think *'codebases'* as a single word is the more common usage. It's already been used that way earlier on this same page.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+Our suggestion is to use this opportunity to think through the architecture and refactor any larger components that have too much responsibility on their own.
+
+For instance, if you have a component has the following methods:
```
skirtles-code, line 141: Missing *'that'*.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+    response: null,
+    isLoading: false,
+    hasErrors: null
+  })
+
+  const apiCall = async () => {
+    state.isLoading = true
+    try {
+      let res = await fetch(api, options)
+      state.response = await res.json()
```
skirtles-code, line 193: This version is using `await` twice, whereas the original code used `await` and `then`. Is there a reason for that and could they be both written the same way to avoid the change being a distraction?

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+},
+created() {
+  this.getGitHubProjects()
+}
+```
+
+### After, with Composition API
+
+```js
+import { reactive, toRefs } from '@vue/composition-api'
```
skirtles-code, line 180: Vue 2 package.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+export function useFetchAPI(api, options) {
+  const state = reactive({
+    response: null,
+    isLoading: false,
+    hasErrors: null
```
skirtles-code, line 186: To me, `hasErrors` sounds like a boolean but that doesn't appear to be how it's used.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+export default {
+  data() {
+    return {
+      gitHubData: [],
+      errors: null,
+      loading: false
+    }
+  },
+  methods: {
+    async callGitHub() {
```
skirtles-code, line 222: I don't think there's any need for the `async` here. The function doesn't return anything and it isn't using `await`.

Shouldn't this example still have the `created` hook?

More generally, I find this example a little odd. I guess I'd have expected most of this logic to be in `setup` with `callGitHub` just being the returned alias for `apiCall`. That might just be my lack of experience with the Composition API...

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+<iframe src="https://codesandbox.io/embed/refactor-fetch-with-composition-api-klyji?fontsize=14&hidenavigation=1&theme=dark"
+     style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;"
+     title="Refactor Fetch with Composition API"
+     allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+     sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+   ></iframe>
+
+The second example is reusable throughout the application across multiple components, not just for the gitHub API as the first example shows.
```
skirtles-code, line 243: Capital G on *'GitHub'*?

I'm struggling a bit to parse the end of this sentence.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+[VueUse](https://vueuse.js.org/) is a great resource to explore that covers many of these encapsulated use cases, ready to use in your application.
+
+## Alternative Patterns
+
+Not everything needs to be refactored, though. Use data driven refactoring to guide you and avoid pre-optimizing when it's not necessary. The Options API is still a valid approach that will still allow you to build powerful applications. And in the event you find there's a lot of reusable logics and/or state amongst your application, that might be a good time to consider the Composition API to enhance your application!
```
skirtles-code, line 249: Hyphen in *'data driven'*?

I'm not sure whether *'logics'* is the correct plural here. I'd have gone with *'logic'*.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+Remember: the strength of the Composition API is in:
+
+- Reuse
+- Organization
+- Explicitly stating what's being returned
+- Flexibility
+
+These same flexible traits that make it easy to compose them mean they can be used for many purposes, so it takes some vigilance to provide clarity those who are reading the code what the purpose may be. We suggest refactoring with that type of organization in mind.
```
skirtles-code, line 258: I'm unclear what *'them'* and *'they'* refer to here.

I think there are some missing words here too. I think it needs *'for'* before *'those'* and *'as to'* before *'what'*.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+      this.isLoading = false;
+    }
+  }
+},
+created() {
+  this.getGitHubProjects()
+}
+```
+
+### After, with Composition API
```
skirtles-code, line 177: Perhaps a *'the'* before *'Composition API'*?

skirtles-code (changes requested):
Some thoughts from me.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+And it was being used in a modal component in `components/modal.vue`:
+
+```html
+<template>
+  <div>
+    <h3>Let's trigger this here modal!</h3>
+
+    <button @click="toggleShow">
+      <span v-if="isShowing">Hide child</span>
```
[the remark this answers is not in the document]
LinusBorg: I personally use a ternary expression in an interpolation, provided the text is short enough:

```
{{ isShowing ? 'Hide' : 'Show' }} child
```

src/cookbook/gradual-compositionapi-migration.md
LinusBorg: Could we be using it in `setup` and then demonstrate how the stuff returned from setup is still accessible as normal instance properties for the rest of the Options-API code?

That way, people see that 

1. this stuff "belongs" into setup (at least with lifecycle methods, it has to go there)
2. But it can still interact with the Options API, though only unidirectionally (setup can't access Options-API methods etc)

I  guess the example would have to be modified/extended for this

sdras: Yeah, that's a really good point, I'll update it and make mention of this. Thanks!

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
+# Gradual Composition API Migration
+
+The [Composition API](/guide/composition-api-introduction.html) is an advanced feature introduced in Vue 3. It is purely additive, and the Options API is not legacy. You may find, however, that the Composition API can be a useful feature for more flexible large-scale architectures. So how would we go about introducing it in a larger system, gradually? What follows is merely a recommendation, there are many potential ways forward.
+
+Due to the flexibility of the Composition API, you may need to be more explicit about organization so that future maintainers can quickly understand intended purpose of pieces of the application.
```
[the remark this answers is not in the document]
sdras:
```suggestion
Due to the flexibility of the Composition API, you may need to be more explicit about organization so that future maintainers can quickly understand the intended purpose of pieces of the application.
```

src/cookbook/gradual-compositionapi-migration.md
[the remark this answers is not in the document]
sdras: Yeah, agreed with @LinusBorg, will make that change :)

src/cookbook/gradual-compositionapi-migration.md
sdras: It's not on the created hook because the example is called from a button. While it's possible to call the api with created and then display with a method, that's a lot more for someone to parse while trying to understand something else. The rest is dealt with in another comment, but thanks!

src/cookbook/gradual-compositionapi-migration.md
sdras:
```suggestion
The second example is reusable throughout the application across multiple components, while the first example would only work for the GitHub API.
```

src/cookbook/gradual-compositionapi-migration.md
sdras:
```suggestion
Not everything needs to be refactored, though. Use data-driven refactoring to guide you and avoid pre-optimizing when it's not necessary. The Options API is still a valid approach that will still allow you to build powerful applications. And in the event you find there's a lot of reusable logic and/or state amongst your application, that might be a good time to consider the Composition API to enhance your application!
```

src/cookbook/gradual-compositionapi-migration.md
sdras: Hm, I think this one I will leave as-is, unless I hear from others that it's confusing, I guess I'm unclear on what's unclear :)

src/cookbook/gradual-compositionapi-migration.md
sdras:
```suggestion
### After, with the Composition API
```

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+## Refactor Your Mixins
+
+If you've made use of mixins for reusable logic across components, it's recommended that you start your refactor here. Mixins translate very directly.
+
+### Base Example
+
+Let's say we had this very small toggle mixin in `mixins/toggle.js`:
+
+```js
+const toggle = {
```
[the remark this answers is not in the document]
sdras: Yep, that was left over from this originally being in codepen as one JS file, nice catch :)

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+
+    <div v-if="isShowing" class="modal">
+      <h2>Here I am!</h2>
+      <button @click="toggleShow">Close</button>
+    </div>
+  </div>
+</template>
+```
+
+```js
```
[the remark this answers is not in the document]
sdras: Hm, this isn't the wrong language but I'm happy to update the whole sample to vue

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+<script>
+  import { toggle } from '@/mixins/toggle';
+
+  export default {
+    mixins: [toggle]
+  }
+</script>
+```
+
+We could refactor this by creating a `composables` folder and create `composables/useToggle.js` instead. We suggest using a directory named `composables` so that you can communicate that this is being used slightly differently from a component, it's reusable logic that you can consume.
```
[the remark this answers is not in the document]
sdras: Typically with the Composition API, we use camelCasing and a prefix of `use`.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+      return {
+        isShowing,
+        toggleShow
+      };
+    },
+  }
+</script>
+```
+
+<iframe src="https://codesandbox.io/embed/refactor-mixin-to-composition-api-hnloh?fontsize=14&hidenavigation=1&theme=dark"
```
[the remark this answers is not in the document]
sdras: Ah, yeah, I changed the sandbox to be more clear what the focus was. I'll update that. Thanks!

And yeah, it's using Vue 2 intentionally because we're showing how to gradually update from an existing Vue 2 application, not from Vue 3 greenfield.

src/cookbook/gradual-compositionapi-migration.md
sdras:
```suggestion
For instance, if you have a component that has the following methods:
```

src/cookbook/gradual-compositionapi-migration.md
sdras: My spell check says no, it forced me to correct it to what you see here :)

src/cookbook/gradual-compositionapi-migration.md
sdras:
```suggestion
The Composition API is very flexible, but you may lose that implicit declaration in communication to fellow maintainers. Our suggestion is that if you do use it as a centralized state management store, that you place it in a `stateManagement` folder, or something similarly named, so that responsibilities are clear.
```

src/cookbook/gradual-compositionapi-migration.md
sdras: I believe two commas is considered comma splicing, but could be persuaded here.

src/cookbook/gradual-compositionapi-migration.md
sdras: As mentioned above, the examples are Vue 2 intentionally because the aim is to show how to migrate an existing large Vue 2 application, not a greenfield Vue 3 application

src/cookbook/gradual-compositionapi-migration.md
sdras: This is updated! @yyx990803 when you take a look at this, you might want to sound off as we're defining a new convention.

src/cookbook/gradual-compositionapi-migration.md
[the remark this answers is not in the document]
sdras: Turns out I had already updated the template sample to be more simple for the purposes of learning anyways, so this is now outdated.

src/cookbook/gradual-compositionapi-migration.md
```diff
@@ -0,0 +1,258 @@
...
+      <h2>Here I am!</h2>
+      <button @click="toggleShow">Close</button>
+    </div>
+  </div>
+</template>
+```
+
+```js
+<script>
+  import { useToggle } from "@/composables/useToggle.js";
```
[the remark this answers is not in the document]
sdras: Yep! I'll make those updates. Thanks!

src/cookbook/gradual-compositionapi-migration.md
sdras: It was, but I made it bold instead, as per this guidance. I'm going to push back on swapping the paragraphs, though, they both reference the code above.

src/cookbook/gradual-compositionapi-migration.md
sdras: I think that could be argued- you can also say "these are the errors the component has", I think I will keep it as-is unless  others agree, if you don't mind.

src/cookbook/gradual-compositionapi-migration.md
sdras: Thanks! Fixed!

src/cookbook/gradual-compositionapi-migration.md
sdras: This is now updated! Thank you both :) Feel free to let me know if you have you have further changes.

src/cookbook/gradual-compositionapi-migration.md
sdras: Oh, for the `setup` vs `use`- I agree but didn't make that change yet because I think there are already so many examples out there that have `use` as a convention, including the VueUse site, that there would probably be bigger community impacts there.

src/cookbook/gradual-composition-api-migration.md
```diff
@@ -0,0 +1,245 @@
...
+
+And now we can refactor our earlier component like this:
+
+```js
+import { useFetchAPI } from '@/features/useFetchAPI.js'
+
+export default {
+  data() {
+    return {
+      gitHubData: [],
```
NataliaTepluhina, line 194: With all the changes made, do we still need these reactive properties in data? I mean we could alias and use those we were exposing from `useFetchAPI`.

I understand this kinda makes the method meaningless and kills the point of the example. Maybe we could add one more 'side' data property and side effect in the method changing it? something like `isModalOpen` and change it to false after the API call is resolved?

LinusBorg: I agree that we can replace the data proprties with the return values of the composition function.

Since that makes the method redundant, maybe the example could instead have a `watch: { errors: ... }` that emits an event when there has been an error?

That would suffiently demonstrate how to access setup return values in Options API

src/cookbook/gradual-composition-api-migration.md
```diff
@@ -0,0 +1,245 @@
...
+      response,
+      apiCall
+    }
+  },
+  methods: {
+    async callGitHub() {
+      await this.apiCall()
+      this.gitHubData = this.response
+      this.errors = this.hasErrors
+      this.loading = this.isLoading
```
skirtles-code, line 217: As currently written, the `loading` flag is never set to `true`. `isLoading` does change to `true` but it isn't copied across to `loading` until the awaited promise has resolved, by which time it's back to `false`.

I'm guessing this example is going to be tweaked anyway, so this problem may well disappear as part of that.

skirtles-code (commented):
Here are a few follow-up thoughts and explanations of things I suggested previously.

src/cookbook/gradual-composition-api-migration.md
```diff
@@ -0,0 +1,245 @@
+# Gradual Composition API Migration
+
+The [Composition API](/guide/composition-api-introduction.html) is an advanced feature introduced in Vue 3. It is purely additive, and the Options API is not legacy. You may find, however, that the Composition API can be a useful feature for more flexible large-scale architectures. So how would we go about introducing it in a larger system, gradually? What follows is merely a recommendation, there are many potential ways forward.
```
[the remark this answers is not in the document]
sdras: We can mention it, sure- same response here as to some of the other questions before- we're using Vue 2 with a Vue 3 package because the purpose of the article is to show how people can migrate an existing Vue 2 project, not a greenfield Vue 3 one. It's also mentioned that this isn't the only way to go about it.

src/cookbook/gradual-composition-api-migration.md
```diff
@@ -0,0 +1,245 @@
...
+    toggleShow() {
+      this.isShowing = !this.isShowing
+    }
+  }
+}
+```
+
+And it was being used in a modal component in `components/AppModal.vue`:
+
+```vue
```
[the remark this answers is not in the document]
sdras: Makes sense, thanks!

src/cookbook/gradual-composition-api-migration.md
sdras: good points, I'll refactor. thanks!

src/cookbook/gradual-composition-api-migration.md
```diff
@@ -0,0 +1,245 @@
...
+
+You may notice we used `reactive` here instead of `refs`. This is intentional- though this example is slim, if you're refactoring a codebase, potentially full of **many** values, `reactive` translates faster and more directly as the API is closer to Options, you're still using that same object notation. We are using `toRefs` here for the purposes of destructuring.
+
+## In Place of Vuex
+
+It is possible to use the Composition API in place of Vuex, and save yourself a dependency. That said, it's not exactly necessary, and like many approaches, there are tradeoffs.
+
+If you're using Vuex, it's very clear exactly what centralized state is being used across the application. In addition, you gain the benefit of using a standardized pattern that comes with officially supported tooling in the DevTools which can make a big difference when it comes to debugging code.
```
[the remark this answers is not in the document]
sdras:
```suggestion
If you're using Vuex, it's very clear exactly what centralized state is being used across the application. In addition, you gain the benefit of using a standardized pattern that comes with officially supported tooling in the DevTools. This can make a big difference when it comes to debugging code.
```

devdreamsolution (approved):

src/cookbook/gradual-compositionapi-migration.md
sdras: For those following along, we chatted with Evan, we decided that the ship has already sailed on `use` vs `setup` though Ben makes an excellent point. We also decided to stick to `composables` for now- I'll be reverting that.

src/cookbook/gradual-composition-api-migration.md
```diff
@@ -0,0 +1,245 @@
...
+      this.isLoading = true
+      let projects = await fetch(
+        `https://api.github.com/users/sdras/repos?page=1&per_page=100`
+      )
+      this.gitHubProjects = await projects.json()
+    } catch (error) {
+      this.error = error
+      console.error(error);
+    } finally {
+      this.isLoading = false;
```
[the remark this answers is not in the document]
sdras: Thanks for catching that! Artifacts from bringing things back over from CodeSandbox

skirtles-code:
I'm going to close this. I believe the new docs cover all the key material from this PR.

PR #666, opened 2020-11-02
Fix broken text css-vs-js-ease.svg

u-sho:
## Description of Problem
#564 The text crawls over the text
#612 was good but I think that the text looks inverted like `limtied`, yet.

**#612 solution**
スクリーンショット 2020-11-02 15 02 51

## Proposed Solution

**this PR**
スクリーンショット 2020-11-02 15 02 40


u-sho:
I checked with chrome.
If it depends on my environment, and if this PR looks bad on other's environment, sorry.

phanan:
Thanks for the effort, but I think this version is broken on Chromium browsers:

image

I'll go ahead and close the PR. Feel free to open another once the problem is actually fixed.

u-sho:
@phanan I see.
I appreciate for your review :bow:

PR #667, opened 2020-11-02
chore: fix heading level

kawamataryo:
## Description of Problem
Header level is not matched with other files.

## Proposed Solution

Changed the heading level of `Overview` from h1 to h2


phanan:
Great catch, thanks!

PR #668, opened 2020-11-02
Add dynamic slots section to key-attribute.md page.

lukewarlow:
## Description of Problem
Vue 3 now requires a key attribute to be added to a slot with a dynamic name. As far as I can tell this is currently undocumented.

## Proposed Solution
I have added a section to the key-attribute.md page to explain this requirement.

I believe this is the correct place, but happy to move it if there's somewhere more suitable.

## Additional Information
https://github.com/vuejs/vue-next/issues/2535 from my issue here, it appears this is expected behaviour, so this MR is to document it.


skirtles-code:
I think we should hold off on merging this until it is confirmed that this is an intended change. Using a `key` may be necessary with 3.0.2 but https://github.com/vuejs/vue-next/issues/2535 is currently marked as an open bug. I dare say there'll be more discussion on that issue over the next few days.

devdreamsolution (approved):

yyx990803:
This is now fixed in core.

PR #670, opened 2020-11-03
Document removal of $children API on Migration guide

NataliaTepluhina:
`$children` instance property is removed but this removal is not documented. This PR adds a required change to Migration guide

Close https://github.com/vuejs/docs-next/issues/454

### DIFF
--- a/src/guide/migration/children.md
+++ b/src/guide/migration/children.md
@@ -1,0 +1,38 @@
+---
+badges:
+  - removed
+---
+
+# $children <MigrationBadges :badges="$frontmatter.badges" />
+
+## Overview
+
+`$children` instance property removed from Vue 3.0 and no longer supported.
+
+## 2.x Syntax
+
+In 2.x, developers could access direct child components of the current instance with `this.$children`:
+
+```html
+<div ref="app" id="app">
+  <img alt="Vue logo" src="./assets/logo.png" width="25%" />
+  <my-button>Change logo</my-button>
+</div>
+```
+
+```js
+export default {
+  name: "App",
+  components: {
+    MyButton,
+  },
+  mounted() {
+    console.log(this.$children); // [VueComponent]
+  },
+};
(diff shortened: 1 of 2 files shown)


src/guide/migration/children.md
```diff
@@ -0,0 +1,38 @@
...
+  mounted() {
+    console.log(this.$children); // [VueComponent]
+  },
+};
+</script>
+```
+
+## 3.x Update
+
+In 3.x, `$children` property is removed and no longer supported. Instead, if you need to access a child component instance, we recommend using [$refs](/guide/component-template-refs.html#template-refs).
```
LinusBorg, line 38: refs only work in a straightforward way when accessing children in a component's own template.

However, I think there are quite a few plugins out there that rely on `$children` in order to access child components in slots - think if a compound component like this:

```html
<tabs>
  <tab  name="B" />
  <tab name="B" />
</tabs>
```

It's not a particularly good pattern in my opinion as you could easily end up with children that you don't expect, but it seems it's used in the community.

Can we provide some guidance how to avoid such a pattern?

* One way would be to provide a ref function (like for `v-for`) through `v-slot` props, but that's not really ergonomic.
* Provide/inject your be used to make children register themselves with the parent.
  * that pattern could even be wrapped in a mixin /comp-Function to simulate `$children`behaviour

LinusBorg (commented):
Looking good. I commented a bit bout edge cases that we might want to provide guidance for, but we can consider adding that later and getting this info online first.

src/guide/migration/children.md
NataliaTepluhina: @LinusBorg, I agree with the addition but (as you mentioned in he above comment) let's add it in the follow-up PR :+1:

seanwuapps: would love to see guidance on this pattern please

PR #675, opened 2020-11-05
Fix a typo

kidonng:
## Description of Problem

There is a typo on the "Translations" page.

## Proposed Solution

It is now corrected.


NataliaTepluhina (approved):
@kidonng thank you for the fix!

PR #677, opened 2020-11-06
Add inject example to reactive section of docs

Timibadass:
This example might come in handy for users that are yet to read on `Computed and Watch` in the documentation so they know that the properties they're making reactive using `computed` is available under `property.value` and not `property`.

### DIFF
--- a/src/guide/component-provide-inject.md
+++ b/src/guide/component-provide-inject.md
@@ -107,6 +107,13 @@ app.component('todo-list', {
     }
   }
 })
+
+app.component('todo-list-statistics', {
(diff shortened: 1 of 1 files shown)


skirtles-code:
You raise a really good point.

As far as I can tell, this is the first time the Composition API is mentioned anywhere in the guide. The paragraph just before this example uses various terms that the reader won't have met yet. There's no acknowledgement (in that paragraph) that these are concepts from later in the guide, it just presents them like something the reader is expected to know about.

devdreamsolution (approved):

PR #679, opened 2020-11-08
Increase header depth and tweak wording in transitions-overview.md

skirtles-code:
3 small changes to `transitions-overview.md`:

1. Change `<h1>` to `<h2>` for a sub-header.
2. Change *affects* to *effects* as that seems to be what was intended.
3. Change *have* to *make* as I think the sentence reads better that way.


devdreamsolution (approved):

NataliaTepluhina (approved):
@skirtles-code changes look good to me 👍🏻

PR #682, opened 2020-11-08
Restore search page

NoelDeMartin:
Restored search page that was merged in #363.

The page was removed in https://github.com/vuejs/docs-next/commit/2738ccfc7717b788deda024a977800a340b93235, I'm not sure why but looking at the code it seems like some things were broken after updating the algolia dependencies.

In the first commit of this PR I'm restoring what was already discussed in #363, the new code is on the second commit.

### DIFF
--- a/src/.vuepress/components/search/index.vue
+++ b/src/.vuepress/components/search/index.vue
@@ -1,0 +1,264 @@
+<template>
+  <main id="search-page">
+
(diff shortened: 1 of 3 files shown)


NataliaTepluhina (approved):
Thank you @NoelDeMartin! Yes, the page was removed because it was breaking an Algolia integration, thank you for restoring it! 🙇🏻‍♀️

PR #684, opened 2020-11-09
update jsdelivr link

isaacl:
point to current vue release


NataliaTepluhina (approved):
@isaacl thank you for the fix!

PR #704, opened 2020-11-20
Update application-config.md

vjandrei:
I added this app.mount to the code block to visual more what the before mounting your application means. 
This way it´s clear to check that the mount() is after the config. I personal used time to asking why my code is not working and got answer that the config needs to be before mount. After this I readed again the documentation and saw the first lines "You can modify its properties listed below before mounting your application"


NataliaTepluhina (approved):
@vjandrei thank you for the addition 👍🏻

PR #705, opened 2020-11-21
Add an emits option to examples that emit events

### LINKED ISSUE #702: Why v-model on the сustom component throw a warning without emits?
According to the [docs](https://v3.vuejs.org/guide/component-basics.html#using-v-model-on-components) this code should be enough 
**MyCopmponent.vue**

```
<template>
  <button @click="update">+1</button>
</template>

<script>
export default {
  name: 'MyComponent',

  // emits: ['update:modelValue'], <--- without this I got a warning

  props: {
    modelValue: {
      type: Number,
      required: true,
      default: 0
    }
  },

  methods: {
    update () {
      this.$emit('update:modelValue', this.modelValue + 1)
    }
  }
}
</script>
```

but every time I change modelValue I got warning
`[Vue warn]: Component emitted event "update:modelValue" but it is neither declared in the emits option nor as an "onUpdate:modelValue" prop. runtime-core.esm-bundler.js:156 `

Adding **emits: ['update:modelValue']** to the component solve this, but is this behavior correct? I didn't find in the docs adding this instruction is necessary.

I use @vue/cli 4.5.8 and vue 3.0.2

skirtles-code:
**Note:** There is a CodePen at the bottom of `list.md` that should be updated to match the code in this PR.

Closes #702.

I have added an `emits` option to examples that `$emit` events. This isn't strictly necessary for any of these examples, as they function correctly without it. However, using `emits` is encouraged elsewhere in the documentation and omitting it can cause problems in situations that are very similar to those in the documentation.

An event listener registered with `@` or `v-on` will be included in the `$attrs` unless it is listed in the `emits`. In most cases this does no harm but it does result in a listener being added to the root DOM node of the component. So long as the event name doesn't coincide with a DOM event name this is largely harmless, though wasteful.

Events created by `v-model` are slightly different. They are not included in the `$attrs` even if they are omitted from the `emits`. No listener is added to the DOM node.

However, if a component uses the `emits` option then it has to list all events, otherwise there'll be a warning logged. So even `v-model` events will need to be included if a component has any other events. This was why I decided to include `emits` for all examples, even those using `v-model`.

A few other notes:

1. Some `foo` and `bar` references have been changed.
2. I renamed the event `give-advice` to `advise` to dodge the problems around kebab-case vs camelCase (using `emits: ['give-advice']` didn't actually work correctly). That's a problem that needs addressing properly and I didn't want to get into it here.
3. I chopped `const app = Vue.createApp({})` in a few places where it didn't seem necessary. Note that the other examples on that page weren't using it.
4. I added ` app.mount('#emit-example-argument')` to one example. It was a complete example apart from that one line.

### DIFF
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -222,6 +222,7 @@
   })
 
   app.component('welcome-button', {
+    emits: ['welcome'],
     template: `
       <button v-on:click="$emit('welcome')">
         Click me to be welcomed
@@ -236,7 +237,7 @@
 
   ```html
   <div id="emit-example-argument">
-    <advice-component v-on:give-advice="showAdvice"></advice-component>
+    <advice-component v-on:advise="showAdvice"></advice-component>
   </div>
   ```
 
@@ -250,6 +251,7 @@
   })
 
   app.component('advice-component', {
+    emits: ['advise'],
     data() {
       return {
         adviceText: 'Some advice'
@@ -258,12 +260,14 @@
     template: `
       <div>
         <input type="text" v-model="adviceText">
-        <button v-on:click="$emit('give-advice', adviceText)">
+        <button v-on:click="$emit('advise', adviceText)">
           Click me for sending advice
         </button>
       </div>
     `
   })
+  
+  app.mount('#emit-example-argument')
   ```
 
 - **See also:**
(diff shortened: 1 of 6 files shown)


NataliaTepluhina (approved):
@skirtles-code thank you for these changes! CodePen updated as well 👍🏻

PR #708, opened 2020-11-23
Update render-function.md

exarus:
Simplify example for a little bit.

## Description of Problem

`Array.apply(null, { length: n })` is too hard to understand.

## Proposed Solution

`Array.from({ length: n })` is much more straightforward in given context. Arrow functions are used in examples, so we can use ES2015. Example is for advanced users, so some  complexity is fine. Given the textual context the suggested code is clear enough.


NataliaTepluhina (approved):
@exarus thank you! Looks good to me 👍🏻

PR #709, opened 2020-11-24
updated routing.md

SaurabhIsIn:
change the documentation link from https://router.vuejs.org/ to https://next.router.vuejs.org/


NataliaTepluhina (approved):
Thanks for the fix!

PR #712, opened 2020-11-26
Expand the page on render functions, especially their use of slots

skirtles-code:
https://deploy-preview-712--vue-docs-next-preview.netlify.app/guide/render-function.html

I was involved in a couple of threads on the Vue Forum discussing how to use slots in `render` functions. From that it became clear that some key aspects aren't covered in the documentation. I've expanded the relevant section to mention some of the common use cases with explanations of some of the most common pitfalls.

I've added an earlier section explaining the creation of component VNodes. This introduces `resolveComponent`, which was already used in some of the examples but wasn't explained. However, my main objective was to make it more explicit that there are differences between creating VNodes for HTML elements and components. Otherwise the use of component VNodes in the section about slots comes a bit out of the blue.

I've also added a section about how to migrate `<component is="name">`. While it might seem obvious to someone who knows how to do it, I've seen several beginners get completely lost trying to write it as a `render` function. This section also introduces `resolveDynamicComponent`, which I felt needed to be mentioned somewhere.

### DIFF
--- a/src/guide/render-function.md
+++ b/src/guide/render-function.md
@@ -62,9 +62,7 @@ const app = Vue.createApp({})
 
 app.component('anchored-heading', {
   render() {
-    const { h } = Vue
-
-    return h(
+    return Vue.h(
       'h' + this.level, // tag name
       {}, // props/attributes
       this.$slots.default() // array of children
@@ -164,6 +162,8 @@ h(
 )
 ```
 
+If there are no props then the children can usually be passed as the second argument. In cases where that would be ambiguous, `null` can be passed as the second argument to keep the children as the third argument.
+
 ## Complete Example
 
 With this knowledge, we can now finish the component we started:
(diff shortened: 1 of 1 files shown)


src/guide/render-function.md
```diff
@@ -240,6 +240,45 @@ render() {
...
+
+To create a VNode for a component, the first argument passed to `h` should be the component itself:
+
+```js
+render() {
+  return Vue.h(ButtonCounter)
+}
+```
+
+If we need to resolve a component by name then we can call `resolveComponent`:
```
bencodezen, line 253: When are some use cases for resolving a component by name?

You mention later that it's often unnecessary, so as a reader, I'm curious to know a scenario where knowing this helper method would be important.

src/guide/render-function.md
```diff
@@ -346,34 +387,34 @@ render() {
...
   // `<div><slot :text="message"></slot></div>`
-  return Vue.h('div', {}, this.$slots.default({
+  return Vue.h('div', this.$slots.default({
     text: this.message
   }))
 }
 ```
 
-To pass slots to a child component using render functions:
+For component VNodes we need to pass the children to `h` as an Object rather than an Array. Each property is used to populate the slot of the same name:
```
bencodezen, line 409:
```suggestion
For component VNodes, we need to pass the children to `h` as an Object rather than an Array. Each property is used to populate the slot of the same name:
```

src/guide/render-function.md
```diff
@@ -384,6 +425,95 @@ render() {
...
+render() {
+  // Calls to resolveComponent should be outside the slot function
+  const Button = Vue.resolveComponent('MyButton')
+  const Icon = Vue.resolveComponent('MyIcon')
+  
+  return Vue.h(
+    Button,
+    null,
+    {
+      // Using an arrow function to preserve the `this` value
```
bencodezen, line 441:
```suggestion
      // Use an arrow function to preserve the `this` value
```

bencodezen (changes requested):
This is a fantastic enhancement @skirtles-code! Just have a few questions and suggestions for improvements. 

Looking forward to getting this in the docs soon!

src/guide/render-function.md
skirtles-code: I've updated the next few sentences to try to clarify. I struggled to find a wording that wasn't a clumsy mess. It still doesn't flow as cleanly as I'd like but hopefully it's clear enough.

bencodezen (approved):
This is a great iteration @skirtles-code! Going to approve and merge so people can benefit from your work. 

We can always submit improvements if people find anything confusing. Great work!

PR #713, opened 2020-11-26
refactor: move composition api guide to increase discoverability

bencodezen:
## Description of Problem

Users had troubles finding the guide in advanced and expected it to be in Reusability and Composition.

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -58,6 +58,16 @@ const sidebar = {
       title: 'Reusability & Composition',
       collapsable: false,
       children: [
+        {
+          title: 'Composition API',
+          children: [
+            '/guide/composition-api-introduction',
+            '/guide/composition-api-setup',
+            '/guide/composition-api-lifecycle-hooks',
+            '/guide/composition-api-provide-inject',
+            '/guide/composition-api-template-refs'
+          ]
+        },
         '/guide/mixins',
         '/guide/custom-directive',
         '/guide/teleport',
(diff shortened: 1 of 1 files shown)


skirtles-code (commented):
A couple of observations, neither of which prevent this from being merged:

1. There's a certain amount of overlap between the Composition API and Reactivity guides. To me it seems more natural to introduce reactivity prior to composition, though the way the guides are currently written it doesn't seem to rely on either order.
2. The guide to Render Functions should arguably move the other way. If any core topic counts as an 'Advanced Guide' then surely it's that. However, that move would be a little trickier after this PR because the Composition API guide mentions returning a render function from `setup`.

sdras (approved):
Looks great, thanks for following up with this

sdras:
> A couple of observations, neither of which prevent this from being merged:
> 
> 1. There's a certain amount of overlap between the Composition API and Reactivity guides. To me it seems more natural to introduce reactivity prior to composition, though the way the guides are currently written it doesn't seem to rely on either order.
> 2. The guide to Render Functions should arguably move the other way. If any core topic counts as an 'Advanced Guide' then surely it's that. However, that move would be a little trickier after this PR because the Composition API guide mentions returning a render function from `setup`.

Hey @skirtles-code, this is something we discussed in the vue docs meeting we hold once every two weeks- would you like to attend? If so, please email me: sarah.drasner@gmail.com, we think it would be great to have you there.

We discussed what you were saying, but were also solving for some discoverability issues.

skirtles-code:
@sdras I'm having trouble sending you an email (my problem, not yours, boring story). If you're on Vue Land Discord I'm available there as skirtle. Or on the Vue forum.

sdras:
@skirtles-code thanks! I'll go find you there.

PR #717, opened 2020-11-28
Update component-custom-events.md

V1RoZ:
## Description of Problem
Match sample text in description and HTML example


skirtles-code (approved):
My mistake. I missed that when I updated the example last week.

bencodezen (approved):
Good catch @V1RoZ!

PR #718, opened 2020-11-28
Updated vm.pick condition for Radio

snehansh:
when 'radio' is checked, `vm.pick` should be `a`.


skirtles-code:
Thank you for the PR but I don't think this change is correct.

The preceding code uses `v-bind:value="a"`, which binds the value of the `a` property to the radio. It doesn't bind the string `'a'`.

PR #719, opened 2020-11-29
feat: added `created` hook to directives

NataliaTepluhina:
## Description of Problem

Current documentation misses the `created` custom directives hook added in https://github.com/vuejs/vue-next/commit/11804fe93f66d43320498bfea564af1c9f7a9eb7

## Proposed Solution

I've added a new hook to API, Migration guide and Guide

/close https://github.com/vuejs/docs-next/issues/430

### DIFF
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -88,7 +88,7 @@ const app = createApp({})
 // register
 app.directive('my-directive', {
   // Directive has a set of lifecycle hooks:
-  // called before bound element's parent component props are applied
+  // called before bound element's attributes or event listeners are applied
   created() {},
   // called before bound element's parent component is mounted
   beforeMount() {},
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -22,7 +22,7 @@ In addition, since the `createApp` method returns the application instance itsel
 - **Returns:**
 
   - The application instance if a `definition` argument was passed
-  - The component definition if a `definition` argument was not passed 
+  - The component definition if a `definition` argument was not passed
 
 - **Usage:**
 
@@ -73,7 +73,7 @@ app.config = {...}
 - **Returns:**
 
   - The application instance if a `definition` argument was passed
-  - The directive definition if a `definition` argument was not passed 
+  - The directive definition if a `definition` argument was not passed
 
 - **Usage:**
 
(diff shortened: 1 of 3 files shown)


skirtles-code (commented):
I've added some thoughts to a couple of lines.

As an aside, it seems I'm to blame for adding a lot of unnecessary spaces at the ends of lines. My text editor is faithfully respecting the settings in `.editorconfig`, which includes the following:

```text
[*.md]
trim_trailing_whitespace = false
```

I understand why it's there but it is causing these extra spaces in some of my contributions. Locally I've now hacked that file so, hopefully, any future changes will be trimmed.

src/api/application-api.md
```diff
@@ -88,6 +88,8 @@ const app = createApp({})
 // register
 app.directive('my-directive', {
   // Directive has a set of lifecycle hooks:
+  // called before bound element's parent component props are applied
```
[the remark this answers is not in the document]
NataliaTepluhina: @skirtles-code it's used the same way for other lifecycle hooks, do you think it'd be better to replace it to `element's parent VNode`?

skirtles-code, line 91:
```suggestion
  // called before bound element's attributes or event listeners are applied
```

src/guide/custom-directive.md
```diff
@@ -45,6 +45,8 @@ Then in a template, you can use the new `v-focus` attribute on any element, like
 
 A directive definition object can provide several hook functions (all optional):
 
+- `created`: called before bound element's parent component props are applied. This is useful in cases where the directive needs to attach event listeners before parent component's prop or template listeners.
```
skirtles-code, line 48:
```suggestion
- `created`: called before the bound element's attributes or event listeners are applied. This is useful in cases where the directive needs to attach event listeners that must be called before normal `v-on` event listeners.
```

src/guide/migration/custom-directives.md
```diff
@@ -39,6 +39,7 @@ Here, in the initial setup for this element, the directive binds a style by pass
 
 In Vue 3, however, we’ve created a more cohesive API for custom directives. As you can see, they differ greatly from our component lifecycle methods even though we’re hooking into similar events. We’ve now unified them like so:
 
+- **created** - new! This is called before bound element's parent component props are applied
```
skirtles-code, line 42:
```suggestion
- **created** - new! This is called before the element's attributes or event listeners are applied.
```

skirtles-code (commented):
Sorry for the delay. This has been buzzing around at the back of my mind all week but I wasn't able to think of a way to write it that I'm entirely happy with.

I've made some concrete suggestions this time. I've removed any mention of the parent component or props and have instead used the phrase *'element's attributes or event listeners'*. I think that's an accurate description and conveys it in a way that will be relatable to most readers.

skirtles-code (approved):

PR #723, opened 2020-12-02
docs: Add type definition for `globalProperties` in TypeScript chapter.

dev-itsheng:
## Description of Problem

See https://github.com/vuejs/vue-next/issues/2220 and https://github.com/vuejs/vue-next/issues/2701 .

The lack of documentation have made many people wonder how to solve TypeScript errors caused by custom types in `globalProperties`, so I tried to fix it.

## Proposed Solution

I added a chapter to the document to explain the origin of this problem and how to solve it.

## Additional Information

In addition, I also found a small sample code syntax error in this document and corrected it.

> In an object, the lack of a comma between the methods will cause a syntax error.

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -135,9 +135,9 @@ const Component = defineComponent({
 
 ### Augmenting Types for Use with Plugins or Declarations by User
 
-Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-a-plugin) to add to Vue’s global properties in each component instance. In some cases, we also need to [do it by ourselves](../api/application-config.html#globalproperties):
+Vue 3 provides a [`globalProperties` object](../api/application-config.html#globalproperties) that can be used to add a global property that can be accessed in any component instance. For example, a [plugin](./plugins.html#writing-a-plugin) might want to inject a shared global object or function.
 
-```typescript
+```ts
 // User Definition
 import axios from 'axios'
 
@@ -154,11 +154,11 @@ export default {
 }
 ```
 
-In order to let TypeScript know the properties we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
+In order to tell TypeScript about these new properties, we can use [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).
 
-For the above example, we (or plugins) can add the following code of type definition:
+In the above example, we could add the following type declaration:
 
-```typescript
+```ts
 import axios from 'axios'
 
 declare module '@vue/runtime-core' {
@@ -168,9 +168,9 @@ declare module '@vue/runtime-core' {
 }
 ```
 
-We should put the above code wherever it can be loaded, such as `main.ts` where you define these properties, or a `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it should be placed in the location specified by the `types` property in `package.json`. 
+We can put this type declaration in the same file, or in a project-wide `*.d.ts` file (for example, in the `src/typings` folder that is automatically loaded by TypeScript). For library/plugin authors, this file should be specified in the `types` property in `package.json`.
 
-The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/2587f36fe311359e2e34f40e8e47d2eebfab7f42/packages/runtime-core/src/componentOptions.ts#L64-L80) and [unit tests for TypeScript types](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
+For more information about the `ComponentCustomProperties` type, see its [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/2587f36fe311359e2e34f40e8e47d2eebfab7f42/packages/runtime-core/src/componentOptions.ts#L64-L80) and [the TypeScript unit tests](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
 
 ### Annotating Return Types
 
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -139,14 +139,18 @@ Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-
 
 ```typescript
 // User Definition
-const app = Vue.createApp()
-app.config.globalProperties.foo = 'bar'
+import axios from 'axios'
 
-// Plugin
+const app = Vue.createApp()
+app.config.globalProperties.$http = axios
+
+// Plugin for validate some data
 export default {
-    install(app, options) {
-        app.config.globalProperties.foo = 'bar'
+  install(app, options) {
+    app.config.globalProperties.$validate = (data: object, rule: object) => {
+      // check whether the object meets certain rules
     }
+  }
 }
 ```
 
@@ -155,16 +159,20 @@ In order to let TypeScript know the attributes we (or plugins) added and their t
 For the above example, we (or plugins) can add the following code of type definition:
 
 ```typescript
+import axios from 'axios'
+
 declare module '@vue/runtime-core' {
-    export interface ComponentCustomProperties {
-        foo: string
-    }
+  export interface ComponentCustomProperties {
+    axios: typeof axios
+  }
 }
 ```
 
-Then you can use it in TypeScript by correct type inference.
+We should put the above code wherever it can be loaded, such as `main.js` where you define these properties, or the `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it locates in the location specified by the `types` property in `package.json`. 
 
-The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L63-L79), [use](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L97) and [unit test](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) in `@vue/runtime-core` to learn more.
+After TypeScript loaded them, we can use it in TypeScript by correct type inference.
+
+The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L63-L111) and [unit tests for Typescript types](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
 
 ### Annotating Return Types
 
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -133,7 +133,7 @@ const Component = defineComponent({
 })
 ```
 
-### Augmenting Types for Use with Plugins or User Definitions
+### Augmenting Types for Use with Plugins or Declarations by User
 
 Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-a-plugin) to add to Vue’s global properties in each component instance. In some cases, we also need to [do it by ourselves](../api/application-config.html#globalproperties):
 
@@ -141,10 +141,10 @@ Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-
 // User Definition
 import axios from 'axios'
 
-const app = Vue.createApp()
+const app = Vue.createApp({})
 app.config.globalProperties.$http = axios
 
-// Plugin for validate some data
+// Plugin for validating some data
 export default {
   install(app, options) {
     app.config.globalProperties.$validate = (data: object, rule: object) => {
@@ -154,7 +154,7 @@ export default {
 }
 ```
 
-In order to let TypeScript know the attributes we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
+In order to let TypeScript know the properties we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
 
 For the above example, we (or plugins) can add the following code of type definition:
 
@@ -168,11 +168,9 @@ declare module '@vue/runtime-core' {
 }
 ```
 
-We should put the above code wherever it can be loaded, such as `main.js` where you define these properties, or the `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it locates in the location specified by the `types` property in `package.json`. 
+We should put the above code wherever it can be loaded, such as `main.ts` where you define these properties, or a `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it should be placed in the location specified by the `types` property in `package.json`. 
 
-After TypeScript loaded them, we can use it in TypeScript by correct type inference.
-
-The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L63-L111) and [unit tests for Typescript types](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
+The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/2587f36fe311359e2e34f40e8e47d2eebfab7f42/packages/runtime-core/src/componentOptions.ts#L64-L80) and [unit tests for TypeScript types](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
 
 ### Annotating Return Types
 
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -133,6 +133,39 @@ const Component = defineComponent({
 })
 ```
 
+### Augmenting Types for Use with Plugins or User Definitions
+
+Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-a-plugin) to add to Vue’s global properties in each component instance. In some cases, we also need to [do it by ourselves](../api/application-config.html#globalproperties):
+
+```typescript
+// User Definition
+const app = Vue.createApp()
+app.config.globalProperties.foo = 'bar'
+
+// Plugin
+export default {
+    install(app, options) {
+        app.config.globalProperties.foo = 'bar'
+    }
+}
+```
+
+In order to let TypeScript know the attributes we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
+
+For the above example, we (or plugins) can add the following code of type definition:
+
+```typescript
+declare module '@vue/runtime-core' {
+    export interface ComponentCustomProperties {
+        foo: string
+    }
+}
+```
+
+Then you can use it in TypeScript by correct type inference.
+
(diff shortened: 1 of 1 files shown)


src/guide/typescript-support.md
```diff
@@ -133,6 +133,39 @@ const Component = defineComponent({
...
 ```
 
+### Augmenting Types for Use with Plugins or User Definitions
+
+Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-a-plugin) to add to Vue’s global properties in each component instance. In some cases, we also need to [do it by ourselves](../api/application-config.html#globalproperties):
+
+```typescript
+// User Definition
+const app = Vue.createApp()
+app.config.globalProperties.foo = 'bar'
```
NataliaTepluhina, line 143: **suggestion**: we try to avoid abstract meaningless names for variables in docs. Would you mind changing `foo` and `bar` to something more 'real-life'? 😅

src/guide/typescript-support.md
```diff
@@ -133,6 +133,39 @@ const Component = defineComponent({
...
+
+// Plugin
+export default {
+    install(app, options) {
```
NataliaTepluhina, line 147: **nitpick**: we use 2-spaces for the indentation

src/guide/typescript-support.md
```diff
@@ -133,6 +133,39 @@ const Component = defineComponent({
...
+        app.config.globalProperties.foo = 'bar'
+    }
+}
+```
+
+In order to let TypeScript know the attributes we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
+
+For the above example, we (or plugins) can add the following code of type definition:
```
NataliaTepluhina, line 155: **suggestion**: should we explain where we add type definitions?

src/guide/typescript-support.md
```diff
@@ -133,6 +133,39 @@ const Component = defineComponent({
...
+declare module '@vue/runtime-core' {
+    export interface ComponentCustomProperties {
+        foo: string
+    }
+}
+```
+
+Then you can use it in TypeScript by correct type inference.
+
+The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L63-L79), [use](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L97) and [unit test](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) in `@vue/runtime-core` to learn more.
```
NataliaTepluhina, line 167: **suggestion**: I believe this part goes into too many details. We can shorten it to `Please check the source code of `@vue/runtime-core` to learn more`, what do you think?

NataliaTepluhina (changes requested):
@dev-itsheng thank you for this awesome addition! I've left a few comments here and there, back to you for discussion

src/guide/typescript-support.md
dev-itsheng: The definiton and test in different folders, so I merged the first two links and made additional explanations for the last one.

dev-itsheng:
I have fixed them and added more explanation.

src/guide/typescript-support.md
```diff
@@ -133,6 +133,47 @@ const Component = defineComponent({
 })
 ```
 
+### Augmenting Types for Use with Plugins or User Definitions
```
skirtles-code, line 136: I don't use TypeScript myself so this might be a really dumb question, but is 'User Definitions' a technical term? I don't really understand what it means here (or where it's used in the example).

src/guide/typescript-support.md
```diff
@@ -133,6 +133,47 @@ const Component = defineComponent({
...
+Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-a-plugin) to add to Vue’s global properties in each component instance. In some cases, we also need to [do it by ourselves](../api/application-config.html#globalproperties):
+
+```typescript
+// User Definition
+import axios from 'axios'
+
+const app = Vue.createApp()
+app.config.globalProperties.$http = axios
+
+// Plugin for validate some data
```
skirtles-code, line 147:
```suggestion
// Plugin for validating some data
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,47 @@ const Component = defineComponent({
...
+import axios from 'axios'
+
+declare module '@vue/runtime-core' {
+  export interface ComponentCustomProperties {
+    axios: typeof axios
+  }
+}
+```
+
+We should put the above code wherever it can be loaded, such as `main.js` where you define these properties, or the `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it locates in the location specified by the `types` property in `package.json`.
```
skirtles-code, line 171:
```suggestion
We should put the above code wherever it can be loaded, such as `main.ts` where you define these properties, or a `*.d.ts` file in the `src/typings` folder automatically loaded by TypeScript. For plugins, it should be placed in the location specified by the `types` property in `package.json`. 
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,47 @@ const Component = defineComponent({
...
+  export interface ComponentCustomProperties {
+    axios: typeof axios
+  }
+}
+```
+
+We should put the above code wherever it can be loaded, such as `main.js` where you define these properties, or the `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it locates in the location specified by the `types` property in `package.json`. 
+
+After TypeScript loaded them, we can use it in TypeScript by correct type inference.
+
```
skirtles-code, line 174:
```suggestion
```

I would remove this sentence. I don't think it really adds anything.

src/guide/typescript-support.md
```diff
@@ -133,6 +133,47 @@ const Component = defineComponent({
...
+The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L63-L111) and [unit tests for Typescript types](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
```
skirtles-code, line 175:
```suggestion
The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/componentOptions.ts#L63-L111) and [unit tests for TypeScript types](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
```

I'm also wondering about the first link. It points to specific lines on the `master` branch. I'm not sure what the best approach is here but those lines could easily move the next time that file is changed on `master`.

src/guide/typescript-support.md
```diff
@@ -133,6 +133,47 @@ const Component = defineComponent({
...
+export default {
+  install(app, options) {
+    app.config.globalProperties.$validate = (data: object, rule: object) => {
+      // check whether the object meets certain rules
+    }
+  }
+}
+```
+
+In order to let TypeScript know the attributes we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
```
skirtles-code, line 157: Is 'attributes' a TypeScript term? Should it be 'properties' instead?

src/guide/typescript-support.md
```diff
@@ -133,6 +133,47 @@ const Component = defineComponent({
...
 
+### Augmenting Types for Use with Plugins or User Definitions
+
+Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-a-plugin) to add to Vue’s global properties in each component instance. In some cases, we also need to [do it by ourselves](../api/application-config.html#globalproperties):
+
+```typescript
+// User Definition
+import axios from 'axios'
+
+const app = Vue.createApp()
```
skirtles-code, line 144:
```suggestion
const app = Vue.createApp({})
```

I believe an empty object has to be passed to avoid an error being thrown when the app is mounted.

skirtles-code (changes requested):
I've tried to provide some feedback but it may not all be entirely helpful.

I don't use TypeScript myself and I'm largely unfamiliar with the technical terms it uses and the idiomatic way to describe those concepts in English. I've had to use some educated guesses to make suggestions.

src/guide/typescript-support.md
dev-itsheng: My English is so poor that I can't describe my thoughts accurately. But I read the relevant documentation of TypeScript to learn the exact definition, and found that they use "declaration" to describe a type, so I think it is better to replace it with "declarations by user".

src/guide/typescript-support.md
dev-itsheng: I have fixed them and left some thoughts, Thanks for your suggestions.

dev-itsheng:
I have fixed them and left some thoughts, Thanks for your suggestions.

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
 })
 ```
 
+### Augmenting Types for Use with Plugins or Declarations by User
+
+Vue 3 provides `globalProperties` property for [plugins](./plugins.html#writing-a-plugin) to add to Vue’s global properties in each component instance. In some cases, we also need to [do it by ourselves](../api/application-config.html#globalproperties):
```
danielroe, line 138:
```suggestion
Vue 3 provides a [`globalProperties` object](../api/application-config.html#globalproperties) that can be used to add a global property that can be accessed in any component instance. For example, a [plugin](./plugins.html#writing-a-plugin) might want to inject a shared global object or function.
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+
+```typescript
```
danielroe, line 140:
```suggestion
```ts
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+    }
+  }
+}
+```
+
+In order to let TypeScript know the properties we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
+
+For the above example, we (or plugins) can add the following code of type definition:
+
+```typescript
```
danielroe, line 161:
```suggestion
```ts
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+export default {
+  install(app, options) {
+    app.config.globalProperties.$validate = (data: object, rule: object) => {
+      // check whether the object meets certain rules
+    }
+  }
+}
+```
+
+In order to let TypeScript know the properties we (or plugins) added and their types, the [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature can help us.
```
danielroe, line 157:
```suggestion
In order to tell TypeScript about these new properties, we can use [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+
+For the above example, we (or plugins) can add the following code of type definition:
```
danielroe, line 159:
```suggestion
In the above example, we could add the following type declaration:
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+import axios from 'axios'
+
+declare module '@vue/runtime-core' {
+  export interface ComponentCustomProperties {
+    axios: typeof axios
+  }
+}
+```
+
+We should put the above code wherever it can be loaded, such as `main.ts` where you define these properties, or a `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it should be placed in the location specified by the `types` property in `package.json`.
```
danielroe, line 171:
```suggestion
We can put this type declaration in the same file, or in a project-wide `*.d.ts` file (for example, in the `src/typings` folder that is automatically loaded by TypeScript). For library/plugin authors, this file should be specified in the `types` property in `package.json`.
```

src/guide/typescript-support.md
```diff
@@ -150,7 +189,7 @@ const Component = defineComponent({
     // needs an annotation
     greeting(): string {
       return this.message + '!'
-    }
+    },
```
danielroe, line 192:
```suggestion
    }
```

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+declare module '@vue/runtime-core' {
+  export interface ComponentCustomProperties {
+    axios: typeof axios
+  }
+}
+```
+
+We should put the above code wherever it can be loaded, such as `main.ts` where you define these properties, or a `*.d.ts` file in `src/typings` folder automatically loaded by TypeScript. For plugins, it should be placed in the location specified by the `types` property in `package.json`. 
+
+The `ComponentCustomProperties` type can help us define the custom options, see the code of [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/2587f36fe311359e2e34f40e8e47d2eebfab7f42/packages/runtime-core/src/componentOptions.ts#L64-L80) and [unit tests for TypeScript types](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
```
danielroe, line 173:
```suggestion
For more information about the `ComponentCustomProperties` type, see its [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/2587f36fe311359e2e34f40e8e47d2eebfab7f42/packages/runtime-core/src/componentOptions.ts#L64-L80) and [the TypeScript unit tests](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
```

src/guide/typescript-support.md
LinusBorg: Adding this to a .d.ts file is not a good idea in my experience, since that will effectively re-declare the runtime-core modules which means all the types that this module exports will be no longer accessible.

danielroe: That is only the case if nothing is imported into the module declaration file, or if the user attempts to extend a type rather than an interface.

dev-itsheng: According to TypeScript docs, the declarations in an augmentation will be **merged** rather than **replaced**, whether it is declared in `.ts` or `.d.ts` file.

From another perspective, the developer of the library/plugin can only generate the type declaration to a separate `.d.ts` file (because the final release is the JS file packaged by the rollup), such as [the Vue-Router Type declaration file (Please see the end of the file)](https://www.unpkg.com/vue-router@4.0.5/dist/vue-router.d.ts).

danielroe: @dev-itsheng If there is no other import in the file then it is not treated as a module and the behaviour that @LinusBorg describes occurs.

dev-itsheng:
Thanks for the guidance on modifying the content style. I have corrected everything except that comma. Regarding the issue of coding style, please read the comments, I have some different opinions.

src/guide/typescript-support.md
dev-itsheng: Yes, I just tried some cases. Indeed, if only use `declare module` in `d.ts` (and `.ts`) files, the declaration will be overwritten instead of merged. When I add an `import` (no matter what I import), it will return to normal.

Then I checked the TypeScript documentation. It did not describe my scenario, but mentioned in [another place](https://www.typescriptlang.org/docs/handbook/modules.html) that files with top-level import/export would be regarded as modules, otherwise they would be scripts. I guess TypeScript might be able to deal with these two Different file types are handled differently.

So, I used `export` instead of `import` (such as `export const a: number` or `export const a = 1`) and found that it still works.

In this case, I think it is necessary to explain it in the document, too.

src/guide/typescript-support.md
dev-itsheng: When I saw this document for the first time, I realized that there is something unusual about this code, which is different from other codes.

First of all, it is mentioned here, it writes a comma less, if there is a missing comma between the method and method of the JS object declaration, it will throw a syntax error. This is such a small defect, so I am handy Just fixed it.

Then when I re-read this code, I found more problems. It is different from other code styles in the document. In addition to the comma above, there are two omitable commas below (according to the style of other codes, it should Is omitted), but this does not cause a syntax error, so I didn't notice it at first.

```ts
import { defineComponent } from 'vue'

const Component = defineComponent({
  data() {
    return {
      message: 'Hello!'
    }
  },
  computed: {
    // needs an annotation
    greeting(): string {
      return this.message + '!'
    }     // <--------------------------------------------- There should be a comma here

    // in a computed with a setter, getter needs to be annotated
    greetingUppercased: {
      get(): string {
        return this.greeting.toUpperCase();
      },
      set(newValue: string) {
        this.message = newValue.toUpperCase();
      },  // <--------------------------------------------- There should not be a comma here
    },    // <--------------------------------------------- There should not be a comma here, too
  }
})
```

I think they should be corrected, either here or in the next PR.

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+In order to tell TypeScript about these new properties, we can use [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).
+
+In the above example, we could add the following type declaration:
+
+```ts
+import axios from 'axios'
+
+declare module '@vue/runtime-core' {
+  export interface ComponentCustomProperties {
+    axios: typeof axios
```
IlCallo, line 166: Shouldn't this be `$http: typeof axios` to match previous code?

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
...
+  }
+}
```
lmiller1990, line 168: Might be good to include a type of `app.config.globalProperties.$validate` since you included that in your example above, too.

lmiller1990 (changes requested):

src/guide/typescript-support.md
```diff
@@ -133,6 +133,45 @@ const Component = defineComponent({
 })
 ```
 
+### Augmenting Types for Use with Plugins or Declarations by User
```
skirtles-code, line 136: I'm wondering whether something like this might give a clearer sense of what follows?

```suggestion
### Augmenting Types for `globalProperties`
```

If I've understood correctly then that's what this section is really about.

skirtles-code:
@dev-itsheng In case you were wondering, this PR was discussed on a docs team call last week. We reached out for some help reviewing it, which is why it suddenly received so much attention.

When you have time, could you take a look at the final few suggestions that have been made? I think once those 3 remaining suggestions have been handled we should be able to merge this.

Thanks for all your work on this.

src/guide/typescript-support.md
dev-itsheng: Yes, at first I used "Use with Plugins" because the [documentation of Vue 2.x](https://vuejs.org/v2/guide/typescript.html#Augmenting-Types-for-Use-with-Plugins) uses it, so I copied from the old document. Next, I realized that not only plugins, users can also define some properties (This is exactly what I realized that the documentation was not clear about this point), so I add "or Declarations by User".

In fact, I did not realize the essential meaning of the problem before, whether it is a user or a plug-in, they are all "add" global properties.

dev-itsheng:
Thanks for your attention.

I have made some changes and added some explanations about the questions you raised.

src/guide/typescript-support.md
```diff
@@ -133,6 +133,52 @@ const Component = defineComponent({
...
+}
+```
+
+We can put this type declaration in the same file, or in a project-wide `*.d.ts` file (for example, in the `src/typings` folder that is automatically loaded by TypeScript). For library/plugin authors, this file should be specified in the `types` property in `package.json`.
+
+::: warning Make sure the declaration file is a TypeScript module
+According to the [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/modules.html), any file containing a top-level `import` or `export` is considered a module. Otherwise, it is a script. Only in module, the declarations in an augmentation will be merged, if in script, they will replace the original types, it will make all the default properties missing.
+
+Please make sure there is at least one top-level `import` or `export` in your file -- Especially when you customize the type instead of extending a third-party type in a separate file, at this time, you can add `export {}` to meet the requirements.
+:::
```
danielroe, line 178: Some suggested wording:

```suggestion
::: warning Make sure the declaration file is a TypeScript module
In order to take advantage of module augmentation, you will need to ensure there is at least one top-level `import` or `export` in your file, even if it is just `export {}`.

[In TypeScript](https://www.typescriptlang.org/docs/handbook/modules.html), any file containing a top-level `import` or `export` is considered a 'module'. If type declaration is made outside of a module, it will overwrite the original types rather than augmenting them.
:::
```

skirtles-code (approved):
I think this is good to merge. Thanks to everyone involved, especially @dev-itsheng.

PR #724, opened 2020-12-02
feat: add carbon ads

### LINKED ISSUE #30: Integrate Carbon ads

posva:
## Additional Information

Closes #30 

I couldn't test locally because the `serve` script fails right away

There are a few necessary changes left to be done to display the ads correctly and mimic vuejs.org

### DIFF
--- a/src/.vuepress/theme/layouts/Layout.vue
+++ b/src/.vuepress/theme/layouts/Layout.vue
@@ -17,12 +17,6 @@
     <Sidebar :items="sidebarItems" @toggle-sidebar="toggleSidebar">
       <template #top>
         <slot name="sidebar-top" />
-        <!-- <CarbonAds
-          v-if="$site.themeConfig.carbonAds"
-          :key="'ca:' + $page.path"
-          :code="$site.themeConfig.carbonAds.carbon"
-          :placement="$site.themeConfig.carbonAds.placement"
-        /> -->
       </template>
       <template #bottom>
         <slot name="sidebar-bottom" />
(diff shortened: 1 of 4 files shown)


src/.vuepress/theme/layouts/Layout.vue
```diff
@@ -192,10 +216,26 @@ export default {
...
+      window[method]('scroll', this.fixMenuAfterBanner)
     }
   }
 }
 </script>
+
+<style>
+/* add a placeholder to give space to ads so they don't overlap with content */
+@media screen and (max-width: 1376px) {
+  .content__default::before {
```
posva, line 230: This to make the ad take some space

vuejs.org, ad inside content

Screen Shot 2020-12-02 at 13 25 04

vuepress doesn't:

Screen Shot 2020-12-02 at 13 25 11


But it creates overlap for tip containers unless we add `overflow-x: auto` but that breaks the icon display:

Screen Shot 2020-12-02 at 13 20 02

Which is the reason I added the background

src/.vuepress/theme/layouts/Layout.vue
```diff
@@ -77,7 +97,7 @@ export default {
       )
     },
 
-    shouldShowSidebar () {
+    shouldShowSidebar() {
```
posva, line 100: Apparently other functions are also following this convention. I simply forgot to turn off auto format when saving but I can revert these changes

NataliaTepluhina:
@posva looks really nice! Should we remove the `Draft` state and merge the PR?

NataliaTepluhina (approved):

posva:
Is the background solution satisfactory enough?
I'm only worried because I'm using hacky stuff in CSS but I couldn't figure out a different way 😓

NataliaTepluhina:
@posva I think it's justified to be hacky 😅

posva:
Okay, I thought the same 😅

PR #730, opened 2020-12-05
docs(ts-support): annotating props with validators and default values

JensD98:
## Description of Problem
The example under [Annotating Props](https://v3.vuejs.org/guide/typescript-support.html#annotating-props) in the TypeScript Support section of the docs would break the prop type for the component because of the `validator` function.
## Proposed Solution
Give an example of how to correctly use `validators` and `default` values for objects and arrays.
## Additional Information
[Related issue](https://github.com/vuejs/vue-next/issues/2474)

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -172,10 +172,10 @@ Vue does a runtime validation on props with a `type` defined. To provide these t
 ```ts
 import { defineComponent, PropType } from 'vue'
 
-interface ComplexMessage {
+interface Book {
   title: string
-  okMessage: string
-  cancelMessage: string
+  author: string
+  year: number
 }
 
 const Component = defineComponent({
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@JensD98 thank you so much for this addition! The only fix here: could we reuse the example from above instead of changing it to use a different prop so readers would have a smooth transition?

JensD98:
@NataliaTepluhina I can see what you mean. However, I feel like having an object with three properties for demonstrating default values two times can get a bit lengthy. Would adding the numbers array prop to the example above already help?

JensD98:
I changed both examples to use the `Book` interface. Maybe this makes the most sense because it was already used a lot in other examples throughout the chapter.

NataliaTepluhina (approved):
@JensD98 thank you for the applied fixes! Looks good to me 👍🏻

PR #732, opened 2020-12-05
doc(api/basic-reactivity): fix syntax highlighting

cRAN-cg:
## Description of Problem
Improper use of syntax highlighting for fenced code blocks

## Proposed Solution
Remove meta after the specified language keyword (before the fenced code blocks)


skirtles-code:
Could you provide some more details about why you think this change is necessary? The existing line, ` ```js{7-15} `, indicates that lines 7 to 15 should be highlighted with a different background colour and that does appear to be working. Are you suggesting that those lines shouldn't be highlighted?

cRAN-cg:
Aah my bad, the markdown compiler used by Github is different than what is used on the docs site. Closing this.

Issue #733, opened 2020-12-05
`watch` and proxied objects


fimion:
So I've been doing some exploring with the composition api,  `watch`,  and reactive arrays. Since the array is wrapped in a proxy, if you attempt to use `watch` on the array, both the old and the new values given to you point to the proxied object and are identical.

I'd like to add a section to the [watch guide](https://v3.vuejs.org/guide/reactivity-computed-watchers.html#watch) about how to do proxied object comparisons. For instance, if you are wanting to compare the changes of a proxied array, you would need to provide `watch` a function that shallow copies the values of the array.

```js
const myArray = reactive([1,2,3,4]);
watch(()=>[...myArray], (newVal, oldVal) =>{
   // do things...
})
```
also provide an example for objects.

maybe some wording about utilities that can be used to deeply copy an object/array, such as `lodash.cloneDeep`

PR #734, opened 2020-12-05
Adding documentation for watch and proxied objects

fimion:
## Description of Problem

The way a developer needs to use watch on a proxied object is not well documented. I have outlined more details in #733

## Proposed Solution

Add details about shallow copying and deep copying an object to see both the old and new values of it.

### DIFF
--- a/src/guide/reactivity-computed-watchers.md
+++ b/src/guide/reactivity-computed-watchers.md
@@ -262,7 +262,7 @@ watch(
 state[0].attributes.name = "Alex"; // Logs: "deep " ["Alex", ""] ["Alex", ""]
 ```
 
-However, watching a reactive Object or Array will always return a reference to the current value of that Object. To fully watch deeply nested Objects and Arrays, a deep copy of values may be required. This can be achieved with a utility such as [lodash.cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep)
+However, watching a reactive object or array will always return a reference to the current value of that object. To fully watch deeply nested objects and arrays, a deep copy of values may be required. This can be achieved with a utility such as [lodash.cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep)
 
 ```js
 import _ from 'lodash';
--- a/src/guide/reactivity-computed-watchers.md
+++ b/src/guide/reactivity-computed-watchers.md
@@ -228,7 +228,7 @@ watch(
 numbers.push(5) // logs: [1,2,3,4,5] [1,2,3,4]
 ```
 
-Attempting to check for changes of properties in a deeply nested Object or Array will still require the `deep` option to be true, if watching an Object or Array.
+Attempting to check for changes of properties in a deeply nested object or array will still require the `deep` option to be true:
 
 ```js
 const state = reactive([
--- a/src/guide/reactivity-computed-watchers.md
+++ b/src/guide/reactivity-computed-watchers.md
@@ -4,11 +4,7 @@
 
 ## Computed values
 
-Sometimes we need state that depends on other state - in Vue this is handled with
-component [computed properties](computed.html#computed-properties). To directly create a computed value, we can use
-the `computed` method: it takes a getter function and returns an immutable
-reactive [ref](reactivity-fundamentals.html#creating-standalone-reactive-values-as-refs) object for the returned value
-from the getter.
+Sometimes we need state that depends on other state - in Vue this is handled with component [computed properties](computed.html#computed-properties). To directly create a computed value, we can use the `computed` method: it takes a getter function and returns an immutable reactive [ref](reactivity-fundamentals.html#creating-standalone-reactive-values-as-refs) object for the returned value from the getter.
 
 ```js
 const count = ref(1)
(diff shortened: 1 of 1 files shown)


fimion:
@bencodezen told me to tag him in this as well.

src/guide/reactivity-computed-watchers.md
```diff
@@ -205,6 +205,24 @@ watch([fooRef, barRef], ([foo, bar], [prevFoo, prevBar]) => {
...
+watch(
+  () => [...numbers],
+  (numbers, prevNumbers) =>{
+    console.log(numbers, prevNumbers);
+})
+
+numbers.push(5) // logs: [1,2,3,4,5] [1,2,3,4]
+```
+
+For deeply nested object and arrays, a deep copy of values may be required. This can be achieved with a utility such as `lodash.cloneDeep`
```
NataliaTepluhina, line 224: **suggestion**: as we're assuming some non-Vue knowledge here, it would be nice to add a link:
```suggestion
For deeply nested object and arrays, a deep copy of values may be required. This can be achieved with a utility such as [lodash.cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep)
```

src/guide/reactivity-computed-watchers.md
```diff
@@ -205,6 +205,24 @@ watch([fooRef, barRef], ([foo, bar], [prevFoo, prevBar]) => {
 })
 ```
 
+### Watching Reactive Objects
+
+Using a watcher to compare values of an Array or Object that are reactive requires that it has a copy made of just the values.
```
NataliaTepluhina, line 210: **question**: would `deep` watcher option work with an Object in this case?

NataliaTepluhina (commented):
@fimion thank you for working on this! I've added a few comments to the PR

src/guide/reactivity-computed-watchers.md
fimion: so I hadn't specifically been testing `deep`, and I just made an experiment where I test deep vs not deep

https://codepen.io/fimion/pen/XWjdYpB

the behavior between the 2 are the same. I don't know if this is a bug, or if I am not nesting things correctly to trigger a need for `deep`.

fimion: I commented and then I tested it with the function and now I see the difference. Let me rethink some of this and make sure I've got the edges covered.

fimion: So, I've tried to add some clarification to this about `deep` and some of the weirdness of proxied objects. Let me know if I need to clarify it more?

src/guide/reactivity-computed-watchers.md
```diff
@@ -211,6 +212,79 @@ firstName.value = "John"; // logs: ["John",""] ["", ""]
...
+      prevAttrs.map((e) => e.name)
+    );
+  },
+  { deep: true }
+)
+
+state[0].attributes.name = "Alex"; // Logs: "deep " ["Alex", ""] ["Alex", ""]
+```
+
+However, watching a reactive Object or Array will always return a reference to the current value of that Object. To fully watch deeply nested Objects and Arrays, a deep copy of values may be required. This can be achieved with a utility such as [lodash.cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep)
```
NataliaTepluhina, line 265:
```suggestion
However, watching a reactive object or array will always return a reference to the current value of that object. To fully watch deeply nested objects and arrays, a deep copy of values may be required. This can be achieved with a utility such as [lodash.cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep)
```

NataliaTepluhina (commented):
@fimion thank you for all the changes! I've added a few more points to review

src/guide/reactivity-computed-watchers.md
```diff
@@ -211,6 +212,79 @@ firstName.value = "John"; // logs: ["John",""] ["", ""]
...
+watch(
+  () => [...numbers],
+  (numbers, prevNumbers) => {
+    console.log(numbers, prevNumbers);
+  })
+
+numbers.push(5) // logs: [1,2,3,4,5] [1,2,3,4]
+```
+
+Attempting to check for changes of properties in a deeply nested Object or Array will still require the `deep` option to be true, if watching an Object or Array.
```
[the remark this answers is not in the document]
fimion: Ah cool, I wasn't entirely sure which way y'all went on that.

src/guide/reactivity-computed-watchers.md
```diff
@@ -211,6 +212,79 @@ firstName.value = "John"; // logs: ["John",""] ["", ""]
...
+
+```js
+const state = reactive([
+  { attributes: { name: "" }, id: 1 },
+  { attributes: { name: "" }, id: 2 },
+]);
+
+watch(
+  () => state.map((e) => e.attributes),
```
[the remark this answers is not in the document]
fimion: I've hopefully simplified both examples where they still make sense.

bencodezen (approved):
Thanks for your work on this @fimion and @NataliaTepluhina for the great reviews!

I think we're good to go for release so people can utilize this information. We can always iterate in another PR!

PR #735, opened 2020-12-05
watch guide cleanup

fimion:
## Description of Problem

While working on #734 I noticed some issues with examples and spelling

## Proposed Solution

I have added a more real example of using an array for `watch` and fixed a couple of syntax errors.


NataliaTepluhina (approved):
@fimion thank you! Looks great to me

PR #736, opened 2020-12-07
fix: Resolve typo in application-config example

potato4d:


bencodezen (approved):
Thanks for catching this!

PR #740, opened 2020-12-09
Revise component-basics.md

skirtles-code:
I've revised `component-basics.md`. In a few places there were descriptions of Vue 2 functionality that no longer apply with Vue 3. I've also tweaked grammar and wording in various places.

I've made inline comments for the most significant changes.

There are two things I've removed in multiple places:

1. The word *fortunately*. The word itself is fine but it seems to be introducing core features as though their existence is a lucky coincidence.
2. The expression *keep in mind*. There's an implication in that expression that a concept has been discussed previously. It can't really be used to introduce something new.

### DIFF
--- a/src/guide/component-basics.md
+++ b/src/guide/component-basics.md
@@ -97,7 +97,7 @@ app.component('blog-post', {
 app.mount('#blog-post-demo')
 ```
 
-When a value is passed to a prop attribute, it becomes a property on that component instance. The value of that property is accessible within the template, just like any other property of the component instance.
+When a value is passed to a prop attribute, it becomes a property on that component instance. The value of that property is accessible within the template, just like any other component property.
 
 A component can have as many props as you like and, by default, any value can be passed to any prop.
 
--- a/src/guide/component-basics.md
+++ b/src/guide/component-basics.md
@@ -40,7 +40,7 @@ app.mount('#components-demo')
 
 <common-codepen-snippet title="Component basics" slug="abORVEJ" tab="js,result" :preview="false" />
 
-Since components are reusable instances, they accept the same options as a root instance, such as `data`, `computed`, `watch`, `methods`, and lifecycle hooks. The only exceptions are a few root-specific options like `el`.
+Since components are reusable instances, they accept the same options as a root instance, such as `data`, `computed`, `watch`, `methods`, and lifecycle hooks.
 
 ## Reusing Components
 
@@ -66,7 +66,7 @@ It's common for an app to be organized into a tree of nested components:
 
 For example, you might have components for a header, sidebar, and content area, each typically containing other components for navigation links, blog posts, etc.
 
-To use these components in templates, they must be registered so that Vue knows about them. There are two types of component registration: **global** and **local**. So far, we've only registered components globally, using `component` method of created app:
+To use these components in templates, they must be registered so that Vue knows about them. There are two types of component registration: **global** and **local**. So far, we've only registered components globally, using the `component` method of our app:
 
 ```js
 const app = Vue.createApp({})
@@ -76,7 +76,7 @@ app.component('my-component-name', {
 })
 ```
 
-Globally registered components can be used in the template of `app` instance created afterwards - and even inside all subcomponents of that root instance's component tree.
+Globally registered components can be used in the template of any component within the app.
 
 That's all you need to know about registration for now, but once you've finished reading this page and feel comfortable with its content, we recommend coming back later to read the full guide on [Component Registration](component-registration.md).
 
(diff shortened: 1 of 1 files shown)


src/guide/component-basics.md
```diff
@@ -319,11 +321,7 @@ Now `v-model` should work perfectly with this component:
 <custom-input v-model="searchText"></custom-input>
 ```
 
-Another way of creating the `v-model` capability within a custom component is to use the ability of `computed` properties' to define a getter and setter.
-
-In the following example, we refactor the `custom-input` component using a computed property.
-
-Keep in mind, the `get` method should return the `modelValue` property, or whichever property is being using for binding, and the `set` method should fire off the corresponding `$emit` for that property.
+Another way of implementing `v-model` within this component is to use the ability of `computed` properties to define a getter and setter. The `get` method should return the `modelValue` property and the `set` method should emit the corresponding event:
```
skirtles-code, line 324: I've condensed 3 paragraphs down to 1 but I don't think anything important has been lost.

Previously the wording made reference to the possibility of binding other properties with `v-model`. That concept hasn't been introduced yet and I don't think there's much value in mentioning it here.

src/guide/component-basics.md
```diff
@@ -398,7 +396,7 @@ In the example above, `currentTabComponent` can contain either:
 
 See [this sandbox](https://codepen.io/team/Vue/pen/oNXaoKy) to experiment with the full code, or [this version](https://codepen.io/team/Vue/pen/oNXapXM) for an example binding to a component's options object, instead of its registered name.
 
-Keep in mind that this attribute can be used with regular HTML elements, however they will be treated as components, which means all attributes **will be bound as DOM attributes**. For some properties such as `value` to work as you would expect, you will need to bind them using the [`.prop` modifier](../api/directives.html#v-bind).
+You can also use the `is` attribute to create regular HTML elements.
```
skirtles-code, line 399: Most of this paragraph is no longer true in Vue 3, so I've removed it. The lone remaining sentence feels a bit out of place but it is an important point that I didn't want to lose.

src/guide/component-basics.md
```diff
@@ -76,15 +76,15 @@ app.component('my-component-name', {
...
+Globally registered components can be used in the template of any component within the app.
 
 That's all you need to know about registration for now, but once you've finished reading this page and feel comfortable with its content, we recommend coming back later to read the full guide on [Component Registration](component-registration.md).
 
 ## Passing Data to Child Components with Props
 
 Earlier, we mentioned creating a component for blog posts. The problem is, that component won't be useful unless you can pass data to it, such as the title and content of the specific post we want to display. That's where props come in.
 
-Props are custom attributes you can register on a component. When a value is passed to a prop attribute, it becomes a property on that component instance. To pass a title to our blog post component, we can include it in the list of props this component accepts, using a `props` option:
+Props are custom attributes you can register on a component. To pass a title to our blog post component, we can include it in the list of props this component accepts, using the `props` option:
```
skirtles-code, line 87: Despite appearances, I haven't removed the middle sentence entirely, I've just moved it to after the example. The 5 sentences that appeared directly before and after this example are all still present but I've moved them around and reworded one of them. I think they form more natural paragraphs this way.

src/guide/component-basics.md
```diff
@@ -40,7 +40,7 @@ app.mount('#components-demo')
 
 <common-codepen-snippet title="Component basics" slug="abORVEJ" tab="js,result" :preview="false" />
 
-Since components are reusable instances, they accept the same options as a root instance, such as `data`, `computed`, `watch`, `methods`, and lifecycle hooks. The only exceptions are a few root-specific options like `el`.
+Since components are reusable instances, they accept the same options as a root instance, such as `data`, `computed`, `watch`, `methods`, and lifecycle hooks.
```
skirtles-code, line 43: I don't believe the final sentence is true in Vue 3 so I've removed it.

src/guide/component-basics.md
```diff
@@ -97,7 +97,9 @@ app.component('blog-post', {
 app.mount('#blog-post-demo')
 ```
 
-A component can have as many props as you'd like and by default, any value can be passed to any prop. In the template above, you'll see that we can access this value on the component instance, just like with `data`.
+When a value is passed to a prop attribute, it becomes a property on that component instance. The value of that property is accessible within the template, just like any other property of the component instance.
```
NataliaTepluhina, line 100: **nitpick**: we have two repeating `component instance` in this paragraph. While it's technically correct, it reads a bit weird, maybe we could replace the secods case with `any other component property`?

NataliaTepluhina (commented):
@skirtles-code awesome refactoring! I've added one non-blocking nitpick but the rest looks good to me 👍🏻

NataliaTepluhina (approved):
Happy to approve this one! I'll give this one more day to allow other maintainers taking a look but if there are no objections in the upcoming day, I'm going to merge it

skirtles-code:
Thanks. Always grateful for the scrutiny. I'd be happy to wait until next week if you want to let the weekend contributors have their opportunity.

NataliaTepluhina:
@skirtles-code seems there are no objections so let's merge it 🎉

PR #747, opened 2020-12-13
Update events.md

shivams906:
The counter should start from 0 since the button has not been clicked yet.

### DIFF
--- a/src/guide/events.md
+++ b/src/guide/events.md
@@ -17,7 +17,7 @@ For example:
 Vue.createApp({
   data() {
     return {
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
@NataliaTepluhina The CodePen will also need updating. Could you also change the `v-on:click` to `@click`, so that it matches the code above? Thanks.

NataliaTepluhina (approved):
@shivams906 @skirtles-code thank you both! Pen updated accordingly

PR #750, opened 2020-12-15
Fix tip colons

FeixuRuins:
## Description of Problem
image

## Proposed Solution
Move `:::` to the next line


NataliaTepluhina (approved):
@FeixuRuins thank you for fixing this!

PR #754, opened 2020-12-18
Changed a recommendation for custom events name casing

### LINKED ISSUE #656: camelCase events can be listened to in the kebab-case
The documentation says it Won't work but it seems to work.

https://v3.vuejs.org/guide/component-custom-events.html#event-names

I checked in the following source code.

```html
<!DOCTYPE html>
<html>
  <head>
    <script src="https://unpkg.com/vue@3.0.2"></script>
  </head>
  <body>
    <div id="counter">
      {{ count }}
      <my-component @my-event="doSomething"></my-component>
    </div>
    <script>
      const App = {
        data() {
          return { count: 0 };
        },
        methods: {
          doSomething() {
            this.count++;
          },
        },
      };

      const app = Vue.createApp(App);

      app.component("MyComponent", {
        template: `<button @click="onClick">+1</button>`,
        emits: ["myEvent"],
        methods: {
          onClick() {
            this.$emit("myEvent");
          },
        },
      });

      app.mount("#counter");
    </script>
  </body>
</html>
```

Also, if I use a compiled component, I will get a warning if I define the emits option in the kebab-case.
https://github.com/vuejs/eslint-plugin-vue/issues/1347
The recommended casing for the event name is the kebab-case, but should we actually recommend the camelCase?

NataliaTepluhina:
## Description of Problem

Out current recommendation for custom events name casing are outdated (since `emits` can have an object syntax and in this case, camelCased event name is more natural)

## Proposed Solution

Change recommendation for events name casing to make it consistent with props

## Additional Information

Close #656

### DIFF
--- a/src/guide/component-custom-events.md
+++ b/src/guide/component-custom-events.md
@@ -14,7 +14,7 @@ this.$emit('myEvent')
 <my-component @my-event="doSomething"></my-component>
 ```
 
-Similarly to [props casing](/guide/component-props.html#prop-casing-camelcase-vs-kebab-case), we recommend using kebab-cased event listeners when you are using in-DOM templates. If you're using string templates, this limitation does not apply.
+As with [props casing](/guide/component-props.html#prop-casing-camelcase-vs-kebab-case), we recommend using kebab-cased event listeners when you are using in-DOM templates. If you're using string templates, this limitation does not apply.
 
 ## Defining Custom Events
 
--- a/src/guide/component-custom-events.md
+++ b/src/guide/component-custom-events.md
@@ -4,32 +4,17 @@
 
 ## Event Names
 
-Unlike components and props, event names don't provide any automatic case transformation. Instead, the name of an emitted event must exactly match the name used to listen to that event.
-
-```js
-this.$emit('my-event')
-```
-
-```html
-<my-component @my-event="doSomething"></my-component>
-```
-
-If we're emitting a camelCased event name:
+Like components and props, event names provide an automatic case transformation. If you emit an event from the child component in camel case, you will be able to add a kebab-cased listener in the parent:
 
 ```js
 this.$emit('myEvent')
 ```
 
-Listening to the kebab-cased version will have no effect:
-
 ```html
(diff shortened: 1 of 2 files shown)


src/guide/component-custom-events.md
```diff
@@ -4,32 +4,17 @@
...
-
 ```html
-<!-- Won't work -->
 <my-component @my-event="doSomething"></my-component>
 ```
 
-Since event names will never be used as variable or property names in JavaScript, there is no reason to use camelCase or PascalCase. Additionally, `v-on` event listeners inside DOM templates will be automatically transformed to lowercase (due to HTML's case-insensitivity), so `@myEvent` would become `@myevent` -- making `myEvent` impossible to listen to.
-
-For these reasons, we recommend you **always use kebab-case for event names**.
+Similarly to [props casing](/guide/component-props.html#prop-casing-camelcase-vs-kebab-case), we recommend using kebab-cased event listeners when you are using in-DOM templates. If you're using string templates, this limitation does not apply.
```
skirtles-code, line 17:
```suggestion
As with [props casing](/guide/component-props.html#prop-casing-camelcase-vs-kebab-case), we recommend using kebab-cased event listeners when you are using in-DOM templates. If you're using string templates, this limitation does not apply.
```

At first reading, the use of *'Similarly to'* derailed me a bit. To me it didn't sound right. My instinctive reaction was to change it to *'Similar to'*.

However, I spent some time researching this online and, with hindsight, I think *'Similarly to'* probably is technically correct. That said, plenty of native speakers on those same forums seemed to agree that it sounds wrong, even if it isn't. A recurring suggestion was to reword the sentence altogether, which is what I've attempted.

skirtles-code (approved):
Looks good.

I've made one suggestion but feel free to ignore it if you prefer the current wording.

I'm surprised there isn't an entry for this in the Style Guide.

NataliaTepluhina:
@skirtles-code thank you for the suggestion! I better trust a native speaker on this matter 😅 

I believe we should add the entry to styleguide, next to props section, I will add one there

PR #759, opened 2020-12-22
fix: error option in async component migration page

lovetingyuan:
## Description of Problem
Wrong option field in async component migration page: https://v3.vuejs.org/guide/migration/async-components.html#_3-x-syntax.

## Proposed Solution
change the wrong option name to correct ones.

## Additional Information
image


skirtles-code (approved):
Looks good.

I'm unclear why that example is needed. It doesn't really add anything that isn't included in the previous example. With a small tweak to the wording of the sentence about the `loader` option, I think we could remove the example altogether.

NataliaTepluhina (approved):

PR #768, opened 2020-12-30
chore: replace node-sass with sass

Xenonym:
## Description of Problem
`node-sass` is used to compile Sass styles. However, [it has been deprecated](https://sass-lang.com/blog/libsass-is-deprecated).

## Proposed Solution
Replace `node-sass` with `sass` as recommended:

> - We recommend all existing LibSass users make plans to eventually move onto Dart Sass, and that all Sass libraries make plans to eventually drop support for LibSass.
>
> [...]
>
> If you’re a user of Node Sass, migrating to Dart Sass is straightforward: just replace node-sass in your package.json file with sass. Both packages expose the same JavaScript API.

## Additional Information
- [`sass/node-sass` #2952: Mark Node Sass as deprecated](https://github.com/sass/node-sass/issues/2952)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "devDependencies": {
     "@vuepress/plugin-pwa": "^1.5.4",
-    "node-sass": "^4.13.1",
+    "sass": "^1.32.0",
     "sass-loader": "^8.0.2",
     "vuepress": "^1.5.4"
   },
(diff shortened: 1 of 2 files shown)


skirtles-code (approved):
I tested this out on my laptop and everything seemed to install/build/render just as it did previously.

This seems like a good change to me.

NataliaTepluhina (approved):

PR #769, opened 2020-12-30
Add `binding` attribute on application-api directive binding example

Darknessshade:
Update `application-api.md` directive binding argument exemple which contain only `el` exemple and no reference at `binding` attribute.

## Description of Problem

The example is under `binding` attribute explaination but it doesn't contain how to use `binding` attribute. This could been confusing.

## Proposed Solution

Add binding attribute on the example

### DIFF
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -132,7 +132,8 @@ An object containing the following properties.
 
 ```js
 app.directive('focus', {
-  mounted(el) {
+  mounted(el, binding) {
+    console.log(binding);
     el.focus()
   }
 })


skirtles-code:
Thank you for the PR.

I agree that this could be clearer but I don't think adding the `binding` is appropriate here. That example is supposed to be part of the explanation of `dir`, not `binding`. The example follows on from the text that precedes it and runs into the next example. The two examples need to match as they're trying to illustrate the value of `dir` property.

In my opinion, we don't really need the examples here. If the text describing `dir` explained that it is the directive's configuration object then I think that would be sufficient.

NataliaTepluhina:
Thank you for the PR! I agree with @skirtles-code here, I don't think this change is necessary to explain what `dir` is

PR #770, opened 2021-01-01
fix(api): remove arguments of `unmount` api

underfin:
## Description of Problem
See https://github.com/vuejs/vue-next/issues/2922
## Proposed Solution
Fixed here https://github.com/vuejs/vue-next/issues/2922

### DIFF
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -263,7 +263,7 @@ app.provide('user', 'administrator')
 
 - **Usage:**
 
-  Unmounts a root component of the application instance on the provided DOM element.
+  Unmounts a root component of the application instance.
 
 - **Example:**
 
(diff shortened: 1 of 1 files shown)


skirtles-code (changes requested):
I think the words:

>  on the provided DOM element

from the **Usage** section should also be removed.

src/api/application-api.md
```diff
@@ -261,10 +261,6 @@ app.provide('user', 'administrator')
 
 ## unmount
 
-- **Arguments:**
-
-  - `{Element | string} rootContainer`
-
 - **Usage:**
 
   Unmounts a root component of the application instance on the provided DOM element.
```
NataliaTepluhina, line 266:
```suggestion
  Unmounts a root component of the application instance.
```

PR #777, opened 2021-01-03
docs: Fix example hightlight in functional-components.md

Alex-Sokolov:
Correctly highlight changed lines in example with functional component


skirtles-code (approved):
The line highlighting change looks good to me.

The language is wrong though. Both SFCs should really be `vue` rather than `js`, otherwise the template doesn't get the correct syntax highlighting.

Alex-Sokolov:
I think highlighting here for changed lines comparing with previous example. And that's not only removing functional attribute but changing props to $props and attrs to $attrs

PR #779, opened 2021-01-04
feat: add note on isCustomElement in runtime-only builds

Xenonym:
## Description of Problem
The `isCustomElement` config option only applies when using the runtime compiler. Since most users will be using the runtime-only compiler (eg. via `vue-loader` or `@vue/cli`), they may attempt to configure this option by mistake (eg. [`vuejs/vue-next` #2779](https://github.com/vuejs/vue-next/issues/2779)).

## Proposed Solution
Let's add a note that they may need to configure `vue-loader`'s `compilerOptions` for custom elements to work properly.

## Additional Information
The language used is [taken from the migration guide](https://github.com/vuejs/docs-next/blob/master/src/guide/migration/global-api.md#L102).

### DIFF
--- a/src/api/application-config.md
+++ b/src/api/application-config.md
@@ -94,6 +94,10 @@ Specifies a method to recognize custom elements defined outside of Vue (e.g., us
 
 > Note that all native HTML and SVG tags don't need to be matched in this function - Vue parser performs this check automatically
 
+::: tip Important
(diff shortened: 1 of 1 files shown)


Xenonym:
Hmm, I am not quite sure if it's an issue with my PR or the CI job that tags alert blocks - is it because I am not a member of the @vuejs org?

NataliaTepluhina (approved):
@Xenonym thank you for adding this! The CI job fails to draw attention to the added `tip` block - we try to avoid having too many of them, but the current one seems to be justified for me

PR #792, opened 2021-01-08
Added example how to bind multiple attributes to the same slot

timkley:
To me, reading the docs at first, didn't make it very clear how to bind multiple attributes to one slot.

I've added an example how to do it.


NataliaTepluhina (approved):
Thank you for the addition!

PR #793, opened 2021-01-08
fix: prevent layout shift and appearance of horizontal scroll

aliataf:
## Description of Problem
The layout was shifting horizontally because some content passes their parents' width, so there was an annoying horizontal scrollbar showing in the documentation.

## Proposed Solution
Prevent the horizontal overflowing by giving the `overflow-x: hidden` style to the `html, body` selector which fixes the problem and prevents any content from being horizontally shifted.

### DIFF
--- a/src/.vuepress/theme/styles/index.styl
+++ b/src/.vuepress/theme/styles/index.styl
@@ -9,6 +9,7 @@ html, body
   padding 0
   margin 0
   background-color #fff
+  overflow-x hidden
 
 body
   font-family $fontPrimary


skirtles-code:
I've not seen this problem myself. Does it only happen on particular pages? Do I need to resize my browser window to a particular size to see it? Does it only happen in a specific browser?

Could you maybe include a screenshot?

aliataf:
@skirtles-code
It happens on the homepage no matter what the size of the browser window and I have tested it on Chrome 87.0.4280.88 and Firefox 84.0.2 as shown in the screenshot. 
vuedocs

skirtles-code:
Thanks, I can reproduce it now. For me it kicks in below 1376px or 1300px depending on the page.

There are some media queries for the ads that seem to be causing it. I'm still digging to understand it properly.

NataliaTepluhina:
Closing this in favor of https://github.com/vuejs/docs-next/pull/798

PR #794, opened 2021-01-09
fix: follow convention for kebab-case

bencodezen:
## Description of Problem

We recommend to use kebab case in the beginning of the docs. So the example should follow suit unless I'm missing something.


skirtles-code:
I think the recommendation was changed as of #754. It may be that it needs changing elsewhere.

bencodezen:
Good call. Thanks @skirtles-code!

PR #796, opened 2021-01-09
Update introduction.md

doingandlearning:
## Description of Problem
I was working through the examples in Vue 3 and got the following error:
```
Elements in iteration expect to have 'v-bind:key' directives  vue/require-v-for-key
```
## Proposed Solution
Adding the key removes this error. 

## Additional Information
I'm new to Vue (coming from React) and don't know if I'm missing something.

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -189,7 +189,7 @@ There are quite a few other directives, each with its own special functionality.
 ```html
 <div id="list-rendering">
   <ol>
-    <li v-for="todo in todos" :key="todo.text">
+    <li v-for="todo in todos" v-bind:key="todo.text">
       {{ todo.text }}
     </li>
   </ol>
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -189,7 +189,7 @@ There are quite a few other directives, each with its own special functionality.
 ```html
 <div id="list-rendering">
   <ol>
-    <li v-for="todo in todos">
+    <li v-for="todo in todos" :key="todo.text">
       {{ todo.text }}
     </li>
   </ol>


NataliaTepluhina:
@doingandlearning Thank you for creating the PR! Having the `key` on `v-for` is optional for Vue 3, let me check with ESLint plugin authors to make sure this rule is not in 'essentials' for v3

skirtles-code:
Thanks for the PR. You raise an interesting point. Personally I'm caught in two minds about whether to add the `key` or not.

Some observations:

1. Vue itself doesn't require a `key` on elements, only components. The error you mentioned is coming from the linter, which enforces some optional best practices.
2. Generally that page tries to avoid bringing up any difficult concepts that might need explaining. It's a blast through the functionality to set the scene for the pages that follow.
3. A later example does include a `key`. That example needs it because it's on a component, not an element. It also includes a comment, effectively telling the reader to ignore the `key` for now. If we do introduce a `key` to the earlier example then we'd need a similar note.
4. That page uses `v-bind:` in full rather than the `:` shorthand. The shorthands are only used after they're introduced on the **Template Syntax** page.

src/guide/introduction.md
```diff
@@ -189,7 +189,7 @@ There are quite a few other directives, each with its own special functionality.
 ```html
 <div id="list-rendering">
   <ol>
-    <li v-for="todo in todos">
+    <li v-for="todo in todos" :key="todo.text">
```
NataliaTepluhina, line 192:
```suggestion
    <li v-for="todo in todos" v-bind:key="todo.text">
```

NataliaTepluhina:
We discussed this a bit with Vue team. @skirtles-code brought a few great points, and they are correct: technically `key` is not required on elements, ESLint is too strict with this warning. However, after some discussion we came up with the point `it's safer to always add a key - overall it results in less mental overhead than reasoning about whether it's safe to not use a key` (for example, we don't use a key on the element and then we add an ``, and apparently now the key is required)

So we will return a section about `key` to this chapter. I've edited your changes and will add more of them in my upcoming PR

PR #800, opened 2021-01-10
Add type inference for emitted events

### LINKED ISSUE #776: Add about type inference of emits option into TS Documentation
[There is a section about props](https://v3.vuejs.org/guide/typescript-support.html#annotating-props) but there is no section about `emits` for now.

[active-rfcs/0030-emits-option](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0030-emits-option.md#type-inference)

NataliaTepluhina:
## Description of Problem

Currently, we don't have `emits` property described in the TS guide

## Proposed Solution

This PR adds a small example for `emits`

## Additional Information

Close #776

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -212,7 +212,7 @@ const Component = defineComponent({
       type: Object as PropType<Book>,
       // Make sure to use arrow functions
       default: () => ({
-        title: "Arrow Function Expression"
+        title: 'Arrow Function Expression'
       }),
       validator: (book: Book) => !!book.title
     },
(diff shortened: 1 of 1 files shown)


src/guide/typescript-support.md
```diff
@@ -232,6 +232,30 @@ const Component = defineComponent({
...
 ```
 
+## Annotating emits
+
+We can annotate a payload for the emitted event. Also, all non-declared emitted events will throw a type error when called:
+
+```ts
+const Component = defineComponent({
+  emits: {
+    addBook: (payload: { bookName: string }) => {
```
skirtles-code, line 242: **Nitpick**: Is there any reason to use an arrow function rather than a regular function in this example?

```suggestion
    addBook(payload: { bookName: string }) {
```

skirtles-code (approved):
TypeScript is a bit of a mystery to me, but to my untrained eye everything seems fine. I tried running the code and got the expected type errors.

I made a suggestion about one line but it's nothing important.

Issue #698, opened 2020-11-17
Clarify the cases using number, bigint or symbol as key


LongTengDao:
Vue 2 doc had add `symbol` for key api in version 2.5.12

In Vue 3 doc, key is still `number | string`.

1. Is `symbol` deprecated? Or just not implement yet?
2. Will `bigint` be supported?
3. Is `1` treated as same key to `"1"`? Whether is or not, is it intended?

skirtles-code:
I've opened an issue against the main repo to get clarification about what the expected behaviour is:

https://github.com/vuejs/vue-next/issues/2633

I believe the TS definitions currently only allow `number | string` but, in my opinion, that's overly restrictive. If the TS definitions are updated then the docs can follow.

LongTengDao:
I'm not sure that's a bug or feature, so I didn't ask in vue-next. I even have no idea it belong to Vue 2 / 3 / 2&amp;3...

Problem more than restrictive is that, if support all these types, they will be use as it is, or stringify any type (except symbol) to string firstly?

I can study the soure, but I'm not sure that's intended and stable.

PR #810, opened 2021-01-14
Added keys to examples on List Rendering

NataliaTepluhina:
As explained in https://github.com/vuejs/docs-next/pull/796, we decided to make keys recommended for all cases. To make this more visible, we add `key` to all the code examples after we explained what `key` is

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -189,7 +189,7 @@ There are quite a few other directives, each with its own special functionality.
 ```html
 <div id="list-rendering">
   <ol>
-    <li v-for="todo in todos" v-bind:key="todo.text">
+    <li v-for="todo in todos">
       {{ todo.text }}
     </li>
   </ol>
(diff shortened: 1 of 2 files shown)


src/guide/introduction.md
```diff
@@ -189,7 +189,7 @@ There are quite a few other directives, each with its own special functionality.
 ```html
 <div id="list-rendering">
   <ol>
-    <li v-for="todo in todos" v-bind:key="todo.text">
```
NataliaTepluhina, line 192: @skirtles-code after some thinking I agreed with your point and removed this attribute to prevent a cognitive overhad

skirtles-code (approved):
Looks good.

Longer term I'd like to change some of those examples to use the object directly as the `key`, rather than a property. That seems to work with Vue 3 but I don't believe it's been officially acknowledged as a supported feature, so I don't think we can use it in the docs yet. There's already #698 tracking this.

PR #818, opened 2021-01-18
Use `@vitejs/app` instead of `vite-app`

hanhan9449:
## Description of Problem
When I use `vite-app`, `Tailwind CSS` can't work.
## Proposed Solution
Use `@vitejs/app` instead of `vite-app`


NataliaTepluhina (approved):
Thank you for the fix!

PR #836, opened 2021-02-01
Fix wrong comment in basic-reactivity.md

JohnNiang:
## Description of Problem

Wrong comment `obj.value` as below:

```markdown
// it will update `obj.value`
count.value++
console.log(count.value) // 2
console.log(obj.count) // 2
```

## Proposed Solution

```diff
- // it will update `obj.value`
+ // it will update `obj.count`
```


NataliaTepluhina (approved):
@JohnNiang thank you for the fix!

PR #856, opened 2021-02-12
Enhance errorHandler docs

nobitagit:
## Description of Problem
A user (me!) can mistakenly expect `errorHandler` to be invoked by errors originating from instance methods. This is not the case.

The situation is well [described here](https://stackoverflow.com/a/49215338/1446845) and has interactive examples.

## Proposed Solution

Explicitly state that instance methods errors won't be handled by errorHandler.

## Additional Information
I also stated the extra `info` argument that is available to the consumer.

---

If this is all good and approved I can port this to the v2 docs as well.
Thanks for reviewing.

### DIFF
--- a/src/api/application-config.md
+++ b/src/api/application-config.md
@@ -26,7 +26,7 @@ app.config.errorHandler = (err, vm, info) => {
 }
 ```
 
-Assign a handler for uncaught errors during component render function and watchers. The handler gets called with the error and the application instance.
+Assign a handler for uncaught errors during component render function and watchers (but **not** those coming from instance methods). The handler gets called with the error, the application instance and Vue specific error info.
 
 > Error tracking services [Sentry](https://sentry.io/for/vue/) and [Bugsnag](https://docs.bugsnag.com/platforms/browsers/vue/) provide official integrations using this option.
 


skirtles-code (commented):
I think the situation may have changed slightly from Vue 2. I tried porting the example from the Stack Overflow answer across to Vue 3 and the error handler was triggered.

I believe that Vue 3 will trigger the error handler for errors in any top-level process that it controls. That includes rendering, computed properties, watchers, lifecycle hooks, `setup` and event listeners. Uncaught errors throw by any of those processes should trigger the error handler.

In some cases Vue will also handle rejected promises returned by those processes. So, for example, if a `mounted` hook returns a promise and that promise is subsequently rejected then that will also trigger the error handler. While it's not particularly common for someone to manually return a promise like this it does frequently happen implicitly as part of using `async`/`await`.

It is true that `methods` don't directly trigger the error handler. Vue has no way of knowing whether a method is supposed to throw an error. Such an error might be caught be whatever called the method, in which case it wouldn't be appropriate to trigger the error handler. If the error isn't caught then it should propagate up to the relevant top-level process, which will trigger the error handler.

As far as I can tell the only problem case is using a method as part of an asynchronous process. e.g. Calling it from a timer, or as a callback for an XHR request. In those cases, Vue won't know about the error so it can't call the error handler.

skirtles-code:
@nobitagit Did you have any further thoughts on how best to document this in light of my explanation above?

I'd be reluctant to claim that the `errorHandler` isn't triggered by instance methods. That could be misconstrued as meaning that errors in methods will definitely not trigger the `errorHandler`, which is not true. Such errors would propagate up and in most cases they would trigger the `errorHandler` in the caller.

nobitagit:
Hi Skirtle, sorry for not following up sooner and many thanks for reviewing so thoroughly.

In light of your explanation I would say that the documentation for v3 is indeed fine as is. I think this is still valid for v2 but this still means that we can close this PR as it would be misleading. Thanks again for the explanation and following up!

PR #858, opened 2021-02-12
fix instead tab to 2 spaces

harapeko:
## Description of Problem
Tabs were used in the sample.

## Proposed Solution
I've unified them into a 2 spaces.

## Additional Information
https://v3.vuejs.org/guide/composition-api-setup.html#props:~:text=function%3A-,%2F%2F%20MyBook.vue,%7D


NataliaTepluhina (approved):
@harapeko thank you for the fix

PR #914, opened 2021-02-25
Use local copies of images rather than external URLs

skirtles-code:
A handful of pages were using external URLs to pull in images. This PR adds those images to this repo and uses local URLs instead.

The motivation here is:

1. Keep as much content as possible under source control, so it can't change unexpectedly.
2. Avoid dependencies on other domains that could be blocked or otherwise unavailable for some users.

I have not migrated the 'coffee cup' animation.


NataliaTepluhina (approved):
Thank you @skirtles-code!

PR #916, opened 2021-02-27
Added `propsData` to migration guide

NataliaTepluhina:
## Description of Problem

Currently, we don't reflect in the Migration guide that `propsData` option was removed. This PR aims to fix this

Close https://github.com/vuejs/docs-next/issues/848

### DIFF
--- a/src/guide/migration/props-data.md
+++ b/src/guide/migration/props-data.md
@@ -1,9 +1,9 @@
 ---
 badges:
-  - breaking
+  - removed
 ---
 
-# `propsData` passed on instance creation <MigrationBadges :badges="$frontmatter.badges" />
+# `propsData` <MigrationBadges :badges="$frontmatter.badges" />
 
 ## Overview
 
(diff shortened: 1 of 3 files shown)


src/guide/migration/props-data.md
```diff
@@ -0,0 +1,47 @@
...
+  { username: 'Evan' }
+)
+```
+
+```html
+<div id="app">
+  <!-- Will display 'Evan' -->
+  {{ username }}
+</div>
+```
```
skirtles-code, line 47:
````suggestion
```js
const app = createApp(
  {
    props: ['username'],
    template: '<div>{{ username }}</div>'
  },
  { username: 'Evan' }
)
```
````

The diff looks complicated but all I've done is move the template to the `template` option, same as the Vue 2 example.

skirtles-code (changes requested):
A few small tweaks from me.

NataliaTepluhina:
@skirtles-code fixes applied! 👍🏻

skirtles-code (approved):

PR #923, opened 2021-03-02
Add self hosting Vue as an option in the installation guide

### LINKED ISSUE #753: Installation by direct script include
The vue 2.0 installation documentation features a section for installing by direct script include, and offers downloads of the script file.

In the new vue documentation this option is omitted.  I am assuming this is an oversight, or perhaps an evaluation that people like me who eschew bundling and CDNs are an edge case best left to their own devices.

I am new to Vue, migrating from Knockout.  I only really care about data binding.  I have no current need for or use of webpack or any other bundlers, and (by choice) I use a very strict Content Security Policy that excludes all CDNs.  I am capable of finding and downloading and keeping a local copy to serve, but I'm unclear on whether the message is that this is ill-advised or unsupported (there be monsters there!) as opposed to just being "old-fashioned" and not publicized.  It seems like a very low-friction low-barrier-to-entry scenario that, in my opinion, should not be dismissed.

I guess this is more of a question than an issue.  

If it is an omission, then the issue is "please include documentation and downloads for direct script include installations."

skirtles-code:
Closes #753.

This PR adds a fourth option to the installation guide: downloading the JS file from the CDN and hosting it yourself.

### DIFF
--- a/src/guide/installation.md
+++ b/src/guide/installation.md
@@ -2,11 +2,12 @@
 
 Vue.js is built by design to be incrementally adoptable. This means that it can be integrated into a project multiple ways depending on the requirements.
 
-There are three primary ways of adding Vue.js to a project:
+There are four primary ways of adding Vue.js to a project:
 
 1. Import it as a [CDN package](#cdn) on the page
-2. Install it using [npm](#npm)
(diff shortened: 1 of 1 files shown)


bencodezen (approved):
Thanks for adding this! Looks great!

Issue #524, opened 2020-09-20
Some corrections seem to be needed in Reactivity in Depth section


t0yohei:
When I tried translating to Japanese, I noticed some points need to be corrected in the part of `How Vue Tracks These Changes` in Reactivity in Depth section.
https://v3.vuejs.org/guide/reactivity.html#how-vue-tracks-these-changes

> We mentioned before that in order to have an API that updates a final value when something changes, we’re going to have to set new values when something changes. 

In the part above, I think we have to say about function trucking, however, it seems to be said about updating new value.

> Track the function that changes it: We do this in a getter within the proxy, called `effect`

I think the function name is not `effect` but `track`.

skirtles-code:
I'm looking into updating this now. I think there are a few other tweaks required to that section too, so it may take a few days to get this ready for PR.

Brisstor:
@skirtles-code Since you're working on it, found one more typo here. I'm not sure if you have seen it before, so here it is:

> We mentioned before that in order to have an API that updates a final value when something changes, we’re going to have to set new values when something changes. We do this in the handler, in a function called `track`, where we pass in the `target` and `key`.

Suppose there should be `prop` instead of `key`.

skirtles-code:
@Brisstor Thanks.

I'm going to change all the references to `prop` and `key` to something else because both of those terms have other meanings in a Vue context and I'd rather avoid any potential for confusion.

Despite the apparent lack of progress I am still working on this. There are a few kinks that are proving difficult to iron out.

Brisstor:
@skirtles-code Thanks, also I'll be grateful if you mention me after you finish this issue so I would be able to update Russian translation of that section.

skirtles-code:
@t0yohei @Brisstor. I now have a draft of my proposed changes in #934.

I understand that making large changes like these places a burden on the translators. Hopefully the improvements are worth it. Please feel free to review what I've done and provide any feedback you may have on the PR. Thanks.

PR #925, opened 2021-03-04
Fix typo, track function name should be 'track' instead of 'effect'

tianzhou:


skirtles-code:
Thanks for the fix.

This is related to #524. The reason I hadn't made the change proposed here is because there's some subtlety to this and it's arguable as to whether `track` or `effect` is really what's intended. However, given it's taking me some time to fix this properly I think it is better for it to say `track` for now.

PR #928, opened 2021-03-05
Update plugins.md

guusvandewal:
## Description of Problem
I followed the plugin documentation but I didn't get any value in my {{ $translate('hi) }} string.
## Proposed Solution
I know it's just an example and putting the greetings object in that function is not the best solution but it should work if you follow along.
## Additional Information
This could probably have a better solution by using a different function to return the value

### DIFF
--- a/src/guide/plugins.md
+++ b/src/guide/plugins.md
@@ -66,7 +66,7 @@ export default {
   install: (app, options) => {
     app.config.globalProperties.$translate = key => {
       return key.split('.').reduce((o, i) => {
-        if (o) return o[i]
+        if (o) return o.greetings[i]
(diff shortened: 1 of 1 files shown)


skirtles-code:
Based on line 51, I think the intention is that you would call it as:

```html
{{ $translate('greetings.hello') }}
```

guusvandewal:
> Thanks for the PR.
> 
> Based on line 51, I think the intention is that you would call it as:
> 
> ```
> {{ $translate('greetings.hello') }}
> ```
> 
> As far as I can tell this is working correctly.

Yes, that is correct. Bit silly of me. Plz close the PR

PR #933, opened 2021-03-07
chore(template-syntax): fix heading level

kangetsu121:
Changed some heading levels in `template-syntax.md` because the heading level of `Caveats` is 3 and it looks like a child of `Shorthands`  whose heading level is 2 in the side navigation menu, though there doesn't seem to be a parent-child relationship.
If the heading level of `Caveats` is 3 intentionally for it is a kind of subtopic, please close this PR, thank you.


bencodezen:
Thanks for the PR @kangetsu121! I could see how it might be considered a caveat to the overall section, but at this time, the content is intended to be caveats to the shorthand section only. Appreciate it!

Issue #849, opened 2021-02-07
Common Problems using Vue 3 Reactivity


skirtles-code:
This is an umbrella issue, intended to gather examples of common problems, mistakes and misunderstandings when using the proxy-based reactivity system. The objective is to gather source material that might be used to add a new page to the documentation, or enhance the existing pages.

The idea is that it will be something similar to the Reactivity Caveats for Vue 2. However, I'm keen for the scope to be quite broad at this stage and we can decide what is actually worth documenting later. Anything related to reactivity that might catch out a beginner is in scope for now, even if it is already documented.

While almost all reactivity problems can be reduced down to 'you need to go through the proxy', I think there are plenty of cases where it isn't immediately obvious how that applies, especially for newcomers. My hope is that documenting some of these edge cases will save people from having to learn about them the hard way, while also giving them a deeper understanding of how reactivity works.

I'll add some examples of my own to this over the next few days.

skirtles-code:
Full example: https://jsfiddle.net/skirtle/wcdfe1ub/

Key section:

```js
async created () {
  const list = await loadList()
    
  this.items = list
    
  const list2 = await loadList2()
    
  list.push(...list2)
}
```

This uses the options API with no explicit mention of `reactive` or `ref`. The final line needs to be `this.items.push(...list2)` to go through the proxy. It is not immediately obvious that `list` and `this.items` are not equivalent after the line `this.items = list`. This is contrary to how things work in 'normal' JavaScript.

In Vue 2 this would have worked fine.

A similar example inside a Vuex action would be even more fiddly, with `commit` and `state` complicating things further. A user may attempt to put an extra call to `commit` at the end but that won't actually help.

skirtles-code:
There's a common question that gets asked: What is the difference between `reactive` and `ref` and when should I use one instead of the other?

Usually the question is in the context of objects rather than primitive values.

One way of explaining it is with analogy to `const` and `let`.

It's also potentially worth explaining that `ref` uses `reactive` on its value, so `ref` is just an extra wrapper around `reactive` to allow for reassignment.

skirtles-code:
Reassigning a local variable does nothing from a reactivity perspective:

```js
setup () {
  let myValue = ref(false)

  const update () {
    myValue = ref(true)
    // or even:
    // myValue = true
  }

  return {
    myValue,
    update
  }
}
```

While the example above uses `ref`, there are similar examples using `reactive` and or even a plain value.

[...]

PR #934, opened 2021-03-08
Rework the Reactivity in Depth guide

### LINKED ISSUE #524: Some corrections seem to be needed in Reactivity in Depth section

### LINKED ISSUE #787: Proxy vs. original identity – add/explain behavior in ref and reactive and of pass-by(copy/value) values
the section [proxy vs. original identity](https://v3.vuejs.org/guide/reactivity.html#proxy-vs-original-identity) states that a proxy is not identical to the original object it wraps. 

I assume `proxy` means the use of `reactive(myObject)` or `ref(myValue)`. 

Indeed,   
 
    let data = {myObject:{myNumber:3}}
    let reactiveData = reactive(newData);
    data === reactiveData; // → false

And it also applies to nested objects

    data.myObject === reactiveData.myObject // → false

However, what the guide does not mention is that this does not apply to the properties as long as they are not objects themselves
    
    data.myObject.myNumber === reactiveData.myObject.myNumber // → true

This also applies to the example which gives the caveat for the strong-comparison array methods like `.find`

    let myArray = [1,2,3,4]
    let myReactiveArray = reactive(myArray);
     myReactiveArray.find(number => number === 2) === myArray[1] //→ true

My suggestions would be: 

* Illustrate that `proxy` in vue can mean using the implicitly proxied values in the method’s API or ref/reactive in the composition API.
* Tell, that while proxied objects are not equal their original, pass-by-value types (number,string) are.

skirtles-code:
## Description of Problem

Rendered: https://deploy-preview-934--vue-docs-next-preview.netlify.app/guide/reactivity.html

This started as an attempt to make some small tweaks to the Reactivity in Depth page. One tweak led to another and the end result is a significant reworking of the existing page. Much of the original structure remains but the diff will likely be a mess and I suggest that any reviewers approach it is as though it were a new page.

The biggest addition is a section about effects. The term 'effect' is used by subsequent pages in the guide (e.g. when discussing effect flush timing) but it wasn't really explained anywhere. It also appears in the animation at the bottom of this page. To my mind, this is a really important concept that forms a key part of the 'how reactivity works' puzzle. Once the concept is introduced I have then weaved it in where appropriate for the explanations that follow.

I have intentionally avoided using the term 'watcher'. Roughly speaking, a watcher is a combination of an effect and a separate callback function. I didn't want to introduce yet another concept on this page if I could help it and, hopefully, the explanations hold together without explicitly mentioning it.

Some of the later examples on the existing page use a function called `reactive`. However, it wasn't explicitly stated what that function was or how it related to the proxies shown earlier. While `reactive` and related functions are covered more thoroughly on the next page of the guide, I've tried to tie everything together a little more explicitly, before it gets used by those later examples.

The original proxy examples are mostly unchanged. I've renamed the parameters to avoid the names `prop` and `key`, as those terms have other connotations in Vue that could lead to confusion. I have also merged the first couple of proxy examples. Those two examples only differed by a single line of console logging, so a single example seemed sufficient.

There are various other changes throughout but I hope they are mostly self-explanatory. Please feel free to to ask for clarification about any of the changes I've made.

## Additional Information

* Closes #524
* Closes #787

### DIFF
--- a/src/guide/reactivity.md
+++ b/src/guide/reactivity.md
@@ -13,41 +13,98 @@ This term comes up in programming quite a bit these days, but what do people mea
   Your browser does not support the video tag.
 </video>
 
-If you put the number two in the first cell, and the number 3 in the second and asked for the SUM, the spreadsheet would give it to you. No surprises there. But if you update that first number, the SUM automagically updates too.
+If you put the number 2 in the first cell, and the number 3 in the second and asked for the SUM, the spreadsheet would give it to you. No surprises there. But if you update that first number, the SUM automagically updates too.
 
-JavaScript doesn’t usually work like this -- If we were to write something comparable in JavaScript:
+JavaScript doesn’t usually work like this. If we were to write something comparable in JavaScript:
 
 ```js
-var val1 = 2
-var val2 = 3
-var sum = val1 + val2
+let val1 = 2
+let val2 = 3
+let sum = val1 + val2
 
-// sum
-// 5
+console.log(sum) // 5
 
 val1 = 3
 
-// sum
-// 5
+console.log(sum) // Still 5
 ```
 
 If we update the first value, the sum is not adjusted.
 
 So how would we do this in JavaScript?
 
-- Detect when there’s a change in one of the values
-- Track the function that changes it
-- Trigger the function so it can update the final value
+As a high-level overview, there are a few things we need to be able to do:
+
+1. **Track when a value is read.** e.g. `val1 + val2` reads both `val1` and `val2`.
+2. **Detect when a value changes.** e.g. When we assign `val1 = 3`.
+3. **Re-run the code that read the value originally.** e.g. Run `sum = val1 + val2` again to update the value of `sum`.
+
+We can't do this directly using the code from the previous example but we'll come back to this example later to see how to adapt it to be compatible with Vue's reactivity system.
+
+First, let's dig a bit deeper into how Vue implements the core reactivity requirements outlined above.
+
+## How Vue Knows What Code Is Running
+
+To be able to run our sum whenever the values change, the first thing we need to do is wrap it in a function:
+
+```js
+const updateSum = () => {
+  sum = val1 + val2
+}
+```
+
+But how do we tell Vue about this function?
+
+Vue keeps track of which function is currently running by using an *effect*. An effect is a wrapper around the function that initiates tracking just before the function is called. Vue knows which effect is running at any given point and can run it again when required.
+
+To understand that better, let's try to implement something similar ourselves, without Vue, to see how it might work.
+
+What we need is something that can wrap our sum, like this:
+
+```js
+createEffect(() => {
+  sum = val1 + val2
+})
+```
+
+We need `createEffect` to keep track of when the sum is running. We might implement it something like this:
+
+```js
+// Maintain a stack of running effects
+const runningEffects = []
+
+const createEffect = fn => {
+  // Wrap the passed fn in an effect function
+  const effect = () => {
+    runningEffects.push(effect)
+    fn()
+    runningEffects.pop()
+  }
+
+  // Automatically run the effect immediately
+  effect()
+}
+```
+
+When our effect is called it pushes itself onto the `runningEffects` array, before calling `fn`. Anything that needs to know which effect is currently running can check that array.
+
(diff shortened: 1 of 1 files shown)


src/guide/reactivity.md
```diff
@@ -13,72 +13,110 @@ This term comes up in programming quite a bit these days, but what do people mea
...
 
-Ok, so far, we’re just wrapping that object and returning it. Cool, but not that useful yet. But watch this, we can also intercept this object while we wrap it in the Proxy. This interception is called a trap.
+But knowing what code is running is just one part of the puzzle. How does Vue know what values the effect uses and how does it know when they change?
+
+## How Vue Tracks These Changes
+
+We can't track reassignments of local variables like those in our earlier examples, there's just no mechanism for doing that in JavaScript. What we can track are changes to object properties.
+
+When we return a plain JavaScript object from a component's `data` function, Vue will wrap that object in a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) with handlers for `get` and `set`. This is an ES6-only feature, but we offer a version of Vue 3 that uses the older `Object.defineProperty` to support IE 11 (this is still in development at the time of writing). Both have the same surface API, but the Proxy version is slimmer and offers improved performance.
```
NataliaTepluhina, line 101: **note**: at the latest Vue contributor days Evan mentioned that there _possibly_ won't be a IE-compatible Vue 3 build and we would recommend those who need IE support to stick with Vue 2 LTS. I propose to remove a mention of IE11 + Vue 3 here to prevent overpromising. We can always add it back later if the compatibility build is announced

src/guide/reactivity.md
```diff
@@ -194,29 +276,36 @@ const wrapped = new Proxy(obj, handlers)
...
+})
+
+console.log(obj.count === 0) // true
+```
 
-When you pass an object to a component instance as data, Vue converts it to a proxy. This proxy enables Vue to perform dependency-tracking and change-notification when properties are accessed or modified. Each property is considered a dependency.
+## How Rendering Reacts to Changes
 
-After the first render, a component would have tracked a list of dependencies &mdash; the properties it accessed during the render. Conversely, the component becomes a subscriber to each of these properties. When a proxy intercepts a set operation, the property will notify all of its subscribed components to re-render.
+The template for a component is compiled down into a [`render`](/guide/render-function.html) function. The `render` function creates the VNodes that describe how the component should be rendered. It is wrapped in an effect, allowing Vue to track the properties that are 'touched' while it is running.
```
NataliaTepluhina, line 303: **suggestion**: should we explain what VNode is here?

NataliaTepluhina (commented):
@skirtles-code great improvements! I've left a few non-blocking points for your consideration

stafyniaksacha:
Is there a way to speak about `ref` and `reactive` functions here ?

skirtles-code:
@stafyniaksacha Those functions are covered more thoroughly on the next page of the guide. This page is trying to explain the underlying mechanics of how the reactivity system works rather than the specific details of how the individual functions behave.

Arguably some of the material on this page, e.g. the sections **Proxied Objects** and **Proxy vs. original identity**, stray a little from the central topic and should be moved elsewhere. Those sections already existed prior to my reworking so I've retained them for now but it may make sense to move them eventually, especially once we introduce a proper page of Vue 3 reactivity caveats (#849).

src/guide/reactivity.md
skirtles-code: I've now reworked that paragraph to remove any mention of IE 11.

NataliaTepluhina: Thank you! 👍🏻

NataliaTepluhina (approved):

bencodezen (approved):
Solid improvements @skirtles-code! Going to go ahead and merge this because we can always iterate with another PR. Thanks for your work on this!

PR #935, opened 2021-03-09
Added watching a nested property to API reference

NataliaTepluhina:
## Description of Problem

We removed watching a nested property with a keypath from API. I returned it back :)


skirtles-code (approved):
I think there's a missing comma in the code but aside from that it looks good to me.

PR #936, opened 2021-03-09
Mention application mounting changes in migration guide and API

NataliaTepluhina:
Close https://github.com/vuejs/docs-next/issues/924

### DIFF
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -191,7 +191,7 @@ Apart from `el`, you should treat these arguments as read-only and never modify 
 
 - **Usage:**
 
-  Mounts a root component of the application instance on the provided DOM element.
+  Mounts a root component of the application instance on the provided DOM element. The `innerHTML` of the DOM element will be replaced with the rendered application.
 
 - **Example:**
 
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -191,7 +191,7 @@ Apart from `el`, you should treat these arguments as read-only and never modify 
 
 - **Usage:**
 
-  Mounts a root component of the application instance on the provided DOM element. If application has a `template` option, the `innerHTML` of the DOM element will be replaced with the rendered template.
+  The `innerHTML` of the provided DOM element will be replaced with the rendered template of the application root component.
 
 - **Example:**
 
--- a/src/api/application-api.md
+++ b/src/api/application-api.md
@@ -191,7 +191,7 @@ Apart from `el`, you should treat these arguments as read-only and never modify 
 
 - **Usage:**
 
-  Mounts a root component of the application instance on the provided DOM element. The `innerHTML` of the DOM element will be replaced with the rendered application.
+  Mounts a root component of the application instance on the provided DOM element. If application has a `template` option, the `innerHTML` of the DOM element will be replaced with the rendered template.
 
 - **Example:**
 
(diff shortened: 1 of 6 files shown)


src/api/application-api.md
```diff
@@ -191,7 +191,7 @@ Apart from `el`, you should treat these arguments as read-only and never modify
 
 - **Usage:**
 
-  Mounts a root component of the application instance on the provided DOM element.
+  Mounts a root component of the application instance on the provided DOM element. If application has a `template` option, the `innerHTML` of the DOM element will be replaced with the rendered template.
```
LinusBorg, line 194:
```suggestion
  Mounts the root component of the application instance on the provided DOM element. If that root component has a `template` option, the `innerHTML` of the provided DOM element will be replaced with the rendered template.
```

The above seems better, but more generally, I think the "if ..." could cause some miunderstandings:

- If the root component is an SFC, it won't have a template - still the innerHTML will be replaced
- This seems to indicate that I can mount an app without a template. Event though that might technically be possible, it's not very sensical.

We should probably just write

"The `innerHTML` of the provided DOM element will be replaced with the rendered template of the root component." and remove the if clause.

src/guide/migration/introduction.md
```diff
@@ -104,13 +104,13 @@ The following consists a list of breaking changes from 2.x:
 - [`<TransitionGroup>` now renders no wrapper element by default](/guide/migration/transition-group.html)
 - [When watching an array, the callback will only trigger when the array is replaced. If you need to trigger on mutation, the `deep` option must be specified.](/guide/migration/watch.html)
 - `<template>` tags with no special directives (`v-if/else-if/else`, `v-for`, or `v-slot`) are now treated as plain elements and will result in a native `<template>` element instead of rendering its inner content.
-- In Vue 2.x, application root container's `outerHTML` is replaced with root component template (or eventually compiled to a template, if root component has no template/render option). Vue 3.x now uses application container's `innerHTML` instead - this means the container itself is no longer considered part of the template.
+- [Mounted application does not replace an element it's mounted to](/guide/migration/mount-changes.html)
```
LinusBorg, line 107:
```suggestion
- [Mounted application does not replace the element it's mounted to](/guide/migration/mount-changes.html)
```

src/guide/migration/mount-changes.md
```diff
@@ -0,0 +1,94 @@
...
+title: 'Mount API changes'
+badges:
+  - breaking
+---
+
+# Mounted application does not replace the element <MigrationBadges :badges="$frontmatter.badges" />
+
+## Overview
+
+In Vue 2.x, when mounting an application that has a `template`, the rendered content replaces the element we mount to. In Vue 3.x, rendered application is appended as a child of such an element, replacing element's `innerHTML`.
```
LinusBorg, line 11:
```suggestion
In Vue 2.x, when mounting an application that has a `template`, the rendered content replaces the element we mount to. In Vue 3.x, the rendered application is appended as a child of such an element, replacing element's `innerHTML`.
```

src/guide/migration/mount-changes.md
```diff
@@ -0,0 +1,94 @@
...
+  },
+  template: `
+    <div id="rendered">{{ message }}</div
+  `
+})
+
+app.$mount('#app')
+```
+
+When we mount an application to the page that has a `div` with the passed selector (in our case, it's `id="app"`):
```
LinusBorg, line 45:
```suggestion
When we mount this application to the page that has a `div` with the passed selector (in our case, it's `id="app"`):
```

src/guide/migration/mount-changes.md
```diff
@@ -0,0 +1,94 @@
...
+  },
+  template: `
+    <div id="rendered">{{ message }}</div>
+  `
+})
+
+app.mount('#app')
+```
+
+A code above, when app is mounted to the page that has a `div` with `id="app"`, will result in:
```
LinusBorg, line 82:
```suggestion
When this app is mounted to the page that has a `div` with `id="app"`, this will result in:
```

LinusBorg (approved):
Awesome work, just found some grammar nitpicks here and there, see comments!

src/api/application-api.md
NataliaTepluhina: I agree with the latter 👍🏻  Changed the description to remove if/else

PR #937, opened 2021-03-10
Ssr webpack

### LINKED ISSUE #533: SSR support
Hey!

Kind of hard to to find any information on that, so I was wondering if there is any roadmap/schedule for Vue 3 SSR support. The current server-renderer only supports version 2. . I peeked into it, to see if I can somehow tweek it to make it work with Vue 3, but it raises too many issues. I'm using the composition API already, but would really love to migrate one of my SSR apps to Vue 3.

Thanks for clarification!

### LINKED ISSUE #726: cookbook idea: Server-side rendering
Hi. I'd like a cookbook guide to show SSR in a vue 3 application. I know it's probably going to look a lot like vue 2, but still I think there are some differences that would be nice to be shown.

NataliaTepluhina:
## Description of Problem

This is a draft for SSR documentation with a build step

Deployed demo here: https://deploy-preview-937--vue-docs-next-preview.netlify.app/guide/ssr/introduction.html

Close #533 
Close #726


src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,83 @@
+# Build Configuration
```
NataliaTepluhina, line 1: **note**: I still don't like how this file looks like, would be happy to have any improvements suggested

NataliaTepluhina:
@skirtles-code @danielroe would you mind checking this draft? It's very raw so far, any suggestions/improvements are much appreciated! 🙇🏻‍♀️

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
+# Source Code Structure
+
+## Avoid Stateful Singletons
+
+When writing client-only code, we are used to the fact that our code will be evaluated in a fresh context every time. However, a Node.js server is a long-running process. When our code is required into the process, it will be evaluated once and stays in memory. This means if you create a singleton object, it will be shared between every incoming request.
```
danielroe, line 5:
```suggestion
When writing client-only code, we can assume that our code will be evaluated in a fresh context every time. However, a Node.js server is a long-running process. When our code is first imported by the process, it will be evaluated once and then stay in memory. This means that if you create a singleton object, it will be shared between every incoming request, with the risk of cross-request state pollution.
```

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
...
+
+As seen in the basic example, we are **creating a new root Vue instance for each request.** This is similar to how each user will be using a fresh instance of the app in their own browser. If we use a shared instance across multiple requests, it will easily lead to cross-request state pollution.
+
+So, instead of directly creating an app instance, we should expose a factory function that can be repeatedly executed to create fresh app instances for each request:
+
+```js
+// app.js
+const { createSSRApp } = require('vue')
+
+module.exports = function createApp() {
```
danielroe, line 15:
```suggestion
import { createSSRApp } from 'vue'

export default function createApp() {
```

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
+# Source Code Structure
+
+## Avoid Stateful Singletons
+
+When writing client-only code, we are used to the fact that our code will be evaluated in a fresh context every time. However, a Node.js server is a long-running process. When our code is required into the process, it will be evaluated once and stays in memory. This means if you create a singleton object, it will be shared between every incoming request.
+
+As seen in the basic example, we are **creating a new root Vue instance for each request.** This is similar to how each user will be using a fresh instance of the app in their own browser. If we use a shared instance across multiple requests, it will easily lead to cross-request state pollution.
+
+So, instead of directly creating an app instance, we should expose a factory function that can be repeatedly executed to create fresh app instances for each request:
```
danielroe, line 9:
```suggestion
Therefore, we need to **create a new root Vue instance for each request.** In order to do that, we need to write a factory function that can be repeatedly executed to create fresh app instances for each request:
```

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
...
+      <div id="app">${appContent}</div>
+    </body>
+  </html>
+  `
+
+  res.end(html)
+})
+```
+
+The same rule applies to router and store instances as well. Instead of exporting it directly from a module and importing it across your app, you need to create a fresh instance in `createApp` and inject it from the root Vue instance.
```
danielroe, line 50:
```suggestion
The same rule applies to other instances as well (such as the router or store). Instead of exporting the router or store directly from a module and importing it across your app, you should create a fresh instance in `createApp` and inject it from the root Vue instance.
```

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Introducing a Build Step
+
+So far, we haven't discussed how to deliver the same Vue app to the client yet. To do that, we need to use webpack to bundle our Vue app. In fact, we probably want to use webpack to bundle the Vue app on the server as well, because:
+
+- Typical Vue apps are built with webpack and `vue-loader`, and many webpack-specific features such as importing files via `file-loader`, importing CSS via `css-loader` would not work directly in Node.js.
+
+- Although the latest version of Node.js fully supports ES2015 features, we still need to transpile client-side code to cater to older browsers. This again involves a build step.
+
+So the basic idea is we will be using webpack to bundle our app for both client and server - the server bundle will be required by the server and used for SSR, while the client bundle is sent to the browser to hydrate the static markup.
```
danielroe, line 60:
```suggestion
So far, we haven't discussed how to deliver the same Vue app to the client yet. To do that, we need to use webpack to bundle our Vue app.

* We need to process the server code with webpack. For example, `.vue` files need to be processed with `vue-loader`, and many webpack-specific features such as importing files via `file-loader` or importing CSS via `css-loader` do not work directly in Node.js.

* Similarly, we need a separate client-side build because although the latest version of Node.js fully supports ES2015 features, older browsers will require the code to be transpiled.

So the basic idea is that we will use webpack to bundle our app for both client and server. The server bundle will be required on the server and used to render static HTML, while the client bundle will be sent to the browser to hydrate the static markup.
```

danielroe (commented):
This is looking great! I've made a few comments for your consideration.

There is also some legacy ES5 syntax in getting started that was untouched by this PR - it may be worth just making sure everything is consistent across the examples.

skirtles-code (changes requested):
I haven't tried running the code examples myself yet but I've had a read through and made some suggestions.

I feel that Vue CLI could do more to facilitate this. Relying on a big chunk of webpack configuration in the SSR docs doesn't seem very maintainable. But we are where we are for now...

I should really spend some time learning webpack properly. I can muddle my way through but I don't really have the knowledge required to make good suggestions at the moment.

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,83 @@
...
+
+We will assume you already know how to configure webpack for a client-only project. The config for an SSR project will be largely similar. This example shows a config for Vue CLI project but can be adapted to be used on any webpack build.
+
+```js
+const ManifestPlugin = require('webpack-manifest-plugin')
+const nodeExternals = require('webpack-node-externals')
+const webpack = require('webpack')
```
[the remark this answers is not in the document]
NataliaTepluhina: Agreed! Will restructure the page 👍🏻

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,129 @@
...
+```js
+// entry-server.js
+const { app, router } = createApp({
+  /*...*/
+})
+```
+
+## Code-Splitting
+
+Code-splitting, or lazy-loading part of your app, helps reducing the amount of assets that need to be downloaded by the browser for the initial render, and can greatly improve TTI (time-to-interactive) for apps with large bundles. The key is "loading just what is needed" for the initial screen.
```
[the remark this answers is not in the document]
NataliaTepluhina: Changed to `size` and applied 👍🏻

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,129 @@
...
+router.isReady().then(() => {
+  app.mount('#app')
+})
+```
+
+For server part, we need to update our `server.js` script:
+
+```js
+// server.js
+const createApp = require('./app')
```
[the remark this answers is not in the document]
NataliaTepluhina: @danielroe will it work fine 🤔 I assumed that this an Express server and it requires ES5 `require` as we're running it as a Node script

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
...
+    template: `<div>Current user is: {{ user }}</div>`
+  })
+}
+```
+
+And our server code now becomes:
+
+```js
+// server.js
+const createApp = require('./app')
```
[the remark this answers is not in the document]
NataliaTepluhina: Same note as about about Express server

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Externals Caveats
+
+Notice that in the `externals` option we are whitelisting CSS files. This is because CSS imported from dependencies should still be handled by webpack. If you are importing any other types of files that also rely on webpack (e.g. `*.vue`, `*.sass`), you should add them to the whitelist as well.
+
+If you are using `runInNewContext: 'once'` or `runInNewContext: true`, then you also need to whitelist polyfills that modify `global`, e.g. `babel-polyfill`. This is because when using the new context mode, **code inside a server bundle has its own `global` object.** Since you don't really need it on the server when using Node 7.6+, it's actually easier to just import it in the client entry.
+
+### Generating `clientManifest`
+
+In addition to the server bundle, we can also generate a client build manifest. With the client manifest and the server bundle, the renderer now has information of both the server _and_ client builds, so it can automatically infer and inject [preload / prefetch directives](https://css-tricks.com/prefetching-preloading-prebrowsing/) and css links / script tags into the rendered HTML.
```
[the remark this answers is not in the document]
NataliaTepluhina: I'll better reword this one 👍🏻

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,129 @@
...
+
+server.get('*', async (req, res) => {
+  const { app, router } = await createApp()
```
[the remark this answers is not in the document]
NataliaTepluhina: @skirtles-code we have an asynchronous code in the client entry 🤔  wouldn't it need `await` here?

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,83 @@
...
+const nodeExternals = require('webpack-node-externals')
+const webpack = require('webpack')
+
+module.exports = {
+  chainWebpack: webpackConfig => {
+    //
+    webpackConfig.module.rule('vue').uses.delete('cache-loader')
+    webpackConfig.module.rule('js').uses.delete('cache-loader')
+    webpackConfig.module.rule('ts').uses.delete('cache-loader')
+    webpackConfig.module.rule('tsx').uses.delete('cache-loader')
```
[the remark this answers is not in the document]
NataliaTepluhina: Comment added 👍🏻

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,83 @@
...
+    )
+  }
+}
+```
+
+### Externals Caveats
+
+Notice that in the `externals` option we are whitelisting CSS files. This is because CSS imported from dependencies should still be handled by webpack. If you are importing any other types of files that also rely on webpack (e.g. `*.vue`, `*.sass`), you should add them to the whitelist as well.
+
+If you are using `runInNewContext: 'once'` or `runInNewContext: true`, then you also need to whitelist polyfills that modify `global`, e.g. `babel-polyfill`. This is because when using the new context mode, **code inside a server bundle has its own `global` object.** Since you don't really need it on the server when using Node 7.6+, it's actually easier to just import it in the client entry.
```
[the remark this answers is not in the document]
NataliaTepluhina: Node version removed 👍🏻

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,129 @@
+# Routing and Code-Splitting
+
+## Routing with `vue-router`
+
+You may have noticed that our server code uses a `*` handler which accepts arbitrary URLs. This allows us to pass the visited URL into our Vue app, and reuse the same routing config for both client and server!
+
+It is recommended to use the official [vue-router](https://github.com/vuejs/vue-router-next) library for this purpose. Let's first create a file where we create the router. Note that similar to `createRootComponent`, we also need a fresh router instance for each request, so the file exports a `createRouter` function:
```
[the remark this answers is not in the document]
NataliaTepluhina: Removed 👍🏻

src/guide/ssr/hydration.md
```diff
@@ -0,0 +1,30 @@
+# Client Side Hydration
+
+Hydration refers to the client-side process during which Vue takes over the static HTML sent by the server and turns it into dynamic DOM that can react to client-side data changes.
+
+In `entry-client.js`, we are simply mounting the app with this line:
+
+```js
+// this assumes App.vue template root element has `id="app"`
```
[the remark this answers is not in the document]
NataliaTepluhina: Removed 👍🏻

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,129 @@
...
+Code-splitting, or lazy-loading part of your app, helps reducing the amount of assets that need to be downloaded by the browser for the initial render, and can greatly improve TTI (time-to-interactive) for apps with large bundles. The key is "loading just what is needed" for the initial screen.
+
+Vue provides async components as a first-class concept, combining it with [webpack 2's support for using dynamic import as a code-split point](https://webpack.js.org/guides/code-splitting-async/), all you need to do is:
+
+```js
+// changing this...
+import User from './User.vue'
+
+// to this:
+const User = () => import('./User.vue')
```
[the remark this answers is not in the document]
NataliaTepluhina: Changed to `defineAsyncComponent` here as we mention Vue support for async components

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
...
+
+So the basic idea is we will be using webpack to bundle our app for both client and server - the server bundle will be required by the server and used for SSR, while the client bundle is sent to the browser to hydrate the static markup.
+
+![architecture](https://cloud.githubusercontent.com/assets/499550/17607895/786a415a-5fee-11e6-9c11-45a2cfdf085c.png)
+
+We will discuss the details of the setup in later sections - for now, let's just assume we've got the build setup figured out and we can write our Vue app code with webpack enabled.
+
+## Code Structure with webpack
+
+Now that we are using webpack to process the app for both server and client, the majority of our source code can be written in a universal fashion, with access to all the webpack-powered features. At the same time, there are [a number of things](./universal.md) you should keep in mind when writing universal code.
```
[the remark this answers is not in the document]
NataliaTepluhina: Fixed 👍🏻

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,142 @@
...
+export default function(args) {
+  const rootComponent = {
+    render: () => h(App),
+    components: { App },
+    setup() {
+      // additional application logic here
+    }
+  }
+
+  const app = createSSRApp(rootComponent)
```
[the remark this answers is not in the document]
NataliaTepluhina: Changed to match Vite SSR examples 👍🏻

src/guide/ssr/routing.md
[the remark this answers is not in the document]
skirtles-code: I'm getting a bit lost with multiple things having the same name, but as far as I can tell this isn't using the client entry. The `createApp` being called here isn't `async` and doesn't return a promise, so awaiting shouldn't be necessary. I've now got all of this code running locally (not sure I'm doing it right though) and it doesn't seem to need the `await`.

skirtles-code (changes requested):
I had a go at running the code in the examples, starting from a fresh CLI project. I've included some notes/suggestions below based on the problems I hit.

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,117 @@
...
+router.isReady().then(() => {
+  app.mount('#app')
+})
+```
+
+We also need to update our `server.js` script:
+
+```js
+// server.js
+const createApp = require('./app')
```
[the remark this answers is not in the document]
NataliaTepluhina: @skirtles-code hmmm, for me it works with `require('./app')` just fine 🤔 I'll leave it like it is for now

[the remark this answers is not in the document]
NataliaTepluhina: Fixed to be

```
const path = require('path')

const appPath = path.join(__dirname, './dist', 'server', manifest['app.js'])
const createApp = require(appPath).default
```

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,89 @@
+# Build Configuration
+
+The webpack config for an SSR project will be similar to a client-only project. If you're not familiar with configuring webpack, you can find more information in the documentation for [Vue CLI](https://cli.vuejs.org/guide/webpack.html#working-with-webpack) or [configuring Vue Loader manually](https://vue-loader.vuejs.org/guide/#manual-setup).
+
+### Example configuration
+
+Below is a sample `vue.config.js` that adds SSR rendering to a Vue CLI project, but it can be adapted for any webpack build.
+
+```js
+const ManifestPlugin = require('webpack-manifest-plugin')
```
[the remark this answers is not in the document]
NataliaTepluhina: Replaced 👍🏻

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,129 @@
...
+
+```js
+// changing this...
+import User from './User.vue'
+
+// to this:
+const User = () => import('./User.vue')
+```
+
+Note that it is still necessary to use `router.isReady` on both server and client before returning / mounting the app, because the router must resolve async route components ahead of time in order to properly invoke in-component hooks. Let's update our client entry:
```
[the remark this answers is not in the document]
NataliaTepluhina: I've rephrased it and added a link to VueRouter `isReady` API 👍🏻

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,89 @@
...
+module.exports = {
+  chainWebpack: webpackConfig => {
+    // We need to disable cache loader, otherwise the client build
+    // will used cached components from the server build
+    webpackConfig.module.rule('vue').uses.delete('cache-loader')
+    webpackConfig.module.rule('js').uses.delete('cache-loader')
+    webpackConfig.module.rule('ts').uses.delete('cache-loader')
+    webpackConfig.module.rule('tsx').uses.delete('cache-loader')
+
+    if (!process.env.SSR) {
```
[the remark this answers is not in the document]
NataliaTepluhina: @skirtles-code good point! My code is slightly different but I believe we should extend the guide with the

1) server code
2) package.json scripts to run the build correctly.

We don't need `.env` as we can set `SSR=1` in the build script. I'll add this section to the guide

src/guide/ssr/universal.md
```diff
@@ -0,0 +1,33 @@
+# Writing Universal Code
```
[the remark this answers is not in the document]
NataliaTepluhina: Added :+1

src/guide/ssr/build-config.md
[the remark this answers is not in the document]
NataliaTepluhina: Added a server code and a few more changes to the build configuration. including `package.json` scripts 👍🏻

NataliaTepluhina:
@skirtles-code @danielroe thank you both for the great review! I added some changes, created a server configuration part and addressed your comments. Would you mind giving a second look to this one?

NataliaTepluhina:
@CyberAP I would be really happy if you could take a look too 🙏🏻 I believe your input on SSR is super valuable

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,107 @@
...
+    // This allows webpack to handle dynamic imports in a Node-appropriate
+    // fashion, and also tells `vue-loader` to emit server-oriented code when
+    // compiling Vue components.
+    webpackConfig.target('node')
+    // This tells the server bundle to use Node-style exports
+    webpackConfig.output.libraryTarget('commonjs2')
+
+    webpackConfig
+      .plugin('manifest')
+      .use(new ManifestPlugin({ fileName: 'ssr-manifest.json' }))
```
CyberAP, line 65: Must've been a typo.

```suggestion
      .use(new WebpackManifestPlugin({ fileName: 'ssr-manifest.json' }))
```

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,107 @@
...
+
+Below is a sample `vue.config.js` that adds SSR rendering to a Vue CLI project, but it can be adapted for any webpack build.
+
+```js
+const { WebpackManifestPlugin } = require('webpack-manifest-plugin')
+const nodeExternals = require('webpack-node-externals')
+const webpack = require('webpack')
+
+module.exports = {
+  chainWebpack: webpackConfig => {
```
CyberAP, line 33: I was thinking whether it is a good idea or not to have 3 configs here?
One for common rules likes aliases, one for client only and one for server only. Both of them deriving from the base config (using `webpackMerge`).

The reasoning behind this is that using `chainWebpack` for re-configuration is hardly maintainable. It does get very complex extremely fast so noone want to touch this config because noone understands what happens here. I have the same setup on my project and it overwrites a ton of rules exactly like shown here: replacements\deletions inside `if` statements which depend on environment. I think a static webpack config is way more readable because you can understand what it does at the first glance just by the structure of that config (given you have an experience with webpack). This programmatic config is hard to read because you have to parse it line by line working as an interpreter. It is also unclear what the base vue-cli config looks like, so all the `delete` calls look like magic to me.

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,107 @@
...
+    // Externalize app dependencies. This makes the server build much faster
+    // and generates a smaller bundle file.
+
+    // Do not externalize dependencies that need to be processed by webpack.
+    // You should also whitelist deps that modify `global` (e.g. polyfills)
+    webpackConfig.externals(nodeExternals({ allowlist: /\.(css|vue)$/ }))
+
+    webpackConfig.optimization.splitChunks(false).minimize(false)
+
+    webpackConfig.plugins.delete('hmr')
```
CyberAP, line 78: `hmr` is actually critical for a good DX when working in an SSR environment. It is tricky to set it up properly with SSR, so maybe it's not a bad idea to make a separate `vue-webpack-ssr-hmr` package just for that.

src/guide/ssr/universal.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Access to Platform-Specific APIs
+
+Universal code cannot assume access to platform-specific APIs, so if your code directly uses browser-only globals like `window` or `document`, they will throw errors when executed in Node.js, and vice-versa.
+
+For tasks shared between server and client but using different platform APIs, it's recommended to wrap the platform-specific implementations inside a universal API, or use libraries that do this for you. For example, [axios](https://github.com/axios/axios) is an HTTP client that exposes the same API for both server and client.
+
+For browser-only APIs, the common approach is to lazily access them inside client-only lifecycle hooks.
+
+Note that if a 3rd party library is not written with universal usage in mind, it could be tricky to integrate it into an server-rendered app. You _might_ be able to get it working by mocking some of the globals, but it would be hacky and may interfere with the environment detection code of other libraries.
```
CyberAP, line 25: I would really like to see some common examples of 3rd party library integrations that do not support SSR evironment. It's probably the most common issue when you're switching from CSR to SSR.

The approaches I would suggest are:
1. Using dynamic imports plus mounted flags (set a flag to `true` when mounted, which triggers dynamic import for components hidden with this flag)
2. Using environment constants: `IS_SERVER`, `IS_CLIENT` and how to define them.

src/guide/ssr/universal.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Custom Directives
+
+Most [custom directives](/guide/custom-directive.html#custom-directives) directly manipulate the DOM, which will cause errors during SSR. There are two ways to work around this:
+
+1. Prefer using components as the abstraction mechanism and work at the Virtual-DOM level (e.g. using [render functions](/guide/render-function.html#the-dom-tree)) instead.
```
CyberAP, line 31: I think the idea here is to replace directives with components as much as possible, but that exact wording is a bit unclear to me unfortunately. It is also unclear how render functions could help me in that case (an example would be much welcomed I think).

src/guide/ssr/universal.md
```diff
@@ -0,0 +1,33 @@
...
+
+2. If you have a custom directive that cannot be easily replaced by components, you can provide a "server-side version" of it using the `directives` option when creating the server renderer.
```
CyberAP, line 33: I think a link to an example here would be great for quicker migration to SSR.

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,107 @@
...
+
+2. We should [externalize application dependencies](https://webpack.js.org/configuration/externals/). This makes the server build much faster and generates a smaller bundle file. When doing this, we have to exclude dependencies that need to be processed by webpack (like `.css`. or `.vue` files).
+
+3. We need to change webpack [target](https://webpack.js.org/concepts/targets/) to Node.js. This allows webpack to handle dynamic imports in a Node-appropriate fashion, and also tells `vue-loader` to emit server-oriented code when compiling Vue components.
+
+4. When building a server entry, we would need to define an environment variable to indicate we are working with SSR. It might be helpful to add a few `scripts` to the project's `package.json`:
+
+```json
+"scripts": {
+  "build:client": "vue-cli-service build --dest dist/client",
```
CyberAP, line 17: Maybe it's worth exploring into the `vue-cli` agnostic SSR idea? Because right now if you're not using `vue-cli` your first migration step would be to migrate your current build config to `vue-cli` and only then apply all the SSR practies in the guide.

src/guide/ssr/getting-started.md
```diff
@@ -64,6 +64,8 @@ yarn add express
 ```
 
 ```js
+// server.js
```
CyberAP, line 67: Since most of the people who want to migrate from CSR to SSR would be having a some kind of `main.js` entry file maybe it would make sense to show this migration path as a step-by-step guide how we extract our current code from `main.js` and split it into three different files: `app.js`, `entry-client.js` and `entry-server.js`. The current approach is also very good, but I think it's more focused on writing an SSR app from scracth. Maybe there should be a separate guide for migrating from CSR to SSR? What do you think?

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,121 @@
...
+
+server.get('*', async (req, res) => {
+  const { app, router } = createApp()
+
+  router.push(req.url)
+  await router.isReady()
+
+  const appContent = await renderToString(app)
+
+  fs.readFile(path.join(__dirname, '/dist/client/index.html'), (err, html) => {
```
CyberAP, line 109: I think this code can be improved by moving the index template reading into some abstract function like `getIndexTemplate`, so we don't mix abstractions and also do some optimisation (right now the file is read on each request which might affect performance).

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,121 @@
...
+
+It is recommended to use the official [vue-router](https://github.com/vuejs/vue-router-next) library for this purpose. Let's first create a file where we create the router. Note that similar to application instance, we also need a fresh router instance for each request, so the file exports a `createRouter` function:
+
+```js
+// router.js
+import { createRouter, createMemoryHistory, createWebHistory } from 'vue-router'
+
+const isServer = typeof window === 'undefined'
+
+const history = isServer ? createMemoryHistory() : createWebHistory()
```
CyberAP, line 15: I think a slightly better approach here would be to pass the `history` instance on an entry-level, so we have better code-splitting here and won't have to check for `isServer` flag.

src/guide/ssr/hydration.md
```diff
@@ -0,0 +1,35 @@
...
+
+Since the server has already rendered the markup, we obviously do not want to throw that away and re-create all the DOM elements. Instead, we want to "hydrate" the static markup and make it interactive.
+
+Vue provides a `createSSRApp` method for use in client-side code (in this case, in our `entry-client.js`) to tell Vue to hydrate the existing static HTML instead of re-creating all the DOM elements.
+
+### Hydration Caveats
+
+In development mode, Vue will assert the client-side generated virtual DOM tree matches the DOM structure rendered from the server. If there is a mismatch, it will bail hydration, discard existing DOM and render from scratch. There will be a warning in the browser console but your site will still work.
+
+However, in production mode, this assertion is disabled for maximum performance, which means that **it is crucial you ensure your client- and server-side DOM trees match**. Otherwise your HTML will remain static and non-reactive.
```
CyberAP, line 19: I am not sure about this one, does it really fail completely on hydration mismatch? I am not able to test this right now, but I remember that hydration algorithm in Vue 3 was significantly improved so the app doesn't crash completely in case of a mismatch.

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,121 @@
+# Routing and Code-Splitting
```
CyberAP, line 1: As an SSR app developer I am also curious how Critical CSS should be implemeneted correctly with this approach.

CyberAP:
I am also curious whether something about vue-meta integration has changed. In Vue 2 it was important for meta config to be defined only on route config or on a page-level component (a component that can be retrieved from the route config.

Vue 3 also has an improved rendering mechanism which allows for a much flexible data fetching. Previously you must declare your data fetching hooks on a page-level component which significantly limited component composition and data flow. In Vue 3 we have async setup and Suspense that remove that limitation. I think that might also be important to cover.

Lastly, there's a renderToString method in vue server-render package. It produces a lower time to first byte and can increase lighthouse metrics. But it also imposes some restrictions on how you should write your code: no synchronous Teleports (can not replace the markup that has already been sent) or async meta configs (or at least we should wait for the meta to resolve before sending any data, since head tag goes first).

src/guide/ssr/hydration.md
danielroe: Good point. I'd need to check this too.

NataliaTepluhina: @CyberAP I still can see a warning in the console for this but in prod mode, it's the same as in dev: app just renders from scratch without crashing 👍🏻 I'll unify both paragraphs removing a distinction for modes

src/guide/ssr/build-config.md
NataliaTepluhina: I agree with the point but I have a few questions regarding it:
- should we move away from Vue CLI in this case? I am not sure how it's done correctly in the CLI environment as I usually always went with `configureWebpack` / `chainWebpack` in the `vue.config.js

src/guide/ssr/build-config.md
NataliaTepluhina: Removed the rule deletion for now and I'll transfer the idea about package to the team 👍🏻

src/guide/ssr/universal.md
NataliaTepluhina: Good point! For the sake of iteration and to prevent growing this PR further, I am transfering this fo the follow-up issue to improve the guide: https://github.com/vuejs/docs-next/issues/968

src/guide/ssr/build-config.md
NataliaTepluhina: Added a follow-up to refactor to CLI agnostic build with a base, server and client configs: https://github.com/vuejs/docs-next/issues/969

src/guide/ssr/build-config.md
NataliaTepluhina: Follow-up: https://github.com/vuejs/docs-next/issues/969

src/guide/ssr/getting-started.md
NataliaTepluhina: Agreed, it would be a great improvement! Created a follow-up: https://github.com/vuejs/docs-next/issues/970

src/guide/ssr/build-config.md
CyberAP: Yeah, this is probably not feasible with current vue-cli API, so it should be a part of vue-cli agnostic build in that case.

src/guide/ssr/routing.md
NataliaTepluhina: Creat catch! Replaced with `readFileSync` and moved out of the server callback :+1:

src/guide/ssr/routing.md
NataliaTepluhina: I agree! However, this guide is not full and there are a few missing parts here. Here is an issue with 'what's currently missing': https://github.com/vuejs/docs-next/issues/971

src/guide/ssr/universal.md
NataliaTepluhina: Rephrased a bit and removed a second point (see below)

src/guide/ssr/universal.md
NataliaTepluhina: As I don't have an example right now (and it will probably take some time to figure it out with the new renderer) I removed this part for now. Will return it back as soon as we have an example 👍🏻

NataliaTepluhina:
@CyberAP thank you so much for the great comments! I addressed 'smaller' ones and created follow-ups for the bigger refactorings 👍🏻 Please feel free to add more points here or to follow-up issues, your input is much appreciated

skirtles-code (changes requested):
A few more suggestions from me.

src/guide/ssr/server.md
```diff
@@ -0,0 +1,101 @@
...
+  server.use('/js', express.static(path.join(__dirname, './dist/client', 'js')))
+  server.use(
+    '/css',
+    express.static(path.join(__dirname, './dist/client', 'css'))
+  )
+  server.use(
+    '/favicon.ico',
+    express.static(path.join(__dirname, './dist/client', 'favicon.ico'))
+  )
+  ```
```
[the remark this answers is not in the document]
NataliaTepluhina: @skirtles-code I prefer to leave it as it is as I believe it's more readable.

src/guide/ssr/structure.md
```diff
@@ -0,0 +1,143 @@
...
+      return {
+        user: 'John Doe'
+      }
+    },
+    template: `<div>Current user is: {{ user }}</div>`
+  })
+}
+```
+
+And our server code now becomes:
```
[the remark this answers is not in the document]
NataliaTepluhina: Fixed the server code

src/guide/ssr/build-config.md
```diff
@@ -0,0 +1,106 @@
...
+  "build:server": "SSR=1 vue-cli-service build --dest dist/server",
+  "build": "rm -rf ./dist && npm run build:client && npm run build:server",
```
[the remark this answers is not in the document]
NataliaTepluhina: Removed

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,121 @@
...
+const { app, router } = createApp({
+  /*...*/
+})
+```
+
+## Code-Splitting
+
+Code-splitting, or lazy-loading part of your app, helps reduce the size of assets that need to be downloaded by the browser for the initial render, and can greatly improve TTI (time-to-interactive) for apps with large bundles. The key is "loading just what is needed" for the initial screen.
+
+Vue provides [async components support](/guide/component-dynamic-async.html#async-components) with `defineAsyncComponent` method, allowing [webpack to code-split at that point](https://webpack.js.org/guides/code-splitting-async/). All you need to do is:
```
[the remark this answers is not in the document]
NataliaTepluhina: Fixed to use router dynamic import

NataliaTepluhina:
@skirtles-code I've addressed your comments, could you please check again?

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,121 @@
...
+```js
+// router.js
+import { createRouter, createMemoryHistory, createWebHistory } from 'vue-router'
+import MyUser from './components/MyUser.vue'
+
+const isServer = typeof window === 'undefined'
+
+const history = isServer ? createMemoryHistory() : createWebHistory()
+
+const routes: [{ path: '/user', component: MyUser }],
```
skirtles-code, line 18:
```suggestion
const routes = [{ path: '/user', component: MyUser }]
```

src/guide/ssr/routing.md
```diff
@@ -0,0 +1,121 @@
...
+
+```js
+// change this...
+import MyUser from './components/MyUser.vue'
+const routes: [{ path: '/user', component: MyUser }],
+
+// to this:
+routes: [{
+  { path: '/user', component: () => import('./components/MyUser.vue') }
+}]
```
skirtles-code, line 74:
```suggestion
const routes = [{ path: '/user', component: MyUser }]

// to this:
const routes = [
  { path: '/user', component: () => import('./components/MyUser.vue') }
]
```

skirtles-code (approved):
Last couple of small changes from me and then I think we're good to merge.

NataliaTepluhina:
Thanks to everyone who was reviewing this monster! 🙌🏻  I am merging this one and then we'll iterate to shape it through the follow up issues

iGiliy:
Excuse me, How to use asyncData in SSR?

iGiliy:
> Excuse me, How to use asyncData in SSR?

I tried to use suspense, but <Suspense> is an experimental feature and its API will likely change.

PR #950, opened 2021-03-18
Update options-data.md

likev:
## Description of Problem
from Vue.js 2.x you can not just watch an expression 

## Additional Information
https://github.com/vuejs/vue/issues/844#issuecomment-271144849

### DIFF
--- a/src/api/options-data.md
+++ b/src/api/options-data.md
@@ -177,7 +177,7 @@
 
 - **Details:**
 
-  An object where keys are data properties or computed properties to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
+  An object where keys are reactive properties — examples include [data](/api/options-data.html#data-2) or [computed](/api/options-data.html#computed) properties — to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
 
 - **Example:**
 
--- a/src/api/options-data.md
+++ b/src/api/options-data.md
@@ -177,7 +177,7 @@
 
 - **Details:**
 
-  An object where keys are reactive properties — examples include [data](/api/options-data.html#data-2) or [computed](/api/options-data.html#computed) properties — to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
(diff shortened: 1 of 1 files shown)


src/api/options-data.md
```diff
@@ -177,7 +177,7 @@
 
 - **Details:**
 
-  An object where keys are expressions to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
+  An object where keys are data properties or computed properties to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
```
bencodezen, line 180:
```suggestion
  An object where keys are reactive properties to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
```

bencodezen: What do you think about this update @likev?

bencodezen (changes requested):
What do you think about these updates?

src/api/options-data.md
likev: good!

likev:
> 
> 
> What do you think about these updates?

I think we can merge and close the pull

skirtles-code:
I personally find this wording a little difficult to follow. The 'to watch' after the second dash throws me off. I tried reading it out loud and I don't seem to be able to put the emphasis in the right place.

src/api/options-data.md
```diff
@@ -177,7 +177,7 @@
 
 - **Details:**
 
-  An object where keys are expressions to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
+  An object where keys are reactive properties — examples include [data](/api/options-data.html#data-2) or [computed](/api/options-data.html#computed) properties — to watch and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
```
bencodezen, line 180:
```suggestion
  An object where keys are reactive properties to watch — examples include [data](/api/options-data.html#data-2) or [computed](/api/options-data.html#computed) properties — and values are the corresponding callbacks. The value can also be a string of a method name, or an Object that contains additional options. The component instance will call `$watch()` for each entry in the object at instantiation. See [$watch](instance-methods.html#watch) for more information about the `deep`, `immediate` and `flush` options.
```

bencodezen:
@skirtles-code Good observation! What do you think of the movement of "to watch" earlier on to better separate out the concepts?

skirtles-code:
I like it. Merging. Thank you both.

PR #951, opened 2021-03-18
fix error about deep option of watch in instance-methods.md

likev:

### DIFF
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -121,7 +121,9 @@
 
 - **Option: deep**
 
-  To also detect nested value changes inside Objects, you need to pass in `deep: true` in the options argument. Note that you don't need to do so to listen for array mutations.
+  To also detect nested value changes inside Objects, you need to pass in `deep: true` in the options argument. Note that in Vue3 you **also** need to do so to listen for array mutations.
+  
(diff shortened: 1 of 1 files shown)


src/api/instance-methods.md
```diff
@@ -121,7 +121,9 @@
 
 - **Option: deep**
 
-  To also detect nested value changes inside Objects, you need to pass in `deep: true` in the options argument. Note that you don't need to do so to listen for array mutations.
+  To also detect nested value changes inside Objects, you need to pass in `deep: true` in the options argument. Note that in Vue3 you **also** need to do so to listen for array mutations.
```
NataliaTepluhina, line 124: **suggestion**: this is a documentation for v3, I don't think we need an additional reminder about the current version here. WDYT about rephrasing:

```suggestion
  To also detect nested value changes inside Objects, you need to pass in `deep: true` in the options argument. This option also can be used to watch array mutations.
```

likev: agree with you

skirtles-code: @likev If you're happy with the change could you please merge it into your branch? You should see a button beneath the suggested change that will commit it for you. Thanks.

likev: There are a lot I need to learn about github, thanks!

NataliaTepluhina (approved):
@likev thank you for the implementing the suggestion!

PR #952, opened 2021-03-18
vueschool free-weekend banner

rahaug:
This PR adds a banner on top of vuejs.org to inform users about Vue School’s free-weekend promotion.

The banner is closable and opens Vue School's free-weekend promo page in a new tab. The banner will not reappear if a user has closed it (stored in local storage).

We’ve based the promotion on a previously used banner.

[Click here to see screenshots of the banner](https://imgur.com/a/lVK0sor)

[PR for V2](https://github.com/vuejs/vuejs.org/pull/2788)


yyx990803:
After dismissing the banner, it still shows up for a split second and then disappear on subsequent reloads (maybe due to late CSS load). We should make it hidden by default to prevent that.

rahaug:
Thanks for the feedback @yyx990803!

I am not able to reproduce the issue, but we've changed the logic so it's hidden by default as suggested.

Please let me know if this isn't sufficient.

PR #956, opened 2021-03-21
proposal of 'allow listing' as an inclusive alternative to 'whitelisting'

joaomelo:
## Description of Problem
Inclusive language has gained importance by the growing awareness of language's effects on minorities' struggle for equality. Alternatives are substituting terms like blacklist, master, and slave with words without race connotation.

## Proposed Solution
Changed occurrences of whitelisting to allow listing.

## Additional Information
Did not change the reference in the link ".../src/globalsWhitelist.ts#L3" inside file "\src\guide\template-syntax.md" since it is a link to another repo. A change there will need to take place before or in sync.

### DIFF
--- a/src/guide/migration/custom-elements-interop.md
+++ b/src/guide/migration/custom-elements-interop.md
@@ -21,7 +21,7 @@ If we want to add a custom element defined outside of Vue (e.g. using the Web Co
 
 ### 2.x Syntax
 
-In Vue 2.x, allow listing tags as custom elements was done via `Vue.config.ignoredElements`:
+In Vue 2.x, configuring tags as custom elements was done via `Vue.config.ignoredElements`:
 
 ```js
 // This will make Vue ignore custom element defined outside of Vue
--- a/src/guide/migration/custom-elements-interop.md
+++ b/src/guide/migration/custom-elements-interop.md
@@ -7,7 +7,7 @@ badges:
 
 ## Overview
 
-- **BREAKING:** Custom elements whitelisting is now performed during template compilation, and should be configured via compiler options instead of runtime config.
+- **BREAKING:** Custom elements allow listing is now performed during template compilation, and should be configured via compiler options instead of runtime config.
 - **BREAKING:** Special `is` prop usage is restricted to the reserved `<component>` tag only.
 - **NEW:** There is new `v-is` directive to support 2.x use cases where `is` was used on native elements to work around native HTML parsing restrictions.
 
@@ -21,7 +21,7 @@ If we want to add a custom element defined outside of Vue (e.g. using the Web Co
 
 ### 2.x Syntax
 
-In Vue 2.x, whitelisting tags as custom elements was done via `Vue.config.ignoredElements`:
+In Vue 2.x, allow listing tags as custom elements was done via `Vue.config.ignoredElements`:
 
 ```js
 // This will make Vue ignore custom element defined outside of Vue
--- a/src/guide/migration/custom-elements-interop.md
+++ b/src/guide/migration/custom-elements-interop.md
@@ -7,7 +7,7 @@ badges:
 
 ## Overview
 
-- **BREAKING:** Custom elements allow listing is now performed during template compilation, and should be configured via compiler options instead of runtime config.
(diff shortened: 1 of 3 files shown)


skirtles-code:
Thank you for the PR.

I think the term 'allow listing' would not be widely understood. In these specific sentences I think it will be read as meaning 'to allow the process of listing' rather than 'a list of things that are allowed'.

Perhaps it would be better to reword the sentences rather than trying to find an appropriate synonym?

joaomelo:
I aimed to find some a synonym in order to be consistent with the original page content since the whole documentation is so well written.

I also found "passlist" and "safelist" as alternatives. If you think one of these options is more appropriate, I can make the change. If the issue persists, I sure can try to rephrase.

src/guide/migration/custom-elements-interop.md
```diff
@@ -7,7 +7,7 @@ badges:
 
 ## Overview
 
-- **BREAKING:** Custom elements whitelisting is now performed during template compilation, and should be configured via compiler options instead of runtime config.
+- **BREAKING:** Custom elements allow listing is now performed during template compilation, and should be configured via compiler options instead of runtime config.
```
skirtles-code, line 10:
```suggestion
- **BREAKING:** The checks to determine whether tags should be treated as custom elements are now performed during template compilation, and should be configured via compiler options instead of runtime config.
```

src/guide/migration/custom-elements-interop.md
```diff
@@ -21,7 +21,7 @@ If we want to add a custom element defined outside of Vue (e.g. using the Web Co
 
 ### 2.x Syntax
 
-In Vue 2.x, whitelisting tags as custom elements was done via `Vue.config.ignoredElements`:
+In Vue 2.x, allow listing tags as custom elements was done via `Vue.config.ignoredElements`:
```
skirtles-code, line 24:
```suggestion
In Vue 2.x, configuring tags as custom elements was done via `Vue.config.ignoredElements`:
```

src/guide/migration/introduction.md
```diff
@@ -89,7 +89,7 @@ The following consists a list of breaking changes from 2.x:
 
 ### Custom Elements
 
-- [Custom elements whitelisting is now performed during template compilation](/guide/migration/custom-elements-interop.html)
+- [Custom elements allow listing is now performed during template compilation](/guide/migration/custom-elements-interop.html)
```
skirtles-code, line 92:
```suggestion
- [Custom element checks are now performed during template compilation](/guide/migration/custom-elements-interop.html)
```

src/guide/template-syntax.md
```diff
@@ -206,4 +206,4 @@ Unless you have a "someattr" property in your instance, your code won't work.
 
 #### JavaScript Expressions
 
-Template expressions are sandboxed and only have access to a [whitelist of globals](https://github.com/vuejs/vue-next/blob/master/packages/shared/src/globalsWhitelist.ts#L3) such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.
+Template expressions are sandboxed and only have access to a [allowlist of globals](https://github.com/vuejs/vue-next/blob/master/packages/shared/src/globalsWhitelist.ts#L3) such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.
```
skirtles-code, line 209:
```suggestion
Template expressions are sandboxed and only have access to a [restricted list of globals](https://github.com/vuejs/vue-next/blob/master/packages/shared/src/globalsWhitelist.ts#L3) such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.
```

skirtles-code (changes requested):
I've made some suggestions for how we might reword some of these.

Feel free to suggest an alternative wording if you prefer.

joaomelo:
All fine by me. 
Thanks for the attention and awesome framework. 
I love Vue :)

skirtles-code:
@joaomelo My first suggestion was marked as resolved but it doesn't seem to have been committed. If you're happy with it could you merge it in please?

joaomelo:
I thought I accepted. Can you check again now?

skirtles-code (approved):

PR #958, opened 2021-03-23
Correct "Handling `v-model` modifiers" URL

jbdebiasio:


NataliaTepluhina (approved):
@jbdebiasio thank you for the fix!

PR #973, opened 2021-03-28
Added Vite mention to SSR guide

NataliaTepluhina:
## Description of Problem

Currently, our SSR guide mentions only DIY approach, Quasar and Nuxt. With Vite SSR becoming more stable we should add it to the list of possible approaches


bencodezen (approved):
This is awesome! Good to go from my end! 🙌

skirtles-code (approved):

PR #979, opened 2021-04-01
Link vue-router-next in routing page

cypressious:
## Description of Problem

The link is to the repository of vue-router 3 for Vue 2.

## Proposed Solution

Link the repository of vue-router 4.


NataliaTepluhina (approved):
Thanks for the fix!

PR #986, opened 2021-04-08
Add banner for vuemastery free weekend

pieer:


yyx990803:
There is a slight flash of the banner on subsequent reloads after it's been dismissed once. Can we fix that?

PR #989, opened 2021-04-10
Fix Vue mastery banner

pieer:


skirtles-code:
@pieer I believe this is effectively the same fix that I made in #988 (opened a few seconds before you opened this).

While I have you...

I also noticed that the page jumps when I scroll down. It seems to happen at the point it transitions into 'fixed' mode.

Not sure if it's related but it seems there's a `vuemastery-weekend-promo` class being used in the CSS that doesn't seem to exist in the markup.

pieer:
I see, fixing now

PR #992, opened 2021-04-11
Fix jump when banner disapear

pieer:


pieer:
@skirtles-code Here is the change to prevent the jump of the content once the banner disapear

NataliaTepluhina (approved):

skirtles-code:
Thanks. Looks good to me.

PR #1006, opened 2021-04-22
Return values for render functions

skirtles-code:
This adds a section to the page on Render Functions explaining the possible return values.

### DIFF
--- a/src/guide/render-function.md
+++ b/src/guide/render-function.md
@@ -567,6 +567,33 @@ render () {
 }
 ```
 
+## Return Values for Render Functions
+
+In all of the examples we've seen so far, the `render` function has returned a single root VNode. However, there are alternatives.
+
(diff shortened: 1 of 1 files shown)


src/guide/render-function.md
```diff
@@ -567,6 +567,33 @@ render () {
...
+render() {
+  return [
+    'Hello',
+    h('br'),
+    'world!'
+  ]
+}
+```
+
+If a component needs to render nothing, perhaps because data is still loading, it can just return `null`. This will be rendered as a comment node in the DOM.
```
NataliaTepluhina, line 595: **suggestion**: should we mention that we can return `null` as well for renderless components?

skirtles-code: Renderless components aren't currently explained anywhere else in the docs. I don't think we can mention them here without several more sentences explaining what they are.

I'd rather leave this how it is for now.

PR #1008, opened 2021-04-23
Add a section about functional components

skirtles-code:
This adds a short section about functional components to the render functions page. The positioning is the same as in the Vue 2 docs.

I originally wrote a lot more material, including a proper example, but in the end I decided to cut it down to a minimal description of the key technical details and leave it at that.

### DIFF
--- a/src/guide/render-function.md
+++ b/src/guide/render-function.md
@@ -136,9 +136,9 @@ The `h()` function is a utility to create VNodes. It could perhaps more accurate
 ```js
 // @returns {VNode}
 h(
-  // {String | Object | Function } tag
-  // An HTML tag name, a component or an async component.
-  // Using function returning null would render a comment.
+  // {String | Object | Function} tag
+  // An HTML tag name, a component, an async component, or a
+  // functional component.
   //
   // Required.
   'div',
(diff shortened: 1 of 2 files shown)


src/guide/render-function.md
```diff
@@ -136,9 +136,9 @@ The `h()` function is a utility to create VNodes. It could perhaps more accurate
 ```js
 // @returns {VNode}
 h(
-  // {String | Object | Function } tag
-  // An HTML tag name, a component or an async component.
-  // Using function returning null would render a comment.
```
skirtles-code, line 141: A function returning null is not really a special case, it's just a functional component. #1006 mentions that returning `null` creates a comment.

src/guide/transitions-list.md
```diff
@@ -300,7 +300,7 @@ Vue.component('my-special-transition', {
 })
 ```
 
-And [functional components](render-function.html#Functional-Components) are especially well-suited to this task:
+And [functional components](render-function.html#functional-components) are especially well-suited to this task:
```
skirtles-code, line 303: This section needs major work to update it for Vue 3 but at least this link goes somewhere now.

NataliaTepluhina (approved):

PR #1022, opened 2021-04-27
Added template refs typing instruction

### LINKED ISSUE #980: Document template ref typing in TS (i.e. InstanceType)
To prevent [issues like this](https://github.com/vuejs/vue-next/issues/1341), [the typescript section](https://v3.vuejs.org/guide/typescript-support.html) should be amended with an exemplary declaration of a component `ref`'s type using `InstanceType`.

Here's a more elaborate example: https://stackoverflow.com/questions/66907606/whats-vue-3-composition-apis-type-safe-way-of-defining-methods/66909582#66909582

NataliaTepluhina:
Close #980

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -379,7 +379,7 @@ Sometimes you might need to annotate a template ref for a child component in ord
 ```ts
 import { defineComponent, ref } from 'vue'
 
-export default defineComponent({
+const MyModal = defineComponent({
   setup() {
     const isContentShown = ref(false)
     const open = () => (isContentShown.value = true)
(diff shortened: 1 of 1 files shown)


src/guide/typescript-support.md
```diff
@@ -372,6 +372,76 @@ year.value = 2020 // ok!
...
+    }
+  }
+})
+```
+
+We want to call this method via a template ref from the parent component:
+
+```html
+<template>
+  <div id="app">
```
skirtles-code, line 399: I tried to run this example myself. It does work but I found myself unsure which style of component is was trying to use. It doesn't seem self-consistent to me.

The `<template>` tag suggests that this is supposed to be read as an SFC. If that's the case, then we should probably go all-in and combine the two code fragments into a single SFC example. In that scenario, the `id=app"` should probably be removed.

An alternative is that this is supposed to be a root-level `App` component, with an in-DOM template. In that case we probably shouldn't have the `<template>` tags around the template. For an in-DOM template we should also use a proper closing tag for `</my-modal>` rather than relying on a self-closing tag.

skirtles-code (commented):
I can't really comment on the specifics from a TS perspective but I did try running the code and it seemed to work as described.

I've made a suggestion for how we present the first example but overall this looks good to me.

src/guide/typescript-support.md
NataliaTepluhina: @skirtles-code thank you for teh review! I've changed the example to inline template, could you please take a look?

src/guide/typescript-support.md
```diff
@@ -372,6 +372,84 @@ year.value = 2020 // ok!
...
+```
+
+We want to call this method via a template ref from the parent component:
+
+```html
+<template>
+  <button @click="openModal">Open from parent</button>
+  <my-modal ref="modal" />
+</template>
+```
```
skirtles-code, line 402: I think we can remove this code block, it seems to be repeating some of the code from the next code block.

skirtles-code (changes requested):

src/guide/typescript-support.md
NataliaTepluhina: Sorry, totally forgot to remove this one. Should be fine now!

Issue #844, opened 2021-02-06
Docs could use a few tweaks to enhance readability


aldencolerain:
I think the Vue 3 docs could use a few readability enhancements.  I know these suggestions are subjective, but I there may be consensus on some of them or if not perhaps some alternate proposals that would enhance readability.  These ideas are fairly raw, but at very least might spark a discussion on suggestions that could make it a little more clean.

* Bold text does not stand out significantly from the regular text
![image](https://user-images.githubusercontent.com/16270737/107100645-41394f00-67d2-11eb-8948-b449fb17b8ea.png)

* The typography and font size a little large for the narrow content width and the current line spacing.  Making it hard to parse out paragraphs, bullets, and sections.
![image](https://user-images.githubusercontent.com/16270737/107101109-7abe8a00-67d3-11eb-8dd2-d57b7803de3f.png)

* Spacing between code blocks, code pens, call outs, etc is inconsistent and maybe not enough compared to the line spacing. There are multiple images here, its a bit hard to see on the issue.
![image](https://user-images.githubusercontent.com/16270737/107101200-cd984180-67d3-11eb-827f-8f556763c0ae.png)

![image](https://user-images.githubusercontent.com/16270737/107101213-d6891300-67d3-11eb-9204-c7318b8812e3.png)

![image](https://user-images.githubusercontent.com/16270737/107101225-e1dc3e80-67d3-11eb-9ff3-99e32cdd690c.png)

![image](https://user-images.githubusercontent.com/16270737/107101233-e6085c00-67d3-11eb-9d3f-9eca9817fa1f.png)

* The left table of contents bar is fairly wide and there is no margin on the right of the document so although the main content is centered inside the containing page its pretty far off center from the reader's window. Perhaps increasing the content width on larger screens so this gap is less pronounced, especially with the fairly large default font size.  Another idea could be adding a sponsors bar on the right like Vue 2 docs?
![image](https://user-images.githubusercontent.com/16270737/107100944-08e64080-67d3-11eb-91df-15ede01dd2bd.png)

* Font size, vertical spacing, and tab size in the table of contents seem a bit off.  It almost looks like there is more vertical spacing than indentation perhaps because of the large font size and larger weight?
![image](https://user-images.githubusercontent.com/16270737/107101314-2536ad00-67d4-11eb-9f71-a9f14276e68b.png)

* We may not have much control over this, but the code pen blocks are fairly noisy visually and break the flow of the document.  
[...]

PR #1023, opened 2021-04-29
Bold text does stand out significantly from the regular text

Blakelist7:
Made some text more Bold so that they stand out from the regular text.
Partial changes : #844 
Can someone please review my PR


skirtles-code:
Hi @Blakelist7. Thanks for the PR.

Sorry for the delay in getting back to you.

I'm currently doing some research into the history of the docs theme and why it is the way it is. The `font-weight` for `<strong>` tags doesn't match the defaults for VuePress and I'd like to understand why before making any decisions. The people involved are all in different timezones so it may take a few more days for me to get to the bottom of it but I'll let you know as soon as I can.

Thanks again for your effort in putting together this PR.

Blakelist7:
Thanks for acknowledgement.
Once you complete your research, it would be beneficial if you could share the resources also.

PR #1035, opened 2021-05-04
docs(webpack): add installation instructions

evenstensberg:
## Description of Problem

There's no way of knowing how to bundle vue 3 with webpack 5.

## Proposed Solution

Documentation on this

### DIFF
--- a/src/guide/installation.md
+++ b/src/guide/installation.md
@@ -52,6 +52,37 @@ npm is the recommended installation method when building large scale application
 $ npm install vue@next
 ```
 
+### webpack
+
+Install vue 3 dependencies.
+
+```bash
+
+npm install --save-dev @vue/compiler-sfc vue@next vue-loader@^16.0.0-beta.8
+```
+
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@evenstensberg thank you for the contribution! However, in this section, we only cover _Vue-specific_ tools from our own ecosystem. That's why we don't cover webpack or Parcel builds here, for example.

However, I believe we could add a link to the end of the Vue CLI section (CLI uses webpack under the hood), mentioning that if you prefer to set up the webpack build manually, you can use [vue-loader](https://vue-loader.vuejs.org/guide/#manual-setup)

evenstensberg:
Could you add that?

NataliaTepluhina:
@evenstensberg Sure! But I would need to open a different PR because I can't make changed to your forked repository

skirtles-code:
I've been experimenting with ways of wording this and I think it might make more sense to add a note to the `npm` section rather than the `CLI` section. I'm going to ruminate on it a little more but I should have a PR up in the next few days.

PR #1009, opened 2021-04-23
style: unify code style

wenfangdu:
## Description of Problem

Currently, many code samples aren't following the same code style which can lead to confusion, e.g. (excerpted from [here](https://v3.vuejs.org/guide/migration/emits-option.html#_3-x-behavior))

```vue
<template>
  <div>
    <p>{{ text }}</p>
    <button v-on:click="$emit('accepted')">OK</button>
  </div>
</template>
<script>
👇 indent
  export default {
    props: ['text'],
    emits: ['accepted']
  }
</script>
```

```vue
<template>
  <button v-on:click="$emit('click', $event)">OK</button>
</template>
<script>
👇 no indent
export default {
  emits: [] // without declared event
}
</script>
```

## Proposed Solution

1. Prettier codebase.
2. Add `pre-commit` hook for prettier.

## Additional Information

This PR has no other changes except the above actions.

PR #1036, opened 2021-05-05
Whitespace changes suggested by Prettier

skirtles-code:
A further follow up on #1009.

These are spacing changes suggested by Prettier. These shouldn't change the rendered output, at least not visually.

It seems `coc/index.md` also uses `CRLF` line endings instead of `LF`, a bit like #1027. I hadn't intended that change to be part of this PR but it seems to have snuck in as those changes weren't apparent in my local diff. I've left it in the PR as it is a worthwhile change to the whitespace, albeit an unintentional one.

For some files the only change is the addition of a blank line at the end of the file. The GitHub diff doesn't always make that entirely clear.

All of these changes should be consistent with what's currently in the project's `.editorconfig`.


NataliaTepluhina (approved):

Alex-Sokolov:
Maybe turn on in `.editorconfig` btw?

```
[*.md]
trim_trailing_whitespace = false
```

Issue #43, opened 2020-01-29
Migrate Automatic Global Registration to Cookbook


bencodezen:
In https://github.com/vuejs/docs-next/pull/42, there was a discussion that this "### Automatic Global Registration of Base Components" is not needed in this section since it's more of a technique than a fundamental concept to understand.

The proposal is to move it to the Cookbook.

skirtles-code:
This was added to the Cookbook in #1041.

PR #1041, opened 2021-05-08
docs: add automatic global registration to cookbook

vilaboim:
## Description of Problem

In #43 there is a proposal by @bencodezen to move "### Automatic Global Registration of Base Components" to Cookbook.

## Proposed Solution

I updated the Vue 2 example and added a page inside Cookbook.

## Additional Information
Related to #565

### DIFF
--- a/src/cookbook/automatic-global-registration-of-base-components.md
+++ b/src/cookbook/automatic-global-registration-of-base-components.md
@@ -1,0 +1,75 @@
+# Automatic Global Registration of Base Components
+
+## Base Example
+
+Many of your components will be relatively generic, possibly only wrapping an element like an input or a button. We sometimes refer to these as [base components](../style-guide/#Base-component-names-strongly-recommended) and they tend to be used very frequently across your components.
+
+The result is that many components may include long lists of base components:
+
+```js
+import BaseButton from './BaseButton.vue'
+import BaseIcon from './BaseIcon.vue'
+import BaseInput from './BaseInput.vue'
+export default {
+  components: {
+    BaseButton,
+    BaseIcon,
+    BaseInput
+  }
+}
+```
+
+Just to support relatively little markup in a template:
+
+```html
+<BaseInput v-model="searchText" v-on:keydown.enter="search" />
+<BaseButton v-on:click="search">
+  <BaseIcon name="search" />
+</BaseButton>
+```
+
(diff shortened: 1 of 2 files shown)


src/cookbook/automatic-global-registration-of-base-components.md
```diff
@@ -0,0 +1,75 @@
...
+Just to support relatively little markup in a template:
+
+```html
+<BaseInput v-model="searchText" v-on:keydown.enter="search" />
+<BaseButton v-on:click="search">
+  <BaseIcon name="search" />
+</BaseButton>
+```
+
+Fortunately, if you're using Webpack (or [Vue CLI](https://github.com/vuejs/vue-cli), which uses Webpack internally), you can use `require.context` to globally register only these very common base components. Here's an example of the code you might use to globally import base components in your app's entry file (e.g. `src/main.js`):
```
NataliaTepluhina, line 31: **suggestion:**  I think we should use `webpack` lowercased everywhere as it's [officially lowercased](https://twitter.com/wsokra/status/855800490713649152?lang=en). I know we should fix this across all the docs but for now let's not add more capitalized Webpack to the codebase 😄

NataliaTepluhina (commented):
@vilaboim great addition to the cookbook! I've left a few nitpicky comments here and there, rest looks good to me :+1:

src/cookbook/automatic-global-registration-of-base-components.md
```diff
@@ -0,0 +1,75 @@
+# Automatic Global Registration of Base Components
+
+## Base Example
+
+Many of your components will be relatively generic, possibly only wrapping an element like an input or a button. We sometimes refer to these as [base components](../style-guide/#Base-component-names-strongly-recommended) and they tend to be used very frequently across your components.
```
skirtles-code, line 5:
```suggestion
Many of your components will be relatively generic, possibly only wrapping an element like an input or a button. We sometimes refer to these as [base components](../style-guide/#base-component-names-strongly-recommended) and they tend to be used very frequently across your components.
```

skirtles-code (changes requested):
I've suggested a tweak to the link so that it jumps to the relevant section.

src/cookbook/automatic-global-registration-of-base-components.md
vilaboim: I hadn't realized that. 😄

src/cookbook/automatic-global-registration-of-base-components.md
vilaboim: Thank you @skirtles-code, great observation.

PR #1045, opened 2021-05-09
docs: Update key-attribute.md

Alex-Sokolov:
## Description of Problem
Fix example, because now there is a warning about duplicated keys

### DIFF
--- a/src/guide/migration/key-attribute.md
+++ b/src/guide/migration/key-attribute.md
@@ -79,8 +79,8 @@ Similarly, when using `<template v-for>` with a child that uses `v-if`, the `key
 ```html
 <!-- Vue 2.x -->
 <template v-for="item in list">
-  <div v-if="item.isVisible" :key="'heading-' + item.id">...</div>
-  <span v-else :key="'content-' + item.id">...</span>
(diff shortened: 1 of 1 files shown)


src/guide/migration/key-attribute.md
```diff
@@ -79,8 +79,8 @@ Similarly, when using `<template v-for>` with a child that uses `v-if`, the `key
 ```html
 <!-- Vue 2.x -->
 <template v-for="item in list">
-  <div v-if="item.isVisible" :key="item.id">...</div>
-  <span v-else :key="item.id">...</span>
+  <div v-if="item.isVisible" :key="'heading-' + item.id">...</div>
+  <span v-else :key="'content-' + item.id">...</span>
```
skirtles-code, line 83: I'm not sure about changing this second example. Using a 'duplicate' key with `v-if`/`v-else` was allowed in Vue 2 and I think this example is easier to follow without the string concatenation.

skirtles-code (changes requested):
The changes to the first example look good to me but I'm not sure about the second example. Unless I'm missing something I think that second change should be reverted. Thanks.

Alex-Sokolov:
@skirtles-code revert second one

Issue #1018, opened 2021-04-26
[Improvement] Not all lifecycle hooks mentioned


TheDutchCoder:
There are currently a couple of different places that talk about lifecycle hooks:
1. [API Reference](https://v3.vuejs.org/api/options-lifecycle-hooks.html)
2. [Guide](https://v3.vuejs.org/guide/instance.html#lifecycle-hooks)
3. [Composition API Guide](https://v3.vuejs.org/guide/composition-api-lifecycle-hooks.html#lifecycle-hooks)

Only the API Reference one lists hooks like `onActivated`, but it's missing from the other ones.

A case could be made that it doesn't belong to the Composition guide one (although it really couldn't hurt)

I think it would be a good idea to mention all the hooks in those places with a link to the API Reference.

If this sounds like a good idea, I can create a PR for it, I just don't think I have time to update the lifecycle diagram (if required).

NataliaTepluhina:
@TheDutchCoder thank you for the issue! `activated`/`deactivated` are not the hooks that happen for every single component - they only appear for kept-alive components. I believe for this reason we don't want to see them in the lifecycle diagram, but surely we could mention them in the Composition API guide/ Would be happy to see a PR!

skirtles-code:
The `activated` and `deactivated` hooks are currently mentioned in 3 places that I can find:

- https://v3.vuejs.org/api/options-lifecycle-hooks.html#activated
- https://v3.vuejs.org/api/built-in-components.html#keep-alive
- https://v3.vuejs.org/style-guide/#component-instance-options-order-recommended

I'm not sure whether it's a conscious decision to omit them from the main guide.

The composition API functions `onActivated` and `onDeactivated` don't seem to be documented at all. At the very least I think they should be added to the API reference. The lifecycle hooks for the options API and composition API are distinct and both need to be in the API reference.

NataliaTepluhina:
Closing an issue as the relevant MR was merged

PR #1051, opened 2021-05-14
Add onActivated and onDeacitvated hooks to the list

connectkushal:
## Description of Problem
All hooks were not mentioned in the list of lifecycle hooks in the new docs, also mentioned in #1018

## Proposed Solution
Updated the docs


NataliaTepluhina (approved):
@connectkushal thank you for the contribution 👍🏻

PR #1053, opened 2021-05-15
docs: update recommended node version

naokie:
## Description of Problem

Node.js 10 is now EOL
https://nodejs.org/en/about/releases/

## Proposed Solution

Update the recommended minimum version


kazupon:
@NataliaTepluhina 
addition (ref): https://github.com/vuejs-jp/ja.vuejs.org/pull/322#discussion_r631870388
We realized this when we were reviewing in translation for Japanese

sdras:
Thank you for catching this!

PR #1070, opened 2021-05-25
Fix: broken gif url

VadymGulka:
## Description of Problem
Time-travel demo gif URL points to nonexistent branch
## Proposed Solution
Change URL to legacy branch


skirtles-code:
Thanks for the fix!

PR #1071, opened 2021-05-26
fix: json format quotation

1sm23:


skirtles-code:
Thanks for the PR!

I'm not sure whether it's better to have the language as JSON or JS, I can see an argument either way. Whichever we go with the code should match, so I'm going to merge this for now and we can always change it if we decide to go the other way.

PR #1075, opened 2021-05-29
Update keep-alive hook documentation

tylermercer:
## Description of Problem

Previously, the docs did not make clear that the `activated` hook is called for the entire component tree that is kept alive, and that `mounted` and `unmounted` are not called anywhere in that tree.

## Proposed Solution

Update the docs.


NataliaTepluhina (approved):
@tylermercer thanks, it definitely looks more clear now 👍🏻

PR #1076, opened 2021-06-02
Fix typo in render-function-api.md

mrts:


NataliaTepluhina (approved):
@mrts thanks for the fix!

PR #1078, opened 2021-06-03
3.1 Updates

yyx990803:
Hold for release.

### DIFF
--- a/src/api/application-config.md
+++ b/src/api/application-config.md
@@ -176,8 +176,6 @@ Setting the value to `'preserve'` will disable (2) and (3).
 
 - **Default:** `{{ "['\u007b\u007b', '\u007d\u007d']" }}`
 
-- **Restrictions:** This option is only available in the full build, with in-browser template compilation.
-
 - **Usage:**
 
 ```js
--- a/src/api/application-config.md
+++ b/src/api/application-config.md
@@ -77,27 +77,6 @@ const app = createApp({})
 app.config.globalProperties.$http = () => {}
 ```
 
-## isCustomElement
-
-- **Type:** `(tag: string) => boolean`
-
-- **Default:** `undefined`
-
-- **Usage:**
-
-```js
-// any element starting with 'ion-' will be recognized as a custom one
-app.config.isCustomElement = tag => tag.startsWith('ion-')
-```
-
-Specifies a method to recognize custom elements defined outside of Vue (e.g., using the Web Components APIs). If component matches this condition, it won't need local or global registration and Vue won't throw a warning about an `Unknown custom element`.
-
-> Note that all native HTML and SVG tags don't need to be matched in this function - Vue parser performs this check automatically
-
-::: tip Important
-This config option is only respected when using the runtime compiler. If you are using the runtime-only build, `isCustomElement` must be passed to `@vue/compiler-dom` in the build setup instead - for example, via the [`compilerOptions` option in vue-loader](https://vue-loader.vuejs.org/options.html#compileroptions).
-:::
-
 ## optionMergeStrategies
 
 - **Type:** `{ [key: string]: Function }`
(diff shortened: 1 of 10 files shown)


src/api/application-config.md
```diff
@@ -139,3 +118,93 @@ The merge strategy receives the value of that option defined on the parent and c
...
+- **Usage:**
+
+```js
+app.config.compilerOptions.whitespace = 'preserve'
+```
+
+By default, Vue removes/condenses whitespaces between template elements to produce more efficient compiled output:
+
+1. Leading / ending whitespaces inside an element are condensed into a single space
+2. Whitespaces between elements that contains newlines are removed
```
skirtles-code, line 168:
```suggestion
2. Whitespaces between elements that contain newlines are removed
```

Personally, I don't use the word 'whitespaces' in the plural but perhaps that's just me. I've retained the plural here and switched the conjugation of the verb to match.

I think this sentence is slightly ambiguous. On first reading I thought 'that contains newlines' referred to the elements, rather than to the whitespace.

src/api/application-config.md
```diff
@@ -139,3 +118,93 @@ The merge strategy receives the value of that option defined on the parent and c
...
+3. Consecutive whitepsaces in text nodes are condensed into a single space
```
skirtles-code, line 169:
```suggestion
3. Consecutive whitespaces in text nodes are condensed into a single space
```

src/api/application-config.md
```diff
@@ -139,3 +118,93 @@ The merge strategy receives the value of that option defined on the parent and c
...
+
+Setting the value to `'preserve'` will disalbe (2) and (3).
+
+### compilerOptions.delimiters
+
+- **Type:** `Array<string>`
+
+- **Default:** `{{ "['\u007b\u007b', '\u007d\u007d']" }}`
+
+- **Restrictions:** This option is only available in the full build, with in-browser template compilation.
```
skirtles-code, line 179: Should we remove this line as this restriction applies to all of these settings?

src/api/special-attributes.md
```diff
@@ -56,15 +56,27 @@
...
+  <component :is="currentView"></component>
+  ```
+
+- **Usage on native elements** <Badge text="3.1+" />
 
-```html
-<!-- component changes when currentView changes -->
-<component :is="currentView"></component>
-```
+  When the `is` attribute is used on native HTML elements, it will be interpreted as [Customized built-in elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-customized-builtin-example) which is a native web platform feature.
```
skirtles-code, line 70:
```suggestion
  When the `is` attribute is used on a native HTML element, it will be interpreted as a [Customized built-in element](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-customized-builtin-example), which is a native web platform feature.
```

src/guide/component-basics.md
```diff
@@ -406,6 +406,18 @@ That's all you need to know about dynamic components for now, but once you've fi
 
 ## DOM Template Parsing Caveats
 
+If you are writing your Vue templates directly in the DOM, Vue will have to retrieve the template string from the DOM. This leads to some caveats due to browser's native HTML parsing behavior.
```
skirtles-code, line 409:
```suggestion
If you are writing your Vue templates directly in the DOM, Vue will have to retrieve the template string from the DOM. This leads to some caveats due to browsers' native HTML parsing behavior.
```

src/guide/component-basics.md
```diff
@@ -406,6 +406,18 @@ That's all you need to know about dynamic components for now, but once you've fi
...
+
+:::tip
+It should be noted that the limitations discussed below only applies if you are writing your templates directly in the DOM. They do NOT apply if you are using string templates from the following sources:
```
skirtles-code, line 412:
```suggestion
It should be noted that the limitations discussed below only apply if you are writing your templates directly in the DOM. They do NOT apply if you are using string templates from the following sources:
```

src/guide/component-basics.md
```diff
@@ -416,28 +428,21 @@ This will lead to issues when using components with elements that have such rest
...
-<!-- Correct -->
-<tr v-is="'blog-post-row'"></tr>
-```
-
+:::tip
+When used on native HTML elements, the value of `is` must be prefixed with `vue:` in order to be interpreted as a Vue component. This is required to avoid confusion with native [customized built-in elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-customized-builtin-example).
 :::
 
-Also, HTML attribute names are case-insensitive, so browsers will interpret any uppercase characters as lowercase. That means when you’re using in-DOM templates, camelCased prop names and event handler parameters need to use their kebab-cased (hyphen-delimited) equivalents:
+### Case Insensivitity
```
skirtles-code, line 443:
```suggestion
### Case Insensitivity
```

skirtles-code (changes requested):
Some suggested tweaks from me below.

In addition, I think there are some other pages that will also need updating:

* `/guide/migration/custom-elements-interop.md` uses `v-is` and mentions `app.config.isCustomElement`.
* `/guide/migration/global-api.md` mentions `app.config.isCustomElement`.
* `/style-guide/README.md` mentions `delimiters` and should probably be changed to `compilerOptions`.

yyx990803:
@skirtles-code thanks for the review! Should all be addressed now.

PR #1088, opened 2021-06-12
fix: await router push in ssr server

NorbertNader:
## Description of Problem

When not awaiting route push I get an error in the console "Hydration completed but contains mismatches" and the routes do not behave as expected i.e. when navigating to a different page using router-link the content of router-view does not get rendered.

## Proposed Solution

await router push

## Additional Information

N/A


NataliaTepluhina (approved):
@NorbertNader thanks for the addition!

PR #1095, opened 2021-06-18
Update Laravel logo

driesvints:
I'm not sure who sent it in but the current Laravel logo has white space around it which makes it rather small on the docs. I've attached a version of our logo without the whitespace.

Screen Shot 2021-06-18 at 18 31 02


NataliaTepluhina (approved):
@driesvints thank you for the fix!

PR #1096, opened 2021-06-18
Correct vue.config.js example

lindsaykwardell:
A bracket is missing from the vue.config.js configuration example, this change would add in the missing bracket so the code example is valid.

## Description of Problem

In the code example for vue.config.js to add the compat rule, a bracket is missing which results in an error when copy/pasted.

## Proposed Solution

The proposed solution would be to add in the bracket, so that the code example is correct and valid.

## Additional Information

### DIFF
--- a/src/guide/migration/migration-build.md
+++ b/src/guide/migration/migration-build.md
@@ -91,6 +91,7 @@ The following workflow walks through the steps of migrating an actual Vue 2 app 
              }
            }
          })
+     }
    }
    ```
 


skirtles-code:
Looks good. Thanks for the fix.

catrope:
I just attempted to submit this same fix before finding that this had already been fixed ~6 weeks ago. Apparently the [deployed version](https://v3.vuejs.org/guide/migration/migration-build.html#installation) has not been updated in that time, because it still has this syntax error.

skirtles-code:
@catrope The docs deploy automatically when changes are merged. Perhaps you're seeing a cached version? The 'Last updated' time at the bottom of that page should read '2021-07-13, 16:56:51 UTC'.

catrope:
I guess I must have seen a cached version somehow. I just opened the same page again and it looks right this time.

PR #1104, opened 2021-06-22
chore: remove axios

Xenonym:
## Description of Problem
axios was added in #31, but is now unused as code examples were moved to CodePen in #57.

## Proposed Solution
Let's remove axios.


skirtles-code (approved):
Looks good. :+1:

Issue #1080, opened 2021-06-06
Suggest: Vue3.0 document cross-domain switching Chinese and English translation routing problem


NaCoLiu:
Suggest: Vue3.0 document cross-domain switching Chinese and English translation routing problem.
It is recommended to modify the translation items of the file system. Obtain the current route and directly point to the route to the Chinese station.
https://github.com/vuejs/docs-next-zh-cn/issues/549

NaCoLiu:
zh-CN dosc website developers are ready

skirtles-code:
I think this is a duplicate of #932.

I don't think this is feasible as new pages get added to the docs all the time and the translations are invariably behind. If you have a suggestion for how to solve that problem please let us know.

Jinjiang:
Hi, @skirtles-code I understand your concern, which is not a technical one. However, in v2 we already implemented that through [this](https://github.com/vuejs/vuejs.org/blob/master/themes/vue/layout/partials/language_dropdown.ejs). I'm afraid of users' side, they feel the experience going backward. Then what about we just take this concern as the v2 is right now and keep the experience the same? :-)
Thanks.

NaCoLiu:
Only add some code, no need recode project, this suggestion will be some dev, thx

skirtles-code:
This is now implemented. See #1105 for details. It includes some improved handling for 404 errors, which will only be fully effective once it is pulled into the translations.

Issue #932, opened 2021-03-07
feature wanted: auto open translation docs with the current url pathname


xgqfrms:
# feature wanted

> auto open translation docs with the current url pathname

## current

> click the translation zh-Hans link, just goto the home page

https://v3.vuejs.org/guide/composition-api-setup.html#props

`=>`

https://v3.cn.vuejs.org/

## wanted

> click the translation zh-Hans link, goto the page with the current url pathname

https://v3.vuejs.org/guide/composition-api-setup.html#props

`=>`

https://v3.cn.vuejs.org/guide/composition-api-setup.html#props

## try failed

https://github.com/vuejs/docs-next/blob/master/src/.vuepress/config.js#L418

```js
          {
            text: '中文',
            link: 'https://v3.cn.vuejs.org/'
          }
```

```js
          {
            text: '中文',
            link: `https://v3.cn.vuejs.org${window?.location?.pathname}`,
          }
```

skirtles-code:
I understand why this would be desirable but I'm not sure whether it would be practical. New pages are added to the documentation all the time and there is inevitably a delay before those pages are added to the translations.

skirtles-code:
I'm going to close this for now. We can reopen it if someone can suggest a viable strategy for handling new pages that aren't yet in the translations.

PR #1105, opened 2021-06-22
Link directly to equivalent pages in translations

skirtles-code:
@Jinjiang @veaba @kazupon @naokie @narusas @ChangJoo-Park @Alex-Sokolov This will impact the translations, so could I get any feedback you might have?

Related issues: #1080, #932.

In the Vue 2 documentation, links to the translations have the same path as the current page. This PR attempts to recreate that for Vue 3.

The translations will never be in perfect sync with this repo, so when pages are added, moved or removed they will hit a 404. My proposed strategy for handling that is:

* Remove the `noreferrer` attribute from these links so that the 404 page can tell where the user came from.
* If the user came from an alternative translation, show a longer message on the 404 page that attempts to explain why that might have happened.

The 404 page previously showed one of 4 different messages. That doesn't work well with SSR, so I've removed that functionality. My proposed English version would now look like this:

404 page

The paragraph about translations is only shown if `document.referrer` matches a known translation.

I've changed the link for the Russian translation to include the `/ru` in the path. That is required to link to the correct pages.

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -423,23 +423,28 @@ module.exports = {
           // Translation maintainers: Please include the link below to the English documentation
           // {
           //   text: 'English',
-          //   link: 'https://v3.vuejs.org/'
+          //   link: 'https://v3.vuejs.org/',
+          //   isTranslation: true
           // },
           {
             text: '中文',
-            link: 'https://v3.cn.vuejs.org/'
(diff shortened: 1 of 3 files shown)


skirtles-code:
The reactions I've had from the translators seem positive, so I'm going to move this PR out of draft, ready to merge. Thanks all.

NataliaTepluhina (approved):

PR #1113, opened 2021-06-29
fix typo in composition-api-template-refs.md

gexin1:
## Description of Problem
when using the `flush: 'post'` option,  print error 
```javascript
 //<div></div>
```
## Proposed Solution
 modify
```javascript
 //<div>This is a root element</div>
```


NataliaTepluhina (approved):
@gexin1 thanks for the fix!

PR #1114, opened 2021-06-29
Add TypeScript example for event handling

### LINKED ISSUE #927: Add TypeScript example for event handling
Since a lot of developers will be dealing with event handling in their apps, it would be a good idea to add some documentation around this when doing event handling in combination with TypeScript.

To give an example, consider the following snippets:

```html
<template>
  <input type="text" @change="handleChange" />
</template>

<script lang="ts">
import { defineComponent } from 'vue'

export default defineComponent({
  setup () {
    const handleChange = (evt) => { // `evt` will be of type `any`
      console.log(evt.target.value) // TS will throw an error here
    }

    return { handleChange }
  }
})
</script>
```

Here `evt.target.value` will throw a TypeScript error, because the event passed by the Vue template isn't typed.
Strictly speaking this might not be Vue's job, since it's a native DOM event, but developers might wonder what's going on.

The solution, currently, is to cast the event target like so:
```ts
    const handleChange = (evt: Event) => {
      console.log((evt.target as HTMLInputElement).value)
    }
```

I think it would be beneficial to add that to the docs somewhere.

NataliaTepluhina:
Close #927

Props to @TheDutchCoder for the examples!

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -486,3 +486,36 @@ export default defineComponent({
   }
 })
 ```
+
+### Typing Event Handlers
+
+When dealing with native DOM events, it might be useful to type the argument we pass to the handler correctly. Let's take a look at this example:
+
+```vue
+<template>
+  <input type="text" @change="handleChange" />
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+
+export default defineComponent({
+  setup() {
+    // `evt` will be of type `any`
+    const handleChange = evt => {
+      console.log(evt.target.value) // TS will throw an error here
+    }
+
+    return { handleChange }
+  }
+})
+</script>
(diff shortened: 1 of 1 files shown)


src/guide/typescript-support.md
```diff
@@ -486,3 +486,36 @@ export default defineComponent({
...
+
+    return { handleChange }
+  }
+})
+</script>
+```
+
+As you can see, without annotating the `evt` argument correctly, TypeScript will throw an error when we try to access the value of the `<input>` element. The solution is to cast the event target with a correct type:
+
+```js
```
skirtles-code, line 517:
````suggestion
```ts
````

skirtles-code (approved):
One small suggestion on the snippet language but otherwise it all looks fine to me.

PR #1121, opened 2021-07-02
Improve Example for TypeScript Property Annotation

gamperl:
## Description of Problem
In the `Annotating Props` section of the chapter `Typescript Support`, it is not clear how to annotate a property with any or a union type. 

## Proposed Solution
Extend the example with a union type property and an any property.


NataliaTepluhina (approved):
@gamperl thank you for the addition!

Issue #727, opened 2020-12-03
cookbook idea: developing in Docker


eikooc:
Hi. I'd like a cookbook guide to show how to develop and serve a vuejs application inside Docker.

skirtles-code:
The Vue CLI documentation includes a section about deploying with Docker:

https://cli.vuejs.org/guide/deployment.html#docker-nginx

Our current feeling within the docs team is that there wouldn't be enough Vue-specific content to justify having a separate entry in the Vue 3 Cookbook.

PR #1127, opened 2021-07-05
Add new article "Dockerizing Vue App" to the cookbook

standbyoneself:
## Description of Problem

This article helps beginner developers be familiar with Docker

## Proposed Solution

The article

## Additional Information

#727


skirtles-code:
@standbyoneself Sorry for the delay getting back to you and thank you for all your work on this.

We've been having some internal discussions within the docs team about the cookbook and what form it should take going forward. We'll try to get back to you as soon as possible once we've figured out where we want to take this.

skirtles-code:
Some references for anyone else looking at this:

* [Vue 2 cookbook Docker guide](https://vuejs.org/v2/cookbook/dockerize-vuejs-app.html)
* [Vue CLI Docker deployment guide](https://cli.vuejs.org/guide/deployment.html#docker-nginx)

skirtles-code:
We've decided that we'd like to take the Vue 3 Cookbook in a slightly different direction from the Vue 2 Cookbook. That is going to impact the migration of most of the Vue 2 recipes in some way and several of them are not going to be migrated because they're no longer a good fit for what we're hoping to create.

Going forward, we're looking to use the Cookbook as a place to store Vue best practices and solutions to common problems that we observe in the community. These will typically be topics that don't fit neatly into the guides because they straddle multiple core Vue concepts.

In the case of using Docker, we feel there isn't sufficient Vue content to justify having a dedicated Cookbook entry for this topic in the core Vue documentation. The [Deployment Guide](https://cli.vuejs.org/guide/deployment.html#docker-nginx) for Vue CLI covers some of the same material and seems like the right place for that content.

I understand this may come as a disappointment given the time you've put into creating a Cookbook entry. I am sorry about that. The effort you put in is very much appreciated.

PR #1129, opened 2021-07-07
Update the API entries for mixins and extends to reference setup

skirtles-code:
I've updated the API reference entries for `mixins` and `extends`:

* Mention the composition API in the entry for `mixins`.
* Reword the entry for `extends` to better reflect Vue 3. Using a function/constructor is no longer a thing and `Vue.extend` doesn't exist anymore.
* Add an explanation of how to combine `extends` and `setup`. This is based on a couple of conversations on Discord where 'don't use `extends`' didn't seem like a legitimate suggestion.

Marking as a draft for now as I want to check that last part isn't too controversial.

### DIFF
--- a/src/api/options-composition.md
+++ b/src/api/options-composition.md
@@ -10,11 +10,15 @@
 
   Mixin hooks are called in the order they are provided, and called before the component's own hooks.
 
+  :::info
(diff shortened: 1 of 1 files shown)


skirtles-code:
Closing this.

The less controversial parts are now in #1146. I'll rethink the part about combining `extends` and `setup`

PR #1135, opened 2021-07-12
fix: broken redirect url

fazaio:
## Description of Problem
link native-vue in mobile.md page is redirect to broken link "https://nativescript.org/vue/"
## Proposed Solutionf
fix redirect link to "https://nativescript-vue.org/"
## Additional Information
https://v3.vuejs.org/guide/mobile.html#hybrid-app-development

### DIFF
--- a/src/guide/mobile.md
+++ b/src/guide/mobile.md
@@ -20,4 +20,4 @@ While Vue.js does not natively support mobile app development, there are a numbe
 
 **Resources**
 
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Thanks for the fix!

According to the Wayback Machine, this is what was linked previously:

https://web.archive.org/web/20201109025815/https://nativescript.org/vue/

It isn't quite the same as the updated link but it does seem to be the closest we can get currently.

I would also note that `nativescript-vue` seems to be using Vue 2.

PR #1137, opened 2021-07-13
A CMS-powered blog with Vue.js and ButterCMS

orlyohreally:
A cookbook recipe on how to quickly build a CMS-powered blog with Vue.js and ButterCMS


skirtles-code:
@orlyohreally Sorry for the delay getting back to you and thank you for all your work on this.

We've been having some internal discussions within the docs team about the cookbook and what form it should take going forward. We'll try to get back to you as soon as possible once we've figured out where we want to take this.

skirtles-code:
We've decided that we'd like to take the Vue 3 Cookbook in a slightly different direction from the Vue 2 Cookbook. That is going to impact most of the Vue 2 recipes in some way and several of them are not going to be migrated because they're no longer a good fit for what we're hoping to create.

Going forward, we're looking to use the Cookbook as a place to store best practices and solutions to common problems that we observe in the community. These will typically be topics that don't fit neatly into the guides because they straddle multiple core concepts.

In the case of this PR, we feel there's some good Vue content included and it would be good to have some of that covered within the Cookbook. However, the emphasis should be on teaching people about Vue and explaining those key concepts. I think what we'd want is Cookbook entries that use whatever examples are best to explain the concepts, rather than starting from the example. We also had concerns about a Cookbook entry that requires readers to sign up for a service in order to run the example.

The effort you put into creating this is very much appreciated and I'm sorry we didn't feel able to include it. We hope to be able to reuse some of the ideas from it in Cookbook entries going forward.

orlyohreally:
@skirtles-code, could you explain what you expect to be in the recipe a little bit more in terms of our recipe. Should we explain the logic behind the steps in more detail rather than jumping to examples?

> We also had concerns about a Cookbook entry that requires readers to sign up for a service in order to run the example.

We can use an actual authToken so the readers are not required to register if they don't want to. We also have a free community developer plan available, so they can test it out for free.

PR #1138, opened 2021-07-13
chore: update installation instructions for Vite

docyx:
`@vitejs/app` is now [deprecated](https://github.com/vitejs/vite/pull/4179). We should instruct users to use the new `create-vite` package instead.

Also, I got rid of the mention of `create-vite-app` since that package is deprecated and says to use `create-vite` instead.


skirtles-code (approved):
Thank you.

PR #1146, opened 2021-07-20
Update the API entries for mixins and extends

skirtles-code:
I've updated the API reference entries for `mixins` and `extends`:

* Mention the composition API in the entry for `mixins`.
* Reword the entry for `extends` to better reflect Vue 3. Using a function/constructor is no longer a thing and `Vue.extend` doesn't exist anymore.

This PR replaces #1129.

### DIFF
--- a/src/api/options-composition.md
+++ b/src/api/options-composition.md
@@ -10,11 +10,15 @@
 
   Mixin hooks are called in the order they are provided, and called before the component's own hooks.
 
+  :::info
+  In Vue 2, mixins were the primary mechanism for creating reusable chunks of component logic. While mixins continue to be supported in Vue 3, the [composition API](/guide/composition-api-introduction.html) is now the preferred approach for code reuse between components.
+  :::
+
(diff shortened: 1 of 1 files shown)


src/api/options-composition.md
```diff
@@ -10,11 +10,15 @@
 
   Mixin hooks are called in the order they are provided, and called before the component's own hooks.
 
+  :::info
+  In Vue 2, mixins were the primary mechanism for creating reusable chunks of component logic. While mixins continue to be supported in Vue 3, the [composition API](/guide/composition-api-introduction.html) is now the preferred approach for code reuse between components.
```
bencodezen, line 14:
```suggestion
  In Vue 2, mixins were the primary mechanism for creating reusable chunks of component logic. While mixins continue to be supported in Vue 3, the [Composition API](/guide/composition-api-introduction.html) is now the preferred approach for code reuse between components.
```

bencodezen (approved):

PR #1148, opened 2021-07-23
Allow users to collapse sidebar on larger devices

torressam333:
## Description of Problem
When reading documentation on larger devices (laptops and larger) I did not find a way to collapse the side bar. The reason I prefer a collapsed sidebar when reading documentation is that it serves as a minor distraction. I wanted a way to collapse it on my larger devices while reading and then expanding the sidebar when I needed it again.

## Proposed Solution
I added a menu icon on the top `Navbar` that is shown when the screen size is on a larger device. (1044px) or greater. This feature includes a Vue component and necessary changes in order for the collapse/expand functionality to work properly.

##Additional Information
I wanted it to be a good solution that also blended well with the existing style layout of the page.


skirtles-code:
Thanks for your work on this and sorry for the delay getting back to you.

Usually a change like this would be integrated into the platform (i.e. VuePress) so we don't have problems maintaining it when we need to upgrade to the latest version. However...

We're currently in the process of migrating the docs from VuePress to VitePress (work in progress is on the `next` branch). Hopefully this won't take too much longer but there isn't a specific timeline in place for getting the migration completed.

Any custom functionality we added to the docs now would be lost during the migration so I think we should put this on hold for now. I'll leave it open for the time being so it doesn't get forgotten.

torressam333:
@skirtles-code sounds good! Thank you very much for getting back to me :). Ill keep my eyes open for any updates on this in the future.

bencodezen:
Hey @torressam333! Thanks again for your work on this. With the new migration to VitePress, I think we should go ahead and close the PR for now. Appreciate it and looking forward to your future contributions!

torressam333:
Thanks @bencodezen! You're a vue legend. Have a nice day.

PR #1149, opened 2021-07-24
feat: added cursor pointer to ensure consistency

akulsr0:
Changed the subscribe button cursor to pointer as all buttons on the page have pointer cursor to make it consistent.

### Before
old_behaviour

### After
new_behaviour


NataliaTepluhina (approved):
@akulsr0 thank you! UX is much better with `pointer` indeed 👍🏻

akulsr0:
@NataliaTepluhina I have also raised PR for the same change in Vue 2 docs.
Here is the link: https://github.com/vuejs/vuejs.org/pull/2848

PR #1157, opened 2021-08-03
3.2 updates

yyx990803:
- Added dedicated API pages for SFC spec, tooling & features
- Added "Vue and Web Components" section
- various 3.2 features

### DIFF
--- a/src/api/directives.md
+++ b/src/api/directives.md
@@ -237,7 +237,7 @@
 
 ## v-bind
 
-- **Shorthand:** `:`
+- **Shorthand:** `:` or `.` (when using `.prop` modifier)
 
 - **Expects:** `any (with argument) | Object (without argument)`
 
@@ -246,6 +246,8 @@
 - **Modifiers:**
 
   - `.camel` - transform the kebab-case attribute name into camelCase.
+  - `.prop` - force a binding to be set as DOM property. <Badge text="3.2+"/>
+  - `.attr` - force a binding to be set as DOM attribute. <Badge text="3.2+"/>
 
 - **Usage:**
 
@@ -278,23 +280,34 @@
   <!-- class binding -->
   <div :class="{ red: isRed }"></div>
   <div :class="[classA, classB]"></div>
-  <div :class="[classA, { classB: isB, classC: isC }]">
-    <!-- style binding -->
-    <div :style="{ fontSize: size + 'px' }"></div>
-    <div :style="[styleObjectA, styleObjectB]"></div>
+  <div :class="[classA, { classB: isB, classC: isC }]"></div>
 
-    <!-- binding an object of attributes -->
-    <div v-bind="{ id: someProp, 'other-attr': otherProp }"></div>
+  <!-- style binding -->
+  <div :style="{ fontSize: size + 'px' }"></div>
+  <div :style="[styleObjectA, styleObjectB]"></div>
 
-    <!-- prop binding. "prop" must be declared in my-component. -->
-    <my-component :prop="someThing"></my-component>
+  <!-- binding an object of attributes -->
+  <div v-bind="{ id: someProp, 'other-attr': otherProp }"></div>
 
-    <!-- pass down parent props in common with a child component -->
-    <child-component v-bind="$props"></child-component>
+  <!-- prop binding. "prop" must be declared in my-component. -->
+  <my-component :prop="someThing"></my-component>
 
-    <!-- XLink -->
-    <svg><a :xlink:special="foo"></a></svg>
-  </div>
+  <!-- pass down parent props in common with a child component -->
+  <child-component v-bind="$props"></child-component>
+
+  <!-- XLink -->
+  <svg><a :xlink:special="foo"></a></svg>
+  ```
+
+  When setting a binding on an element, Vue by default checks whether the element has the key defined as a property using an `in` operator check. If the property is defined, Vue will set the value as DOM property instead of an attribute. This should work in most cases, but you can override this behavior by explicitly using `.prop` or `.attr` modifiers. This is sometimes necessary especially when [working with custom elements](/guide/web-components.html#passing-dom-properties).
+
+  The `.prop` modifier also has a dedicated shorthand, `.`:
+
+  ```html
+  <div :someProperty.prop="someObject"></div>
+
+  <!-- equivalent to -->
+  <div .someProperty="someObject"></div>
   ```
 
   The `.camel` modifier allows camelizing a `v-bind` attribute name when using in-DOM templates, e.g. the SVG `viewBox` attribute:
@@ -475,11 +488,7 @@
   This directive is provided solely for micro optimizations in performance-critical scenarios and should be rarely needed. The most common case where this may prove helpful is when rendering large `v-for` lists (where `length > 1000`):
 
   ```html
-  <div
-    v-for="item in list"
-    :key="itme.id"
-    v-memo="[item.id === selected]"
-  >
+  <div v-for="item in list" :key="itme.id" v-memo="[item.id === selected]">
     <p>ID: {{ id }} - selected: {{ item.id === selected }}</p>
     <p>...more child nodes</p>
   </div>
--- a/src/api/directives.md
+++ b/src/api/directives.md
@@ -246,8 +246,8 @@
 - **Modifiers:**
 
   - `.camel` - transform the kebab-case attribute name into camelCase.
-  - `.prop` - force a binding to be set as DOM property. <Badge text="3.2+"/>
-  - `.attr` - force a binding to be set as DOM attribute. <Badge text="3.2+"/>
+  - `.prop` - force a binding to be set as a DOM property. <Badge text="3.2+"/>
+  - `.attr` - force a binding to be set as a DOM attribute. <Badge text="3.2+"/>
 
 - **Usage:**
 
@@ -299,7 +299,7 @@
   <svg><a :xlink:special="foo"></a></svg>
   ```
 
-  When setting a binding on an element, Vue by default checks whether the element has the key defined as a property using an `in` operator check. If the property is defined, Vue will set the value as DOM property instead of an attribute. This should work in most cases, but you can override this behavior by explicitly using `.prop` or `.attr` modifiers. This is sometimes necessary especially when [working with custom elements](/guide/web-components.html#passing-dom-properties).
+  When setting a binding on an element, Vue by default checks whether the element has the key defined as a property using an `in` operator check. If the property is defined, Vue will set the value as a DOM property instead of an attribute. This should work in most cases, but you can override this behavior by explicitly using `.prop` or `.attr` modifiers. This is sometimes necessary, especially when [working with custom elements](/guide/web-components.html#passing-dom-properties).
 
   The `.prop` modifier also has a dedicated shorthand, `.`:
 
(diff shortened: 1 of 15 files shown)


src/api/sfc-tooling.md
```diff
@@ -0,0 +1,94 @@
...
+
+[WebStorm](https://www.jetbrains.com/webstorm/) also provides decent support for Vue SFCs. However, do note as of now its support for `<script setup>` is still [in progress](https://youtrack.jetbrains.com/issue/WEB-49000).
+
+Most other editors have community-created syntax highlighting support for Vue, but lack the same level of code IntelliSense. In the long run, we do hope we can extend the range of editor support by leveraging the [Language Service Protocol](https://microsoft.github.io/language-server-protocol/) as Volar's core logic is implemented as a standard language server.
+
+## Testing Support
+
+- If using Vite, we recommend [Cypress](https://www.cypress.io/) as the test runner for both unit and e2e tests. Unit tests for Vue SFCs can be done with the [Cypress Component Test Runner](https://www.cypress.io/blog/2021/04/06/introducing-the-cypress-component-test-runner/).
```
afontcu, line 60: 💡 I believe this recommendation is a bit off compared to what docs say in https://vuejs.org/v2/guide/testing.html

src/api/sfc-tooling.md
```diff
@@ -0,0 +1,94 @@
...
+
+- Vue CLI comes with [Jest](https://jestjs.io/) and [Mocha](https://mochajs.org/) integrations.
+
+- If you are manually confugring Jest to work with Vue SFCs, check out [vue-jest](https://github.com/vuejs/vue-jest) which is the official Jest transform for Vue SFCs.
```
afontcu, line 64:
```suggestion
- If you are manually configuring Jest to work with Vue SFCs, check out [vue-jest](https://github.com/vuejs/vue-jest) which is the official Jest transform for Vue SFCs.
```

src/api/sfc-script-setup.md
```diff
@@ -0,0 +1,268 @@
+---
+sidebarDepth: 1
+---
+
+# SFC `<script setup>`
+
+`<script setup>` is a compile-time syntacitc sugar for using [Composition API](https://v3.vuejs.org/api/composition-api.html) inside Single File Components (SFCs). It is the recommended syntax if you are using both SFCs and Composition API. It provides a number of advantages over the normal `<script>` syntax:
```
afontcu, line 7:
```suggestion
`<script setup>` is a compile-time syntactic sugar for using [Composition API](https://v3.vuejs.org/api/composition-api.html) inside Single File Components (SFCs). It is the recommended syntax if you are using both SFCs and Composition API. It provides a number of advantages over the normal `<script>` syntax:
```

src/api/sfc-script-setup.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+`useSlots` and `useAttrs` are actual runtime functions that return the equivalent of `setupContext.slots` and `setupContext.attrs`. They can be used in normal composition API functions as well.
+
+## Usage alongside normal `<script>`
+
+`<script setup>` can be used alongside normal `<script>`. A normal `<script>` maybe needed in cases where we need to:
+
+- Declare options that cannot be expressed in `<script setup>`, for example `inheritAttrs` or custom options enabled via plugins;
+- Declaring named exports;
```
posva, line 183: Does the `;` at the end make sense? Shouldn't it be a `.` or nothing at all in the 3 bullet points?

```suggestion
- Declare options that cannot be expressed in `<script setup>`, for example `inheritAttrs` or custom options enabled via plugins.
- Declaring named exports.
```

src/api/sfc-script-setup.md
```diff
@@ -0,0 +1,268 @@
...
+}>()
+```
+
+- `defineProps` or `defineEmits` can only use either runtime declaration OR type declaration. Using both at the same time will result in a compile error.
+
+- When using type declaration, equivalent runtime declaration is automatically generated from static analysis to remove the need of double declaration and still ensure correct runtime behavior.
+
+  - In dev mode, the compiler will try to infer corresponding runtime validation from the types. For example here `foo: String` is inferred from the `foo: string` type. If the type is a reference to an imported type, the inferred result will be `foo: null` (equal to `any` type) since the compiler does not have information of external files.
+
+  - In prod mode, the compiler will generate the array format declaration to reduce bundle size (the props here will be compiled into `['msg']`)
```
posva, line 239: I didn't know this! 🔥

NataliaTepluhina (commented):
@yyx990803 great work! I've added a bunch of minor nitpicks here and there

src/api/sfc-style.md
```diff
@@ -0,0 +1,193 @@
...
+<style scoped>
+:slotted(div) {
+  color: red;
+}
+</style>
+```
+
+### Global Selectors
+
+If you want just one rule to apply globally, you can use the `:global` pseudo-class:
```
posva, line 79: Maybe it's worth saying 
```suggestion
If you want just one rule to apply globally, you can use the `:global` pseudo-class rather than creating another `<style>` tag without the `scoped` attribute:
```

src/api/sfc-tooling.md
```diff
@@ -0,0 +1,94 @@
...
+## Online Playgrounds
+
+You don't need to install anything on your machine to try out Vue SFCs - there are many online playgrounds that allow you to do so right in the browser:
+
+- [Vue SFC Playground](https://sfc.vuejs.org) (official, deployed from latest commit)
+- [VueUse Playground](https://play.vueuse.org)
+- [Vue on CodeSandbox](https://codesandbox.io/s/vue-3)
+- [Vue on Repl.it](https://replit.com/@templates/VueJS-with-Vite)
+- [Vue on Codepen](https://codepen.io/pen/editor/vue)
+- [Vue on StackBlitz](stackblitz.com/fork/vue)
```
posva, line 12:
```suggestion
- [Vue on StackBlitz](https://stackblitz.com/fork/vue)
```

src/api/sfc-tooling.md
```diff
@@ -0,0 +1,94 @@
...
+
+It is also recommend to use these online playgrounds to provide reproductions when reporting bugs.
```
posva, line 14:
```suggestion
It is also recommended to use these online playgrounds to provide reproductions when reporting bugs.
```

src/api/sfc-tooling.md
```diff
@@ -0,0 +1,94 @@
...
+
+- [Docs](https://github.com/vitejs/vite/tree/main/packages/plugin-vue)
+
+Official plugin that provides Vue SFC support in Vite.
+
+### `vue-loader`
+
+- [Docs](https://vue-loader.vuejs.org/)
+
+The official loader that provides Vue SFC support in webpack. If you are using Vue CLI, also see [docs on modifying `vue-loader` options in Vue CLI](https://cli.vuejs.org/guide/webpack.html#modifying-options-of-a-loader).
```
kazupon, line 94: I think it would be better to include 3rd parties using custom blocks, as it would be more helpful for people who are creating ecosystems for some ideas.

As far as I know, it is the following

- [gridsome](https://gridsome.org/): `<page-query>`
- [vite-plguin-vue-gql](https://github.com/wheatjs/vite-plugin-vue-gql): `<gql>`
- [vue-i18n](https://github.com/intlify/bundle-tools): `<i18n>`

src/api/sfc-tooling.md
yyx990803: Yes, this is intentional. Note this is specific to Vite-based build setups because Jest has issues when integrating with pure ESM dev servers, and we've found Cypress to offer good DX for both unit and e2e.

src/api/effect-scope.md
```diff
@@ -0,0 +1,59 @@
+# Effect Scope API <Badge text="3.2+" />
+
+:::info
+Effect scope is an advanced API primarily intended for library authors. For details on how to leverage this API, please consult its corresponding [RFC](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md).
```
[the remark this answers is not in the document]
yyx990803: Maybe. I want to think about where it should be fit into the doc structure though, maybe a section for "advanced reactivity patterns".

src/api/sfc-style.md
```diff
@@ -0,0 +1,193 @@
...
+</template>
+
+<style scoped>
+p {
+  color: v-bind('theme.color');
+}
+</style>
+```
+
+The actual value will be compiled into a hashed CSS custom property so the CSS is still sattic. The custom property and will be applied to component's root element via inline styles and reactively updated if the source value changes.
```
yyx990803, line 193:
```suggestion
The actual value will be compiled into a hashed CSS custom property so the CSS is still sattic. The custom property will be applied to component's root element via inline styles and reactively updated if the source value changes.
```

src/api/sfc-tooling.md
yyx990803: I think this should be added to the SFC spec page

yyx990803: see a4cf142

src/api/directives.md
```diff
@@ -454,6 +467,37 @@
...
+  ```
+
+  When the component re-renders, if both `valueA` and `valueB` remains the same, all updates for this `<div>` and its children will be skipped. In fact, even the Virtual DOM vnode creation will also be skipped since the memoized copy of the sub-tree can be reused.
+
+  It is important to specify the memoization array correctly, otherwise we may skip updates that should indeed be applied.
+
+  This directive is provided solely for micro optimizations in performance-critical scenarios and should be rarely needed. The most common case where this may prove helpful is when rendering large `v-for` lists (where `length > 1000`):
+
+  ```html
+  <div v-for="item in list" :key="itme.id" v-memo="[item.id === selected]">
```
posva, line 491:
```suggestion
  <div v-for="item in list" :key="item.id" v-memo="[item.id === selected]">
```

src/api/sfc-script-setup.md
```diff
@@ -0,0 +1,268 @@
...
+- Better runtime performance (template is compiled into render function in the same scope without an intermediate proxy)
+- Better IDE type-inference performance (less work for the language server to extract types from code)
+
+## Basic Syntax
+
+To opt-in to the syntax, add the `setup` attribute to the `<script>` block:
+
+```html
+<script setup>
+  console.log('hello script setup')
```
ydcjeff, line 20: I noticed all the JS code inside `<script setup>` of SFC `<script setup>` docs are indented 2 spaces while there is no indent in the rest of docs.

Is this something intentional with `<script setup>`?

yyx990803: Good catch, prettier actually treats embedded `vue` and `html` blocks differently and these code blocks should be using `vue` instead.

skirtles-code (commented):
Some suggested tweaks from me. Let me know if anything I've suggested needs clarification.

src/api/effect-scope.md
```diff
@@ -0,0 +1,59 @@
+# Effect Scope API <Badge text="3.2+" />
```
Oleksii14, line 1: Can you some use cases in this section? It is not clear enough in which situations the developers should use it

yyx990803: See https://github.com/vuejs/docs/pull/1157/files#r682602039

The API reference only serves as a reference for the API signatures. Detailed usage will probably be discussed in a dedicated guide section later.

bencodezen:
@yyx990803 Should this PR be targeting `next` rather than `master`?

yyx990803:
@bencodezen this should land on `master` since we are releasing 3.2 before we finish the new design.

PR #1158, opened 2021-08-04
fix: typo 'kept-alive'

naokie:
## Description of Problem

From JP translation team.

This means the [keep-alive](https://v3.vuejs.org/api/built-in-components.html#keep-alive) component.
But, it seems to be spelled incorrectly.


skirtles-code:
I think this may be intentional. A 'kept-alive component' would be the child of a `keep-alive` component, not the `keep-alive` component itself. It could maybe be worded better but I don't think it would be correct to change 'kept-alive' to 'keep-alive'.

NataliaTepluhina:
@skirtles-code is correct, this is intentional

PR #1183, opened 2021-08-16
Update component-attrs.md 

Yukon123:
`data-status` is a attribute rather than property. It should be consistent with the description on line 100:`With this new configuration, our `data-status` attribute will be applied to our `input` element!`


skirtles-code (approved):
Agreed, it's an attribute, not a property.

PR #1184, opened 2021-08-16
Fix: Missing keyword mounted  in composition-api-template-refs.md 

Yukon123:
## Description of Problem
If watchers that use template refs defined with the flush: 'post' option. This will run the effect after the DOM has been `mounted` or updated.

The pre docs miss keyword  called `mounted` 
## Proposed Solution
Add `mounted` into description

### DIFF
--- a/src/guide/composition-api-template-refs.md
+++ b/src/guide/composition-api-template-refs.md
@@ -118,7 +118,7 @@ But a key difference to lifecycle hooks is that `watch()` and `watchEffect()` ef
 </script>
 ```
 
-Therefore, watchers that use template refs should be defined with the `flush: 'post'` option. This will run the effect *after* the DOM has been updated and ensure that the template ref stays in sync with the DOM and references the correct element.
+Therefore, watchers that use template refs should be defined with the `flush: 'post'` option. This will run the effect *after* the DOM has been mounted or updated and ensure that the template ref stays in sync with the DOM and references the correct element.
 
 ```vue
 <template>


src/guide/composition-api-template-refs.md
```diff
@@ -118,7 +118,7 @@ But a key difference to lifecycle hooks is that `watch()` and `watchEffect()` ef
 </script>
 ```
 
-Therefore, watchers that use template refs should be defined with the `flush: 'post'` option. This will run the effect *after* the DOM has been updated and ensure that the template ref stays in sync with the DOM and references the correct element.
+Therefore, watchers that use template refs should be defined with the `flush: 'post'` option. This will run the effect *after* the DOM has been mounted or updated and ensure that the template ref stays in sync with the DOM and references the correct element.
```
NataliaTepluhina, line 121: **question**: I am not completely sure what do you mean when saying `DOM was mounted` 🤔 I believe both component mount and component update processes update the DOM so the current version of the docs should be sufficient

Yukon123: In the same doc:[composition-api-template-refs:](https://v3.vuejs.org/guide/composition-api-template-refs.html#watching-template-refs)

We can see Line 93 describes:

> But a key difference to lifecycle hooks is that watch() and watchEffect() effects are run *before* the DOM is `mounted` or `updated` so the template ref hasn't been updated when the watcher runs the effect.

Thats  why i make this change to make the descrition consistent.

PR #1185, opened 2021-08-16
Update component-registration.md

olegBuda:
## Description of Problem
"defining component" is repeated twice - doesn't look right.

## Proposed Solution
Change repeated words.


NataliaTepluhina (approved):
@olegBuda nice change! Approved 👍🏻

PR #1187, opened 2021-08-18
Update reactivity-computed-watchers.md

Yukon123:
## Description of Problem
Computed is a funtion rather than method.So does the Watch and WatchEffect

## Proposed Solution

Change `method` into `funtion`

## Additional Information
We can learn from :https://v3.vuejs.org/guide/composition-api-introduction.html#standalone-computed-properties

>Similar to `ref` and `watch`, computed properties can also be created outside of a Vue component with the `computed` **function** imported from Vue.

>Just like how we set up a watcher on the user property inside our component using the watch option, we can do the same using the watch **function** imported from Vue. It accepts 3 arguments:

And it does show on source code

`export function computed<T>(
  getter: ComputedGetter<T>,
  debugOptions?: DebuggerOptions
): ComputedRef<T>`

### DIFF
--- a/src/guide/reactivity-computed-watchers.md
+++ b/src/guide/reactivity-computed-watchers.md
@@ -4,7 +4,7 @@
 
 ## Computed values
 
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Thanks. I concur, they are functions and not methods.

PR #1188, opened 2021-08-18
fix: fixed content hiding behind the previous code snippet.

premnathkulal:
## Description of Problem
Some of the Documentation contents are hiding behind the previous code snippet. I have mentioned two of them and fixed them.

## Proposed Solution
Added additional line breaker in markdown.

### Here contents  are hiding behind the code snippet
Here contents  are hiding behind the code snippet
### Fixed
Fixed
### Here codepen-snippet  hiding behind the code snippet
Here codepen-snippet  hiding behind the code snippet
### Fixed
Fixed

### DIFF
--- a/src/guide/component-props.md
+++ b/src/guide/component-props.md
@@ -206,6 +206,7 @@ app.component('my-component', {
 })
 ```
 
+\
 When prop validation fails, Vue will produce a console warning (if using the development build).
 
 ::: tip Note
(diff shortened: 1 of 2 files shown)


skirtles-code:
Thanks for the PR.

I believe this is only a problem for Linux users and it is caused by a problem with the fonts and CSS. We're currently working on migrating the docs to a new theme and it should be fixed as part of that, without needing any changes to the markdown files.

I'm using Linux myself so I understand how irritating the overflowing line numbers are.

PR #1189, opened 2021-08-18
Wouldn't this be reactive now?

BrianARuff:
Since it's Vue3, wouldn't the injected data be reactive now since everything is reactive in Vue3? The docs here are contradicting that and are saying that you either need to pass the data down as a ref or as a computed value.


skirtles-code:
Thank you for the PR but I believe the current documentation is correct.

The value `this.todos.length` would be reactive in Vue 3. But `provide` does not track changes, so if that number changes it will not be updated in descendent components that `inject` the value.

PR #1192, opened 2021-08-19
fix: missing comma in web-components.md

krienow:
Using the Example Vue CLI Config throws an error because a comma is missing. This PR will fix it.


NataliaTepluhina (approved):
Thank you for the fix!

Issue #1154, opened 2021-07-30
Example for watching multiple data sources with watch() is imprecise


zhenzhenChange:
### Version
3.2.0-beta.7

### Reproduction link
[https://codesandbox.io/s/fast-haze-xcrmb?file=/src/App.vue](https://codesandbox.io/s/fast-haze-xcrmb?file=/src/App.vue)

### Steps to reproduce
Run the above demo

or

The explanation in the documentation is not: a callback is executed after each change

[![WObf4s.png](https://z3.ax1x.com/2021/07/30/WObf4s.png)](https://imgtu.com/i/WObf4s)

### What is expected?
Shouldn't a callback be executed after every change?

### What is actually happening?
Execute only one callback for multiple changes

zhenzhenChange:
No，No，No

This is only one of the questions, there is a more important question is that after synchronizing each item of the changed data source, only one callback is executed?
For example, the two changes in the figure, why the `cb` of `watch` only once, I do not want to `await nextTick()`, because I need the two state before and after the value to do logic out.

Please see the following example:

```js
const a = ref(1)
const b = ref(2)

setTimeout(() => {
    a.value = 11
    b.value = 22
}, 1000)

// Only one `cb` will be executed
watch([a, b], ([a, b], [preA, preB]) => {
    if(a === b) {
        // ......
    }
})
```

```js
const a = ref(1)
const b = ref(2)

setTimeout(() => {
    a.value = 11
}, 1000)

setTimeout(() => {
    b.value = 22
}, 1000)

// will execute `cb` twice
watch([a, b], ([a, b], [preA, preB]) => {
    if(a === b) {
        // ......
    }
})
```

- Multiple changes in the same `setTimeout` will only execute `cb` once
- But separate changes in different `setTimeout` will execute `cb` for each change

Instead, in newer versions of `react`, there is a merge.（I don't mean to brag, just to compare the differences）
https://codesandbox.io/s/great-hopper-0omj1?file=/src/App.js

I think it's an important point to understand the design thinking behind this, it's not just about the documentation, thank you

zhenzhenChange:
Because `Vue` doesn't have `batchUpdate`, it doesn't have the logic of synchronously changing two `ref` twice, and if you split the `setTimeout`, you have the situation of synchronously triggering twice, and the logic is inconsistent

zhenzhenChange:
I hope you can transfer back, maybe we misunderstood, we just want to ask for confirmation

[...]

PR #1195, opened 2021-08-20
docs: Add additional statement in synchronously behavior  on initial run 

Yukon123:
## Description of Problem

The synchronously behavior wiht 'flush'option on initial run is not fully stated.

## Proposed Solution

Add description about synchronously behavior of 'flush' on initial run.

Add the description of what will happen when add `post` & `immediate` both in the same watcher.

## Additional Information

When i learn from the docs ,The `flush` & `immediate` its hard to understand for beginner like me.Therefore i do some test in my computer.Something Strange happen made me confuse more:if i do multiple state mutation in setup function.the initial run will not show asynchronously behavior but synchronously behavior. 

In that case i look into the source code and i find 

> REF:vue-next\packages\runtime-core\src\apiWatch.ts   row 340 => row 387

```js
  let scheduler: EffectScheduler
  if (flush === 'sync') {
    scheduler = job as any // the scheduler function gets called directly
  } else if (flush === 'post') {
    scheduler = () => queuePostRenderEffect(job, instance && instance.suspense)
  } else {
    // default: 'pre'
    scheduler = () => {
      if (!instance || instance.isMounted) {
        queuePreFlushCb(job)
      } else {
        // with 'pre' option, the first call must happen before
        // the component is mounted so it is called synchronously.
        job()
      }
    }
  }
```
Although we can find something description like this in [effect-flush-timing](https://v3.vuejs.org/guide/reactivity-computed-watchers.html#effect-flush-timing)

>Vue's reactivity system buffers invalidated effects and flushes them asynchronously to avoid unnecessary duplicate invocation when there are many state mutations happening in the same "tick". Internally, a component's update function is also a watched effect. When a user effect is queued, it is by default invoked before all component update effects:

```vue
<template>
  <div>{{ count }}</div>
</template>

<script>
export default {
  setup() {
    const count = ref(0)

    watchEffect(() => {
      console.log(count.value)
    })

    return {
      count
    }
  }
}
</script>
```

>In this example:

>- The count will be logged synchronously on initial run.
>- When `count` is mutated, the callback will be called **before** the component has updated.

But I think it's kind of ambigue  that timing of asynchronously flush,and will misleading beginner to think  it's asynchronously flush at any point if without 'sync' option.but option with 'pre' also synchronously flush.

Even if that's an fully stated in [effect-flush-timing](https://v3.vuejs.org/guide/reactivity-computed-watchers.html#effect-flush-timing),I think it's still necessary for this additional statement.

### DIFF
--- a/src/api/instance-methods.md
+++ b/src/api/instance-methods.md
@@ -181,11 +181,11 @@
 
   The default value is `'pre'`, which specifies that the callback should be invoked before rendering. This allows the callback to update other values before the template runs.
 
-  The value `'post'` can be used to defer the callback until after rendering. This should be used if the callback needs access to the updated DOM or child components via `$refs`.
+  The value `'post'` can be used to defer the callback until after rendering. This should be used if the callback needs access to the updated DOM or child components via `$refs`.Except it contain option with `immediate: true` , there will be a result logged before rendering on initial run.
 
   If `flush` is set to `'sync'`, the callback will be called synchronously, as soon as the value changes.
 
-  For both `'pre'` and `'post'`, the callback is buffered using a queue. The callback will only be added to the queue once, even if the watched value changes multiple times. The interim values will be skipped and won't be passed to the callback.
+  For both `'pre'` and `'post'`, the callback is buffered using a queue. The callback will only be added to the queue once, even if the watched value changes multiple times. The interim values will be skipped and won't be passed to the callback.It should be noted that if during the initial run with default `'pre'` option, the callback will not buffered but called synchronously.
 
   Buffering the callback not only improves performance but also helps to ensure data consistency. The watchers won't be triggered until the code performing the data updates has finished.
 


Yukon123:
My test example:
- 

```
<template>
  <div>
    <div @click="foo += 1">{{ foo }}</div>
  </div>
</template>

<script>
import { ref, watch, onMounted, onUpdated, onBeforeMount, onBeforeUpdate } from 'vue'

export default {
  setup() {
    const foo = ref(1)

    //lifecycle-hooks
    onBeforeMount(() => console.log('BeforeMount'))
    onMounted(() => console.log('Mounted'))
    onBeforeUpdate(() => {console.log('BeforeUpdate')})
    onUpdated(() => {console.log('Updated') })

    watch(
      foo,
      (newV, oldV) => {
        console.log('Watch foo post with immediate:', newV, oldV)
      },
      {
        immediate: true,
        flush: 'post',
      }
    )
    watch(
      foo,
      (newV, oldV) => {
        console.log('Watch foo post:', newV, oldV)
      },
      {
        flush: 'post',
      }
    )
    watch(
      foo,
      (newV, oldV) => {
        console.log('Watch foo pre:', newV, oldV)
      },
      {
        flush: 'pre',
      }
    )
    watch(
      foo,
      (newV, oldV) => {
        console.log('Watch foo pre with immediate:', newV, oldV)
      },
      {
        immediate: true,
        flush: 'pre',
      }
    )

    foo.value += 1
    foo.value += 1
    return {
      foo,
    }
  },
}
</script>

```

skirtles-code:
Thanks for the PR.

I've only had chance to skim it so far, but I suspect this is the same problem that is described in https://github.com/vuejs/docs/issues/1154#issuecomment-894452072. I think this needs to be fixed in the code rather than the docs. It hasn't been forgotten, I just haven't had chance to attempt a fix yet. I'm hoping to have time to take a look at it next week.

I would add that mutating something immediately in `setup` to trigger a watcher is not that common in actual applications. It tends to come up more in examples rather than real code. It's still important, just for different reasons.

I'd like to leave this PR open until I've had chance to take a look at it properly. I hope that's OK with everyone else.

zhenzhenChange:
@Yukon123  This is mainly a consistency issue, because you sometimes trigger `twice` and sometimes trigger `once`.
The same code, running in different environments at different moments, executes with different results, and consistency is not guaranteed.

---

`Twice`

[](https://imgtu.com/i/fiKTHg)

---

`Once`

[](https://imgtu.com/i/hZSzq0)

Yukon123:
@zhenzhenChange Thanks for the reply and example.

skirtles-code:
I believe this was fixed in 3.2.34-beta.1.

PR #1197, opened 2021-08-21
Change `is` prop to `is` attr

Yukon123:
## Description of Problem
`is` is special attribute 

https://v3.cn.vuejs.org/api/special-attributes.html#is

## Proposed Solution

Change `is` prop to `is` attr


NataliaTepluhina (changes requested):
@Yukon123 I totally agree with your point 👍🏻 Could we please replace `attr` with `attribute` for all occurrences?

NataliaTepluhina (approved):

NataliaTepluhina:
@Yukon123 thank you for the fixes!

PR #1198, opened 2021-08-21
Change `property` to `attribute`

Yukon123:
## Description of Problem
It's'  `attribute`  rather than `property` .
## Proposed Solution
Change `property` to `attribute`
## Additional Information
This correction will make the docs more precise and easier to understand

I try attribute with 'foo', the result is just like the expect.

### DIFF
--- a/src/guide/migration/v-bind.md
+++ b/src/guide/migration/v-bind.md
@@ -12,11 +12,11 @@ badges:
 
 ## Introduction
 
-When dynamically binding attributes on an element, a common scenario involves using both the `v-bind="object"` syntax as well as individual properties in the same element. However, this raises questions as far as the priority of merging.
+When dynamically binding attributes on an element, a common scenario involves using both the `v-bind="object"` syntax as well as individual attributes in the same element. However, this raises questions as far as the priority of merging.
 
 ## 2.x Syntax
 
-In 2.x, if an element has both `v-bind="object"` and an identical individual property defined, the individual property would always overwrite bindings in the `object`.
+In 2.x, if an element has both `v-bind="object"` and an identical individual attribute defined, the individual attribute would always overwrite bindings in the `object`.
 
 ```html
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@Yukon123 could you please explain why do you think it will be more precise? Binding on the element can be a property or an attribute, so I am not sure I understand why do we need to make this change.

Yukon123:
The  MDN define attribute like https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
>Elements in HTML have attributes; these are **additional values** that configure the elements or adjust their behavior in various ways to meet the criteria the users want.

Attribute are additional values to meet the criteria the users want.What we bind manually  is called attribute

Stack Overflow has a similar view :https://stackoverflow.com/questions/6003819/what-is-the-difference-between-properties-and-attributes-in-html/6004028#6004028

>When writing HTML source code, you can define attributes on your HTML elements. Then, once the browser parses your code, a corresponding DOM node will be created. This node is an object, and therefore it has properties.

It's called binding attribute on element rather than binding property? Isn't it?

Yukon123:
@NataliaTepluhina Sorry for bothering you.I want to explain why i add word *mounted* in early PR #1184 
I not sure if you have been read my last comment.
I agree both component mount and component update processes update the DOM 

In the same doc:[composition-api-template-refs:](https://v3.vuejs.org/guide/composition-api-template-refs.html#watching-template-refs)

We can see Line 93 describes:

> But a key difference to lifecycle hooks is that watch() and watchEffect() effects are run *before* the DOM is `mounted` or `updated` so the template ref hasn't been updated when the watcher runs the effect.

Thats  why i make this change to make the descrition consistent.

NataliaTepluhina (approved):
@Yukon123 thank you for the additional explanation! I agree with your arguments, let's change this to attribute

PR #1213, opened 2021-08-28
Update introduction.md

TheArkhamKnight781526:
Got this error while following along with the List Rendering section - 'Elements in iteration expect to have 'v-bind:key' directives' - and I fixed it by adding the keys. I think this would be a useful thing to include in the documentation.

## Description of Problem
Copying the code given returns this error - 'Elements in iteration expect to have 'v-bind:key' directives'.

## Proposed Solution
Add keys to the mapping to avoid potential errors and make it clear that they are needed for lists.

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -189,7 +189,7 @@ There are quite a few other directives, each with its own special functionality.
 ```html
 <div id="list-rendering">
   <ol>
-    <li v-for="todo in todos">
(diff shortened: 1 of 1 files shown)


wxsms:
The error you mention should be from eslint, not from Vue. Actually, v-for does not required a key to be work.

NataliaTepluhina:
Closing this PR as `key` is not required by the framework indeed. The warning is produced by ESLint plugin

PR #1216, opened 2021-08-30
docs: Update watch.md

wxsms:
## Description of Problem

rest of this article used `deep` **option**. just for consistence.


skirtles-code (approved):
Agreed, I much prefer `option`.

PR #1220, opened 2021-08-31
docs: Update forms.md

zhangenming:
## Description of Problem
a private opinions


NataliaTepluhina:
Thank you for your contribution! I believe the current version  is more clear so I leave it as it is now

PR #1228, opened 2021-09-04
Update component-provide-inject.md

karanodedra20:
## Description of Problem
Typo in the file.

## Proposed Solution
Correction.

### DIFF
--- a/src/guide/component-provide-inject.md
+++ b/src/guide/component-provide-inject.md
@@ -49,7 +49,7 @@ app.component('todo-list-statistics', {
 })
 ```
 
-However, this won't work if we try to provide some component instance property here:
+However, this won't work if we try to provide same component instance property here:
 
 ```js
 app.component('todo-list', {


src/guide/component-provide-inject.md
```diff
@@ -49,7 +49,7 @@ app.component('todo-list-statistics', {
 })
 ```
 
-However, this won't work if we try to provide some component instance property here:
+However, this won't work if we try to provide same component instance property here:
```
skirtles-code, line 52:
```suggestion
However, this won't work if we try to provide a component instance property here:
```

I don't think `same` is the intended word in this context.

Perhaps `a` would be clearer than `some`?

skirtles-code (changes requested):

src/guide/component-provide-inject.md
karanodedra20: I thought we were talking about the instance property of that particular component, therefore _same_, else it's correct.

PR #1231, opened 2021-09-07
Update component-props.md

w1301625107:
add description of condition about implied casting

### DIFF
--- a/src/guide/component-props.md
+++ b/src/guide/component-props.md
@@ -61,6 +61,7 @@ In the two examples above, we happen to pass string values, but _any_ type of va
 
 ```html
 <!-- Including the prop with no value will imply `true`. -->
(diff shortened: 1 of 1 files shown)


src/guide/component-props.md
```diff
@@ -61,6 +61,7 @@ In the two examples above, we happen to pass string values, but _any_ type of va
 
 ```html
 <!-- Including the prop with no value will imply `true`. -->
+<!-- If you don't set is-published's type to Boolean in Props, it will be a empty string instead of true value. -->
```
NataliaTepluhina, line 64: A few fixes:

```suggestion
<!-- If you don't set is-published's type to Boolean in props, it will be an empty string instead of "true" value. -->
```

NataliaTepluhina (changes requested):

NataliaTepluhina (approved):

PR #1239, opened 2021-09-12
update keycode-modifiers.md

wxsms:
## Description of Problem

cddb981d714f6be28780c29613df030


skirtles-code:
Thanks for the PR, but I believe the suggested change is grammatically incorrect.

Something like 'we now recommend using' would be correct, but I don't believe it would be an improvement over what we have currently.

wxsms:
It's find. I'm not a native english speaker, it's a tip from jetbrains ide. maybe you're right.

PR #1240, opened 2021-09-13
use kebab-case for event emission

wysohn:
## Description of Problem

So I am following through with the documentation, and I encountered the custom event emission section and realized that one part of the documentation uses camelCase and the other part use kebab-case. So I did some google search and [this](https://stackoverflow.com/questions/42441952/vue-js-custom-event-naming/50423976) SO article showed up.

Based on the information, the official recommendation ([link](https://vuejs.org/v2/guide/components-custom-events.html#Event-Names)) is to use kebab-case for the event name.

As a newcomer, this is quite mind-boggling since I think it's natural to expect the same variable name for both hands.

It's surprising that it still works for both camelCase and kebab-case when I tried it in the CODEPEN editor example in [this](https://v3.vuejs.org/guide/component-basics.html#base-example) article. But it would be a smoother learning experience for me if the events were indeed using the exact same name.

## Proposed Solution

camelCase to kebab-case

## Additional Information

I am new to Vue, so I don't know much about it yet.

### DIFF
--- a/src/guide/component-basics.md
+++ b/src/guide/component-basics.md
@@ -220,7 +220,7 @@ When we click on the button, we need to communicate to the parent that it should
 Then the child component can emit an event on itself by calling the built-in [**`$emit`** method](../api/instance-methods.html#emit), passing the name of the event:
 
 ```html
-<button @click="$emit('enlargeText')">
+<button @click="$emit('enlarge-text')">
   Enlarge text
 </button>
 ```
@@ -234,7 +234,7 @@ We can list emitted events in the component's `emits` option:
 ```js
 app.component('blog-post', {
   props: ['title'],
-  emits: ['enlargeText']
+  emits: ['enlarge-text']
 })
 ```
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

The references you linked are all for Vue 2. The handling of event names has changed in Vue 3. They now behave more like props, with camelCase and kebab-case being interchangeable. That is explained here: https://v3.vuejs.org/guide/component-custom-events.html#event-names

However, that page comes after the section you mentioned in this PR, so it can't be treated as assumed knowledge.

I'm not sure how best to proceed, but perhaps it would be better to acknowledge the case difference explicitly, rather than removing it? We hit the same problem again with the `v-model` example further down the page. The warning message in that section is potentially confusing if we haven't established the basic idea earlier.

wysohn:
> Thanks for the PR!
> 
> The references you linked are all for Vue 2. The handling of event names has changed in Vue 3. They now behave more like props, with camelCase and kebab-case being interchangeable. That is explained here: https://v3.vuejs.org/guide/component-custom-events.html#event-names
> 
> However, that page comes after the section you mentioned in this PR, so it can't be treated as assumed knowledge.
> 
> I'm not sure how best to proceed, but perhaps it would be better to acknowledge the case difference explicitly, rather than removing it? We hit the same problem again with the `v-model` example further down the page. The warning message in that section is potentially confusing if we haven't established the basic idea earlier.

Ah that makes complete sense now.

For some reason, I arrived at the v2 tutorial when I searched it on google, and I just realized that there is a link to the v3.

I will close it for now and perhaps re-open it once I get a better idea about the Vue in general.

Thank you!

PR #1247, opened 2021-09-16
Add Airbrake integration to errorHandler docs

kyrylo:
I work for Airbrake. While working on a Vue app I stumbled upon the docs that
mention error handling with 3rd party services (Sentry, Bugsnag). I also saw a
PR adding Bugsnag: https://github.com/vuejs/vuejs.org/pull/1311

I thought, it would be nice to mention Airbrake as well, since we have also have
a decent support for Vue.

### DIFF
--- a/src/api/application-config.md
+++ b/src/api/application-config.md
@@ -28,7 +28,7 @@ app.config.errorHandler = (err, vm, info) => {
 
 Assign a handler for uncaught errors during component render function and watchers. The handler gets called with the error and the application instance.
 
-> Error tracking services [Sentry](https://sentry.io/for/vue/) and [Bugsnag](https://docs.bugsnag.com/platforms/browsers/vue/) provide official integrations using this option.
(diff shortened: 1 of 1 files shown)


skirtles-code:
The equivalent PR for the Vue 2 docs: https://github.com/vuejs/vuejs.org/pull/2881

skirtles-code:
Hi @kyrylo. Thanks for the PR.

We're generally quite cautious about mentioning third-party tools in the documentation, as it can be viewed as an endorsement or recommendation. We typically only list tools that the Vue core team have used and had good experiences with.

I reached out to the core team to see whether anyone was familiar with Airbrake but none of them have used it. I'm afraid we won't be able to include Airbrake in the documentation at this stage, but if it gains traction within the Vue community then it should be possible to include it at some point in the future.

destari:
So on. this page:
https://v3.vuejs.org/api/application-config.html#errorhandler
where you mention "official integrations", that isn't actually the case?   Seems very unfortunate for the Due community if the only "official integrations" are the ones that the developers themselves have used.  Seems opposite of open/inclusive.

destari:
Makes sense, thanks for the detail and extra clarification!

PR #1260, opened 2021-09-27
Fixed irregular code block borders

mychidarko:
## Description of Problem

Code blocks with selected lines eg: ***```js{8-10}*** had no consistent radius applied as opposed to code blocks without selected lines

## Proposed Solution

Added a border radius to code blocks with selected lines to maintain design consistency


NataliaTepluhina (approved):
Thank you for the fix!

PR #1272, opened 2021-10-06
Update built-in-components.md

xmasuhai:
provide keys to  <component :is="vnode">

## Description of Problem

When using <component :is="vnode"> and passing vnode of the same type, you need to provide keys

## Proposed Solution

<component :is="current" :key="selected" />

## Additional Information

according to [Issus #2013](https://github.com/vuejs/vue-next/issues/2013)


NataliaTepluhina (approved):
Thank you for the addition!

PR #1273, opened 2021-10-06
feat(themes): new themes page

kiaking:
Implement new themes page.

Screen Shot 2021-10-06 at 19 38 52

Screen Shot 2021-10-06 at 19 38 55


stafyniaksacha:
Hello ! 
Do you think we can add vuero to your list ? We try to follow all good practices from vue 3 (We can send invitation to check if you need)

https://vuero.cssninja.io/

PR #1301, opened 2021-10-27
added missing definite article

geogrfer:
## Description of Problem
there's a missing definite article an an unconventional usage (do not wish to)
## Proposed Solution
I've added a definite article and replaced 'do not wish not' in favor of 'wish not to' 
## Additional Information
n/a


skirtles-code:
Thanks for the PR, but I believe the current wording of that sentence reads better than the suggested improvement.

I don't think adding an indefinite article is correct in this context. I believe the word `routing` is being used as a gerund and English (unlike many other languages) typically doesn't use articles with gerunds.

PR #1306, opened 2021-10-30
Update introduction.md

wxsms:
remove strange markups. this paragraph is already shown on page.

### DIFF
--- a/src/guide/ssr/introduction.md
+++ b/src/guide/ssr/introduction.md
@@ -40,7 +40,7 @@ If you're using [webpack](https://webpack.js.org/), you can add prerendering wit
 
 This guide will be very in-depth and assumes you are already familiar with Vue.js itself, and have a decent working knowledge of Node.js and webpack.
 
-[//]: # 'If you prefer a higher-level solution that provides a smooth out-of-the-box experience, you should probably give [Nuxt.js](https://nuxtjs.org/) a try. It's built upon the same Vue stack but abstracts away a lot of the boilerplate, and provides some extra features such as static site generation. However, it may not suit your use case if you need more direct control of your app's structure. Regardless, it would still be beneficial to read through this guide to understand better how things work together.'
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):

skirtles-code:
I don't think this paragraph, or the one that follows it, are supposed to be shown. I'm pretty sure they weren't visible when the page was first added, so I suspect a dependency change could be to blame. I would imagine that standard HTML comments would hide them?

wxsms:
> I don't think this paragraph, or the one that follows it, are supposed to be shown. I'm pretty sure they weren't visible when the page was first added, so I suspect a dependency change could be to blame. I would imagine that standard HTML comments would hide them?

as I see from the git history, it's written by @NataliaTepluhina , and this paragraph had not change since it was added. which means it was shown from the beginning.

skirtles-code:
My theory was that it was originally hidden and it was revealed later due to upgrading VuePress. However, I just checked the original test deployment at https://deploy-preview-864--vue-docs-v3-legacy.netlify.app/guide/ssr/introduction.html#about-this-guide and that doesn't appear to be the case. It has, indeed, been there from the start.

That said, it definitely wasn't meant to be showing. It was discussed on the original PR and the intention was to hide it, which is what those extra characters were supposed to be do: https://github.com/vuejs/docs/pull/864#discussion_r580370801. Even though there is now a version of Nuxt that supports Vue 3, it isn't yet production ready so the description given in that paragraph is still a bit misleading.

wxsms:
Ah. So it's a mistake to be shown. Please help yourself for that if you wish. I just make it display better while the fact is it's already displaying.

Justineo:
As Nuxt 3 beta is now live maybe we shall update the link to https://v3.nuxtjs.org/ so we don't need to hide it now.

skirtles-code:
@Justineo I'll put in a PR to update the link and tweak the wording to reflect the beta status of Nuxt 3.

PR #1307, opened 2021-10-30
Update deprecated app.config.isCustomElement

Miguel-Bento-Github:
## Description
`app.config.isCustomElement` is deprecated and should be replaced with `app.config.compilerOptions.isCustomElement`

## Solution
Change `app.config.isCustomElement` with `app.config.compilerOptions.isCustomElement` where referenced.


NataliaTepluhina (approved):
@Miguel-Bento-Github Thank you for fixing this! 🎉

PR #1311, opened 2021-11-03
Use createSSRApp instead of createApp

skirtles-code:
I believe `entry-client.js` should be using `createSSRApp` instead of `createApp`. Both will work, but `createApp` will create an entirely new DOM in the browser, whereas `createSSRApp` will hydrate the existing DOM.

I've also made a few other minor corrections. I think they're self-explanatory but let me know if not.

### DIFF
--- a/src/guide/ssr/routing.md
+++ b/src/guide/ssr/routing.md
@@ -25,14 +25,14 @@ And update our client and server entries:
 
 ```js
 // entry-client.js
-import { createApp } from 'vue'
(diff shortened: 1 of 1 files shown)


skirtles-code:
@CyberAP. It seems I can't add you as a reviewer, so I'll tag you in via a comment instead. I thought you might want to review this. Thanks.

CyberAP:
@skirtles-code nice catch, LGTM!

PR #1314, opened 2021-11-05
A better example for v-html

mehanalavimajd:
## Description of Problem
The example in the document is not very good
## Proposed Solution
Changing the example 
## Additional Information
none


NataliaTepluhina (approved):
@mehanalavimajd I agree, detailed example looks better 🎉

PR #1315, opened 2021-11-05
Fix typo

dews:
## Description of Problem
When executing command `npm run build:server` got an error: "ERROR  Build failed with errors."

## Proposed Solution
The App from `import App` is never used and Vue is not defined so it's a typo.
Change text from Vue to App

### DIFF
--- a/src/guide/ssr/structure.md
+++ b/src/guide/ssr/structure.md
@@ -147,7 +147,7 @@ import { createSSRApp } from 'vue'
 import App from './App.vue'
 
 export default function () {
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Thanks. :+1:

I fixed a similar typo in https://github.com/vuejs/docs/pull/1311/files#diff-4f2c4e63ef6698688f98aa73d423f140363ff0121a3d6715430861651eefb9f4R53 but I missed this one. From a quick search I don't think there are any more.

PR #1317, opened 2021-11-07
Correct a common misconception regarding argument passing

rsoltanzadeh:
JavaScript is always pass-by-value. The difference is that with reference types the *value of the reference* is passed (but never the reference itself). If JavaScript were pass-by-reference, the following JSFiddle would have logged `2` instead of `1`: https://jsfiddle.net/zryav4fx/

For more information:
  - https://stackoverflow.com/a/7744623
  - https://stackoverflow.com/a/5314911
  - https://stackoverflow.com/a/430958

### DIFF
--- a/src/guide/composition-api-introduction.md
+++ b/src/guide/composition-api-introduction.md
@@ -157,9 +157,7 @@ counter.value++
 console.log(counter.value) // 1
 ```
 
-Wrapping values inside an object might seem unnecessary but is required to keep the behavior unified across different data types in JavaScript. That’s because in JavaScript, primitive types like `Number` or `String` are passed by value, not by reference:
-
-![Pass by reference vs pass by value](https://blog.penjee.com/wp-content/uploads/2015/02/pass-by-reference-vs-pass-by-value-animation.gif)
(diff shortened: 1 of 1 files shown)


skirtles-code:
Some thoughts...

I do share your concern, I tend to avoid using those terms myself.

There are two key problems with them, in my opinion. One is that they were intended to refer to two evaluation strategies present in programming languages in the early 1960s, with more modern languages not accurately being described by either term. The other problem is that they are only intended to describe how parameters are passed to functions, whereas they tend to get used to try to explain how reference types work more generally.

The difficulty is trying to stay technically accurate whilst also effectively communicating with the reader. I try to use the terms 'primitive types' and 'reference types' when I'm trying to explain how JavaScript behaves, but those terms tend not to resonate quite as well.

I'd also note that what's currently written in the docs is not incorrect. It doesn't actually say that objects are pass-by-reference, though I would agree that it could be taken as implied.

More broadly, I don't think it's really necessary for the docs to get into a discussion of primitives vs references to explain why `ref` is required. This arguably does more harm than good, as there's a common misconception that `ref` is just to be used for primitives.

As far as I can tell, the docs rewrite on the `next` branch has avoided using these terms altogether.

PR #1319, opened 2021-11-07
fix: Explain usage of custom directives with script-setup

LinusBorg:
## Description of Problem

In the docs for `<script setup>`, we explain how to use components directly in the template without registering them, but we don't do the same for custom directives.

## Proposed Solution

add a small section to the `<script setup>` docs after the component explanations.

### DIFF
--- a/src/api/sfc-script-setup.md
+++ b/src/api/sfc-script-setup.md
@@ -131,6 +131,31 @@ import * as Form from './form-components'
 </template>
 ```
 
+## Using Custom Directives
+
+Globally registered custom directives just work as expected, and local ones can be used directly in the template, much like we explained above for components. 
+
+But there's one restriction to be aware of: You must name local custom directives according to the following schema: `vNameOfDirective` in order for them to be directly usable in the template.
+
+```html
+<script setup>
+const vMyDirective = {
+  beforeMount: (el) => {
+    // do someting with the element
+  }
+}
+</script>
+<template>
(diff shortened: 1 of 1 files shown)


src/api/sfc-script-setup.md
```diff
@@ -131,6 +131,31 @@ import * as Form from './form-components'
...
+
+Globally registered custom directives just work as expected, and local ones can be used directly in the template, much like we explained above for components. 
+
+But there's one restriction to be aware of: You must name local custom directives according to the following schema: `vNameOfDirective` in order for them to be directly usable in the template.
+
+```html
+<script setup>
+const vMyDirective = {
+  beforeMount: (el) => {
+    // do someting with the element
```
skirtles-code, line 144:
```suggestion
    // do something with the element
```

src/api/sfc-script-setup.md
```diff
@@ -131,6 +131,31 @@ import * as Form from './form-components'
...
+  }
+}
+</script>
+<template>
+  <h1 v-my-directive>This is a Heading</h1>
+</template>
+```
+```html
+<script setup>
+  // imports also work, and can be renamed to fit the requires naming schema
```
skirtles-code, line 154:
```suggestion
  // imports also work, and can be renamed to fit the required naming schema
```

NataliaTepluhina (approved):
@LinusBorg great addition, thank you!

PR #1321, opened 2021-11-11
Update events.md

wxsms:
## Description of Problem

`.prevent` dose not prevent click itself, but its default behavior.

### DIFF
--- a/src/guide/events.md
+++ b/src/guide/events.md
@@ -171,7 +171,7 @@ To address this problem, Vue provides **event modifiers** for `v-on`. Recall tha
 ```
 
 ::: tip
-Order matters when using modifiers because the relevant code is generated in the same order. Therefore using `@click.prevent.self` will prevent **all clicks** while `@click.self.prevent` will only prevent clicks on the element itself.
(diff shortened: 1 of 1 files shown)


mehanalavimajd (commented):
I agree but I think it would be better if you delete the single quotation

wxsms:
> I agree but I think it would be better if you delete the single quotation

click -> click's default behavior.

clicks -> clicks' default behavior.

skirtles-code:
Thanks for the PR!

In my opinion, trying to be more precise just makes it more difficult to follow. I don't know whether it's worth making a distinction between 'preventing a click' and 'preventing a click's default behavior' in the context of this sentence. It's a pretty subtle distinction and I don't think the reader is likely to be confused by the simplified version. Establishing the exact purpose of the `prevent` modifier is handled earlier in the section, by the point we reach this sentence I think just using the word 'prevent' is sufficient.

But I'll leave this open for now to see what others think.

wxsms:
there's another tip below which mentions about default behavior:

> Don't use `.passive` and `.prevent` together, because `.prevent` will be ignored and your browser will probably show you a warning. Remember, `.passive` communicates to the browser that you _don't_ want to prevent the event's default behavior.

for my opinion, **prevent event** and **prevent event's default behavior** has huge difference, which I think that the front one means prevent the event from trigger.

wxsms:
is merged in https://github.com/vuejs/docs/pull/1425, closing.

PR #1324, opened 2021-11-11
Using a real html tag name instead of unknown `custom1` component in sfc api

mehanalavimajd:
## Description of Problem
Custom1 is not defined so you should make a one(the example will be so long) or using a real tag name like p

### DIFF
--- a/src/api/sfc-spec.md
+++ b/src/api/sfc-spec.md
@@ -25,9 +25,9 @@ export default {
 }
 </style>
 
-<custom1>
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR, but I think you've misunderstood that example. It's showing the use of a custom section in an SFC and a real HTML tag name is not required. The exact name of the tag would depend on the requirements of the tool consuming that section of the SFC. There's more explanation further down that page.

PR #1325, opened 2021-11-11
docs: remove unused var in computed.md

brrrdy:
## Description of Problem
Unused var 'oldQuestion' in question method

## Proposed Solution
Remove it for clarity

## Additional Information
"Watch basic example" codepen at https://codepen.io/team/Vue/pen/GRJGqXp should be updated as well

### DIFF
--- a/src/guide/computed.md
+++ b/src/guide/computed.md
@@ -169,7 +169,7 @@ For example:
     },
     watch: {
       // whenever question changes, this function will run
-      question(newQuestion, oldQuestion) {
+      question(newQuestion) {
         if (newQuestion.indexOf('?') > -1) {
           this.getAnswer()
         }


skirtles-code:
Thanks for the PR!

I believe this dates back to the Vue 2 docs, which didn't use `newQuestion` or `oldQuestion`. I think this is the first point that watchers are introduced and the intention here it to show that those values are available, even if that specific example doesn't use them.

Ideally the example should use both values, but it's not clear how it might be changed to make that possible.

In my opinion, it is better for this example to show the available arguments, even if it doesn't use them.

brrrdy:
Thank you for the explanation, that makes sense. I see now that functionality is detailed more here: [https://v3.vuejs.org/guide/reactivity-computed-watchers.html#watch](https://v3.vuejs.org/guide/reactivity-computed-watchers.html#watch)

> Ideally the example should use both values, but it's not clear how it might be changed to make that possible.

I agree, but can't think of a way to use `oldQuestion` in a meaningful or illustrative way in the given example either. I'll go ahead and close this and revisit it if I can think of a better change.

PR #1326, opened 2021-11-12
Fix bug in the rendering of the v-html example

### LINKED ISSUE #1323: The example in `v-html` is really running
Hi i made a better example for `v-html` in #1314 but it seems it isn't working correctly because the vue code is really running 
screenshot:

image

skirtles-code:
Closes #1323.

### DIFF
--- a/src/api/directives.md
+++ b/src/api/directives.md
@@ -35,7 +35,7 @@
 - **Example:**
 
   ```html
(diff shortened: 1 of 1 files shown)


mehanalavimajd (approved):
The bug fixes!

bencodezen (approved):

PR #1328, opened 2021-11-13
More detailed example in special-attributes.md

mehanalavimajd:
The item.id is not defined in the code so I think the index in the v-for is better. Also `...` in the li isn't very detailed

### DIFF
--- a/src/api/special-attributes.md
+++ b/src/api/special-attributes.md
@@ -12,7 +12,7 @@
 
   ```html
   <ul>
-    <li v-for="item in items" :key="item.id">...</li>
+    <li v-for="(item , index) in items" :key="index">item</li>
   </ul>
   ```
 


src/api/special-attributes.md
```diff
@@ -12,7 +12,7 @@
 
   ```html
   <ul>
-    <li v-for="item in items" :key="item.id">...</li>
+    <li v-for="(item , index) in items" :key="index">item</li>
```
FloEdelmann, line 15: Using the `index` as a `key` is not really helping much, so `item.id` is a better example.

See also https://rimdev.io/the-v-for-key/ or https://stackoverflow.com/questions/44531510/why-not-always-use-the-index-as-the-key-in-a-vue-js-for-loop.

FloEdelmann (changes requested):

NataliaTepluhina:
Closing an issue as we didn't use `index` as a `key` intentionally

PR #1338, opened 2021-11-20
use camelCase

wxsms:
## Description of Problem

`camelCase` should be better, and used lot more in other places of the document.


skirtles-code:
Thanks for the PR!

I checked this with the rest of the docs team and we seem to be in agreement that writing casing conventions in their respective styles is a good idea. The only concern we had was around accessibility, but we did some tests with screen readers and it seems that's not a problem.

PR #1345, opened 2021-11-22
Update sfc-script-setup.md

wxsms:
## Description of Problem

The document has not covered how to use `script setup` with render functions, it ends up with no way to do so (for now).

see https://github.com/vuejs/vue-next/issues/4980


wxsms:
I have no idea why this CI failed, looks like it triggered a HTTP 403 error.

NataliaTepluhina (approved):

PR #1346, opened 2021-11-22
better example for kebab-case vs camelCase

santiago-rodrig:
The whole point of camelCase vs kebab-case when talking about props is when kebab-case is required in DOM templates rather than Vue's instance template property, that's why I think using an example where the attribute is using camelCase is the right thing to do for this explanation.

## Description of Problem

Example about camelCase vs kebab-case props is not good enough.

## Proposed Solution

I provided a better example.

## Additional Information

N/A

### DIFF
--- a/src/guide/component-props.md
+++ b/src/guide/component-props.md
@@ -254,7 +254,14 @@ to validate that the value of the `author` prop was created with `new Person`.
 HTML attribute names are case-insensitive, so browsers will interpret any uppercase characters as lowercase. That means when you're using in-DOM templates, camelCased prop names need to use their kebab-cased (hyphen-delimited) equivalents:
 
 ```js
-const app = Vue.createApp({})
+const app = Vue.createApp({
+  template: `
+    <div>
+      <h1>Using camelCase</h1>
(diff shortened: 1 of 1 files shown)


src/guide/component-props.md
```diff
@@ -254,7 +254,14 @@ to validate that the value of the `author` prop was created with `new Person`.
 HTML attribute names are case-insensitive, so browsers will interpret any uppercase characters as lowercase. That means when you're using in-DOM templates, camelCased prop names need to use their kebab-cased (hyphen-delimited) equivalents:
 
 ```js
-const app = Vue.createApp({})
+const app = Vue.createApp({
```
NataliaTepluhina, line 257: I would suggest moving this example below, after the sentence `Again, if you're using string templates, this limitation does not apply.` otherwise it creates confusion

NataliaTepluhina (changes requested):

skirtles-code:
Thanks for the PR.

We haven't been able to merge this because Natalia's feedback hasn't been addressed.

The new docs will be going live in a couple of weeks, with major changes to this page. At this point I think it would be better to close this PR.

PR #1353, opened 2021-11-30
update transitions-enterleave.md

wxsms:
## Description of Problem

when I read this example, I don't know that the replace button is clickable. I though the hover mask animation is what's talking about (which is not). So a tip might be better.

### DIFF
--- a/src/guide/transitions-enterleave.md
+++ b/src/guide/transitions-enterleave.md
@@ -504,6 +504,8 @@ We can use this to coordinate more expressive movement, such as a folding card, 
 
 <common-codepen-snippet title="Transition Modes Flip Cards" slug="76e344bf057bd58b5936bba260b787a8" :editable="false" />
 
+It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI micro-interactions!
+
 ## Transitioning Between Components
 
 Transitioning between components is even simpler - we don't even need the `key` attribute. Instead, we wrap a [dynamic component](component-basics.html#dynamic-components):
--- a/src/guide/transitions-enterleave.md
+++ b/src/guide/transitions-enterleave.md
@@ -500,7 +500,7 @@ Now let's update the transition for our on/off buttons with `out-in`:
 
 With one attribute addition, we've fixed that original transition without having to add any special styling.
 
-We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below (try clicking the replace button). It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI microinteractions:
+We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below. Try hovering over each image and clicking the replace button.
 
 <common-codepen-snippet title="Transition Modes Flip Cards" slug="76e344bf057bd58b5936bba260b787a8" :editable="false" />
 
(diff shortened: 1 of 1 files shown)


src/guide/transitions-enterleave.md
```diff
@@ -500,7 +500,7 @@ Now let's update the transition for our on/off buttons with `out-in`:
 
 With one attribute addition, we've fixed that original transition without having to add any special styling.
 
-We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below. It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI microinteractions:
+We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below (try clicking the replace button). It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI microinteractions:
 
 <common-codepen-snippet title="Transition Modes Flip Cards" slug="76e344bf057bd58b5936bba260b787a8" :editable="false" />
```
bencodezen, line 505:
```suggestion
<common-codepen-snippet title="Transition Modes Flip Cards" slug="76e344bf057bd58b5936bba260b787a8" :editable="false" />

It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI micro-interactions!


```

src/guide/transitions-enterleave.md
```diff
@@ -500,7 +500,7 @@ Now let's update the transition for our on/off buttons with `out-in`:
 
 With one attribute addition, we've fixed that original transition without having to add any special styling.
 
-We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below. It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI microinteractions:
+We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below (try clicking the replace button). It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI microinteractions:
```
bencodezen, line 503:
```suggestion
We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below. Try hovering over each image and clicking the replace button.
```

bencodezen (changes requested):
What do you think about this rewrite? It encourages the user to play with the demo first with a little bit of instruction that you originally suggested, and then the explanation follows it.

wxsms:
> What do you think about this rewrite? It encourages the user to play with the demo first with a little bit of instruction that you originally suggested, and then the explanation follows it.

looks great!

src/guide/transitions-enterleave.md
```diff
@@ -500,10 +500,12 @@ Now let's update the transition for our on/off buttons with `out-in`:
 
 With one attribute addition, we've fixed that original transition without having to add any special styling.
 
-We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below. It's actually two elements transitioning between each other, but since the beginning and end states are scaling the same: horizontally to 0, it appears like one fluid movement. This type of sleight-of-hand can be very useful for realistic UI microinteractions:
+We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below. Try hovering over each image and clicking the replace button.
```
wxsms, line 503:
```suggestion
We can use this to coordinate more expressive movement, such as a folding card, as demonstrated below. Try hovering over each image and clicking the replace button:
```

skirtles-code:
Sorry for the delay.

I believe this example has been removed from the new docs, so I'm going to close this.

PR #1357, opened 2021-11-30
translations-data: add id url

simon04:


skirtles-code (approved):
Thanks for the PR!

I forget exactly why I didn't include this URL originally. I think it did exist, but my recollection is that it didn't seem to be showing the latest content from the corresponding GitHub repo.

Having just checked it now I don't see any reason why it can't be included.

Issue #1362, opened 2021-12-02
One Way Data Binding Documentation Needs Clarity


bengreenow:
Please can someone provide some clarity around the statement:

`Note that objects and arrays in JavaScript are passed by reference, so if the prop is an array or object, mutating the object or array itself inside the child component will affect parent state.`

made here: https://v3.vuejs.org/guide/component-props.html#one-way-data-flow
This note does not specifically tell us to avoid doing this as it an anti-pattern or bad practice. In Vue2 this was definitely considered bad practice because the parent would not react to it but in vue3 this does not seem to be the case we've been modifying array and object props by using toRefs in and are unsure if this is still considered a bad approach.

If so, please could clarity be added to the docs surrounding reference types specifically.

NataliaTepluhina:
In Vue 2 mutating object prop in child _will_ affect parent. This is clearly still a bad approach and we state this in `This means you should not attempt to mutate a prop inside a child component.` This applies to any kind of props.

I agree with that we should probably update the warning with something like "you should still avoid changing props in child component as it's breaking one-way data flow"

skirtles-code:
Closing following the changes in #1365.

PR #1365, opened 2021-12-03
Update Prop Mutation Docs to be more clear around mutation

MarkTallentire:
## Description of Problem
As noted in Issue 1362 https://github.com/vuejs/docs/issues/1362, there is a lack of clarity around mutating of objects and arrays being a bad practice.

## Proposed Solution

Updated the docs to specifically warn against mutation of any prop in particular Arrays and Objects.

### DIFF
--- a/src/guide/component-props.md
+++ b/src/guide/component-props.md
@@ -155,7 +155,7 @@ computed: {
 ```
 
 ::: warning Warning
-Note that objects and arrays in JavaScript are passed by reference, so if the prop is an array or object, mutating the object or array itself inside the child component **will** affect parent state and vue is unable to warn you against this. As a general rule, you should avoid mutating any prop, including objects and arrays as doing so ignores one way data binding and may cause undesired results.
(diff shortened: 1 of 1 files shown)


src/guide/component-props.md
```diff
@@ -154,8 +154,8 @@ computed: {
 }
 ```
 
-::: tip Note
-Note that objects and arrays in JavaScript are passed by reference, so if the prop is an array or object, mutating the object or array itself inside the child component **will** affect parent state.
+::: warning Warning
+Note that objects and arrays in JavaScript are passed by reference, so if the prop is an array or object, mutating the object or array itself inside the child component **will** affect parent state and vue is unable to warn you against this. As a general rule, you should avoid mutating any prop, including objects and arrays as doing so ignores one way data binding and may cause undesired results.
```
NataliaTepluhina, line 158: **suggestion**: just a few minor grammar fixes

```suggestion
Note that objects and arrays in JavaScript are passed by reference, so if the prop is an array or object, mutating the object or array itself inside the child component **will** affect the parent state and Vue is unable to warn you against this. As a general rule, you should avoid mutating any prop, including objects and arrays as doing so ignores one-way data binding and may cause undesired results.
```

NataliaTepluhina:
@MarkTallentire thank you for adding the note! It looks great overall, I've added just a few little tweaks 👍🏻

PR #1370, opened 2021-12-07
Clerification on disabling attributes inheritance.

anchan42:
Just try to make it clearer on what could be done after the inheritance is disabled.

## Description of Problem
Confusing sentence structure.
## Proposed Solution
Rewrite to make it clearer.
## Additional Information
I'm not a native English speaker, so I might be wrong.


skirtles-code (approved):
Thanks for the PR! I don't think the previous wording really made sense, so this is much better.

PR #1379, opened 2021-12-15
fix(compat): add a note about Nuxt Bridge for Nuxt.js users

superbiche:
## Description of Problem
The current piece of advice for Nuxt.js users (https://v3.vuejs.org/guide/migration/migration-build.html#intended-use-cases) is currently: 
> Server-side rendering: [...] If you are using Nuxt.js, it is probably better to wait for Nuxt 3.

Nevertheless, Nuxt Bridge can be a great alternative, and a nice way to see how much of a codebase will have to be updated to run Nuxt 3.

There's also no link to Nuxt 3, and the fact that it is now in beta phase is not mentioned. I think it would be helpful to give more information about it right the docs.

## Proposed Solution

- Add information + link to Nuxt Bridge "Getting Started" guide  
- Add link to Nuxt.js 3 "Getting Started" guide + a note to inform readers that it is currently in beta.

## Additional Information
I guess someone should add a reminder to update the docs again when Nuxt.js 3 gets a first stable release - it seems overkill to get the version during Vue docs build. I'll try to remember doing this!


NataliaTepluhina (approved):
@superbiche thank you, that's a nice addition 🙇🏻

PR #1384, opened 2021-12-21
Update component-basics.md

ekkoyr:
there is no options object in component at least in this instance


skirtles-code:
Thanks for the PR!

I'm not sure exactly what problem you're looking to address here. The existing docs look correct to me. Could you expand on your explanation of the problem, maybe we can find a way to word it better?

PR #1385, opened 2021-12-22
Fix typo on composition api faq

edimitchel:
## Description of Problem
Wrong word `algin` instead of `align`

## Proposed Solution
Use correct word

## Additional Information
This issue was probably not seen as algin is a correct word.


edimitchel:
I don't understand the check that fail
@yyx990803

NataliaTepluhina (approved):

PR #1388, opened 2021-12-23
Fix grammatical errors and introduce app.component() in introduction.md 

scientist27:
### Line 55
There are two "_the_"s missing, one before the word 'counter' and another before the word "rendered".
### Line 161
There is an unnecessary comma before the word "too". According to [Grammarly](https://www.grammarly.com/blog/comma-before-too/), the comma is not needed when the word 'too' is used at the end of the sentence.
### Line 223
1. There are two "_the_"s missing, one before the word "app" and another before the word '"parent's".
2. The word 'objects' is unnecessarily pluralised.
3. The word 'App' is capitalized, even though the variable is named `app` in every code snippet on the page.
### The `app.component()` method is used out of the blue
The docs say that the way to register a component is to create a new variable and add it to the `components` option of the parent, but the next code snippet uses the `app.component()` method without any prior introduction.
### Line 305
There is a missing "_a_" before the word "more".
### Line 55
Added the missing "_the_"s.
### Line 161
Removed the unneeded comma.
### Line 223
Added the two "_the_"s, made "objects" singular and changed the "A" in "App" to lowercase.
### Line 242
Added a small code snippet detailing how to use the `app.component()` method to register a component.
### Line 305
Added the missing "a".


src/guide/introduction.md
```diff
@@ -239,6 +239,20 @@ const app = Vue.createApp({
 app.mount(...)
 ```
 
+Another way to register a component is to use the `component()` method on the `app` object, passing in the name of the component as the first attribute and an options object as the second:
```
NataliaTepluhina, line 242: **suggestion**: I agree that `app.component` looks a bit weird here but this explanation rather belongs to `Component In-Depth` section. What if we refactor the components below to be local components instead of globally registered ones?

NataliaTepluhina (changes requested):

src/guide/introduction.md
scientist27: I do think that's a better idea, since it would make it less confusing for learners while also reducing redundant documentation. I'll make these changes. Thanks!

NataliaTepluhina: Thank you for the super-fast fixes applied! Happy to merge them to the docs 🎉

NataliaTepluhina (approved):

PR #1389, opened 2021-12-24
Add v-bind:key in todo list

wubaiquan:
Add v-bind:key in todo list to avoid error  "Custom elements in iteration require 'v-bind:key' directives"

## Description of Problem
Miss key in list will report  "error  Custom elements in iteration require 'v-bind:key' directives"
It is petty annoy for beginner that their code same as tutorial, but can not run.

## Proposed Solution
Added the v-bind:key in todo list.


NataliaTepluhina:
Thank you for the contribution but `:key` is required by the ESLint plugin, not the Vue itself.

PR #1405, opened 2021-12-28
Update typescript-support.md with GlobalComponents usage

justin-schroeder:
## Description of Problem

Documentation is missing on how to expose type information for globally registered components. The original conversation took place on #3399.

## Proposed Solution

Add a short documentation block immediately after a similar topic: [Augmenting Types for `globalProperties`](https://v3.vuejs.org/guide/typescript-support.html#augmenting-types-for-globalproperties)

### DIFF
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -215,7 +215,7 @@ The above `MyComponent` would not receive [TypeScript editor support](#editor-su
 
 ```vue
 <template>
-  <MyComponent>
+  <MyComponent />
 </template>
 ```
 
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -207,7 +207,7 @@ When [registering a component globally](../api/application-api.html#component) y
 import { createApp } from 'vue'
 import MyComponent from './components/MyComponent.vue'
 
-const app = createApp()
+const app = createApp({})
 app.component('MyComponent', MyComponent)
 ```
 
--- a/src/guide/typescript-support.md
+++ b/src/guide/typescript-support.md
@@ -199,6 +199,38 @@ In order to take advantage of module augmentation, you will need to ensure there
 
 For more information about the `ComponentCustomProperties` type, see its [definition in `@vue/runtime-core`](https://github.com/vuejs/vue-next/blob/2587f36fe311359e2e34f40e8e47d2eebfab7f42/packages/runtime-core/src/componentOptions.ts#L64-L80) and [the TypeScript unit tests](https://github.com/vuejs/vue-next/blob/master/test-dts/componentTypeExtensions.test-d.tsx) to learn more.
 
+### Adding Global Component Types
+
+When [registering a component globally](../api/application-api.html#component) your component’s type definition is not automatically available in your application’s single file component template blocks. For example:
+
+```ts
+import { createApp } from 'vue'
+import MyComponent from './components/MyComponent.vue'
+
+const app = createApp()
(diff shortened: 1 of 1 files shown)


src/guide/typescript-support.md
```diff
@@ -199,6 +199,38 @@ In order to take advantage of module augmentation, you will need to ensure there
...
+
+const app = createApp()
+app.component('MyComponent', MyComponent)
+```
+
+The above `MyComponent` would not receive [TypeScript editor support](#editor-support) in the following single file component:
+
+```vue
+<template>
+  <MyComponent>
```
dammy001, line 218: This should be <MyComponent />

justin-schroeder: Ok, fixed!

src/guide/typescript-support.md
```diff
@@ -199,6 +199,38 @@ In order to take advantage of module augmentation, you will need to ensure there
...
 
+### Adding Global Component Types
+
+When [registering a component globally](../api/application-api.html#component) your component’s type definition is not automatically available in your application’s single file component template blocks. For example:
+
+```ts
+import { createApp } from 'vue'
+import MyComponent from './components/MyComponent.vue'
+
+const app = createApp()
```
skirtles-code, line 210:
```suggestion
const app = createApp({})
```

Not passing anything to `createApp` will lead to an error being thrown.

skirtles-code (commented):
Thanks for the PR!

I don't really use TypeScript, or VSCode, so I'm a bit unsure about what I'm looking at here. `GlobalComponents` doesn't seem to be part of the Vue 3 codebase at present. It looks like https://github.com/vuejs/vue-next/pull/3399 attempts to add it, but that hasn't been merged.

I found a reference to augmenting `GlobalComponents` in the Volar docs:

https://github.com/johnsoncodehk/volar/tree/master/extensions/vscode-vue-language-features#usage

As far as I can tell this is currently just a Volar thing and doesn't necessarily apply to TS users more generally. I'm unclear whether Volar-specific settings belong in the Vue 3 documentation, but if they do then I think there'd need to be more emphasis on this being a Volar thing and not a TS thing.

justin-schroeder:
> I'm unclear whether Volar-specific settings belong in the Vue 3 documentation

I'm unclear on this too actually, and youre right that `GlobalComponents` is not _currently_ referenced in the Vue codebase — although its not even entirely clear that it ever would need to be. That said, since Volar is specifically called out as "the way to do TS" in the Vue 3 docs, it seems relevant to include some kind of mention about how to make global components properly show up with TypeScript support.

For additional context, we had to dig pretty hard to find this content (incorrect info on SO too) to support TypeScript for a Vue plugin library we are developing — where I think people's expectations would be they get editor support for plugin-installed global components. It's definitely an advanced library-author kind of requirement, not unlike the `globalProperties` documentation immediately proceeding it.

Perhaps some additional language around Volar might push this over the edge, but I'd be happy to hear other opinions on where/how this info should be provided to library authors like myself.

skirtles-code:
I've consulted with @johnsoncodehk and @pikax about this. As I understand it, https://github.com/vuejs/core/pull/3399 is actively being worked on and once that is merged it will be added to the documentation.

Until then, Volar has support for a workaround that uses `GlobalComponents`. That is documented in the Volar docs, in the link I posted previously.

Once it's ready it'll need adding to the `main` branch, which has now gone live, replacing `master`. Hopefully the material you added in this PR will prove useful when the time comes to document it.

PR #1407, opened 2021-12-28
Docs:  typo ';' removed

dammy001:


ShenQingchuan:
Nahh... this semicolon is actually needed...

Because the preceding expression/statement is end with `)`

And here we use the 'no-semi' style of JavaScript...

it's a significant hint for JavaScript to know there is a break for these two statements...

or it would be considered as a function call just like `(preceding expression)(args)`.

dammy001:
Yeah, later saw that it's needed

PR #1420, opened 2022-01-05
Update composition-api.md

caozhong1996:


NataliaTepluhina:
@caozhong1996 thank you for creating the PR! Unfortunately, I have to close it because these mappings don't show the correspondence between _Vue 2_ lifecycle hooks and Composition API lifecycle hooks; they show _Vue 3 Options API_ hooks and _Vue 3 Composition API_ hooks. So, the guide was correct and it doesn't require fixes.

caozhong1996:
Thanks for your reply👍

PR #1423, opened 2022-01-08
Add hook `onServerPrefetch`

alexpts:
Add hook `onServerPrefetch` for SSR to table with all hooks


NataliaTepluhina:
@alexpts I believe it's the only place where we mention `serverPrefetch` across all docs so I would wait with merging this PR until we document it better (because this single mention creates a bit of confusion)

skirtles-code:
The new docs will be going live in a couple of weeks, including documentation for `serverPrefetch` and `onServerPrefetch`.

I believe the table has been removed from the new version, so it won't be possible to apply the change made here.

Sorry about that, the change was a good one, it was just unfortunate timing relative to the release of the new docs.

PR #1427, opened 2022-01-09
Update composition-api-setup.md

chenfan0:
From the source code, when the setup function is called, the component instance is already created using the createComponentInstance function.

### DIFF
--- a/src/guide/composition-api-setup.md
+++ b/src/guide/composition-api-setup.md
@@ -103,7 +103,7 @@ We'll explain the role of `expose` shortly.
 
 ## Accessing Component Properties
 
-When `setup` is executed, although the component instance has already been created, you will only be able to access the following properties:
+When `setup` is executed, you will only be able to access the following properties:
 
 - `props`
 - `attrs`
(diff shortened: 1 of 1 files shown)


src/guide/composition-api-setup.md
```diff
@@ -103,7 +103,7 @@ We'll explain the role of `expose` shortly.
 
 ## Accessing Component Properties
 
-When `setup` is executed, the component instance has not been created yet. As a result, you will only be able to access the following properties:
+When `setup` is executed, although the component instance has already been created, you will only be able to access the following properties:
```
NataliaTepluhina, line 106: **suggestion**: I would probably go even further and reduce it to:
```suggestion
When `setup` is executed, you will only be able to access the following properties:
```

NataliaTepluhina (changes requested):

src/guide/composition-api-setup.md
chenfan0: this is a good suggestion

NataliaTepluhina:
@chenfan0 thank you for applying the suggestion! PR looks great to me, let's get it merged 🎉

PR #1431, opened 2022-01-12
Update: correct code annotation error in code blocks

IsLand-x:
## Description of Problem
The ext name of annotation in code block is '.vue' instead of '.js', it's kind of confusing.

## Proposed Solution
Modify the ext name from '.vue' to '.js'.

## Additional Information
None.

### DIFF
--- a/src/api/composition-api.md
+++ b/src/api/composition-api.md
@@ -82,7 +82,7 @@ A component option that is executed **before** the component is created, once th
   With render function:
 
   ```js
-  // MyBook.vue
+  // MyBook.js
 
   import { h, ref, reactive } from 'vue'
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@IsLand-x it's intended to have `.vue` there as an extension. Despite the fact we have a render function, this component could also have a `style` block so I am not sure why we'd need to change the extension to `.js`

IsLand-x:
image


Please double check. According to the doc and other info I learned about, if you use vue sfc (also .vue file), the file must be consisted of `<script>` `<template>` `<style>` and any other blocks. But this file is obviously a js module ( which is using import/export module system ). Are you sure .vue file allows js code exsists outside the `<script>` block?

PR #1453, opened 2022-01-21
fix: Select option background for dark theme

Lanseria:
## Description of Problem

The selection dropdown box is incorrect in the dark theme

## Proposed Solution

should black background


yyx990803:
Can you post a before / after screenshot? I don't see where the current color is "incorrect".

Lanseria:
> Can you post a before / after screenshot? I don't see where the current color is "incorrect".

before

image

after

image

skirtles-code:
This change seems to impact some other elements too, such as the `<textarea>` example.

skirtles-code:
I've attempted something similar in #1587, but without the impact on other elements.

PR #1456, opened 2022-01-23
Update mobile.md

AchrafBardan:
Nativescript vue only supports vue 2.


skirtles-code:
Thanks for the PR!

I agree that something needs to be done to address this. The new docs (available to preview at https://staging.vuejs.org/) remove that page entirely. They'll be going live soon, so I think we can wait for that rather than making changes to the page now.

PR #1457, opened 2022-01-24
Fix API links for headings with badges

skirtles-code:
Some API entries have badges in their headings, such as `<Suspense> [experimental]`. VitePress includes the badge text in the hash, so we need to do the same.

Currently several links on this page are broken:

https://staging.vuejs.org/api/

They are auto-generated based on the page headings, but currently they ignore the badges.

My fix generates two sets of text for each heading , `shortText` and `fullText`. The former is suitable for displaying to the user in the index (without the badge), whereas the former includes the badge so we can generate the correct link.

Badges probably shouldn't be included in the hash. I don't know how feasible it is to change that in VitePress.

### DIFF
--- a/src/api/ApiIndex.vue
+++ b/src/api/ApiIndex.vue
@@ -6,11 +6,11 @@ import { data as apiIndex, APIGroup } from './api.data'
 import { ref, computed } from 'vue'
 
 const query = ref('')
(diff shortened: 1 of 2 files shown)


skirtles-code:
This has been fixed by e6512dc7d52c1d00adca715e371c123f0ef0b46a.

PR #1461, opened 2022-01-26
chore: fix typos

dammy001:


src/api/ApiIndex.vue
```diff
@@ -5,7 +5,7 @@
 import { data as apiIndex, APIGroup } from './api.data'
 import { ref, computed } from 'vue'
 
-const query = ref('')
+const query = ref<string>('')
```
skirtles-code, line 8: I'm unclear what these changes to `ApiIndex.vue` and `api.data.ts` are. They seem to be unrelated to the typos.

src/guide/essentials/forms.md
```diff
@@ -32,7 +32,7 @@ The `v-model` directive helps us simplify the above to:
 In addition, `v-model` can be used on inputs of different types, `<textarea>`, and `<select>` elements. It automatically expands to different DOM property and event pairs based on the element it is used on:
 
 - `<input>` with text types and `<textarea>` elements use `value` property and `input` event;
-- `<input type="checkbox">` and `<input type="radio">` use `checked` property and `change` event;
+- `<input type="checkbox" />` and `<input type="radio" />` use `checked` property and `change` event;
```
skirtles-code, line 35: I don't think these were mistakes. Writing `<input>` without a slash is correct HTML 5 usage.

skirtles-code (changes requested):
Thanks for the PR!

I've provided some feedback on your proposed changes below. I agree that the spelling mistake in `template-refs.md` needs fixing, as does the incorrect quote in `registration.md`. The other changes don't seem so clear cut to me

src/guide/essentials/forms.md
dammy001: I know, it's just for consistency since we're self-closing input elements.

dammy001: @skirtles-code

skirtles-code: If we're going to make them consistent then I think we should consistently not include the slashes.

dammy001: @skirtles-code alright

src/api/ApiIndex.vue
dammy001: @skirtles-code yeah, not related to the typos. It's just basically adding typing to query `ref` and destructuring the parent array on line 74.

src/guide/essentials/forms.md
```diff
@@ -168,13 +168,13 @@ export default {
 <div class="demo">
   <div>Checked names: {{ checkedNames }}</div>
 
-  <input type="checkbox" id="demo-jack" value="Jack" v-model="checkedNames">
+  <input type="checkbox" id="demo-jack" value="Jack" v-model="checkedNames" />
```
yyx990803, line 171: Why are you adding the `/` in some cases but removing them in other places?

dammy001:
@yyx990803 all updated.

src/guide/essentials/template-refs.md
```diff
@@ -3,7 +3,7 @@
 While Vue's declarative rendering model abstracts away most of the direct DOM operations for you, there may still be cases where we need direct access to the underlying DOM elements. To achieve this, we can use the special `ref` attribute:
 
 ```vue-html
-<input ref="input">
+<input ref="input" />
```
skirtles-code, line 6: Shouldn't this slash be removed?

src/api/ApiIndex.vue
skirtles-code: I think I preferred the code for the template of `ApiIndex.vue` the way it was originally. There might be a small amount of duplication, but given the nested loops I think it's clearer to use, e.g. `section.text` rather than just `text`. It helps to differentiate `section.text` from `item.text`.

What is the motivation behind explicitly adding the type information?

skirtles-code:
I'm going to close this.

The most important typos have been fixed in other PRs.

This PR currently has conflicts and can't currently be merged. There is also feedback that hasn't been addressed that would prevent it being merged.

PR #1463, opened 2022-01-26
Update composables.md

selemonjr:
A small error that is present in the new Vue js documentation.

```diff
- Composition API was in part inspired by React hooks.
+ Composition API was in fact inspired by React hooks.
```

### DIFF
--- a/src/guide/reusability/composables.md
+++ b/src/guide/reusability/composables.md
@@ -353,7 +353,7 @@ The recommendation is to use composables when reusing pure logic, and use compon
 
 ### vs. React Hooks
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

What makes you think this is an error? The original wording seems fine to me.

NataliaTepluhina:
I agree with @skirtles-code, the original wording was intentional

selemonjr:
Thank you for all your comments.

> Thanks for the PR!
> 
> What makes you think this is an error? The original wording seems fine to me.

The Vue Composition API and React hooks are both sets of functions that handle state and the reuse of logic in components, so I thought that the Composition API was FULLY inspired by React Hooks.. Thank You for correcting me 😊

PR #1468, opened 2022-01-28
Use createApp for the first code example

garethredfern:
## Description of Problem
The first examples should use `createApp` as further down under https://staging.vuejs.org/guide/scaling-up/ssr.html#client-hydration the docs refer to:

"To mount an app in hydration mode, we need to use createSSRApp() instead of createApp()"

## Proposed Solution

Swap code examples to use `createApp()`

### DIFF
--- a/src/guide/scaling-up/ssr.md
+++ b/src/guide/scaling-up/ssr.md
@@ -58,11 +58,11 @@ Let's take a look at the most bare-bone example of Vue SSR in action.
 
 ```js
 // this runs in Node.js on the server.
-import { createSSRApp } from 'vue'
+import { createApp } from 'vue'
 // Vue's server-rendering API is exposed under `vue/server-renderer`.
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

As I understand it, the line you mentioned is referring to the client-side code. I believe the server-side code needs to use `createSSRApp` from the outset.

garethredfern:
@skirtles-code ahh thank you, sorry I re-read the page and I understand now. Would it be worth saying:

"To mount an app in hydration mode, we need to use createSSRApp() instead of createApp() in the browser."

On line https://github.com/vuejs/docs/blob/main/src/guide/scaling-up/ssr.md?plain=1#L137 (adding "in the browser" to make it clear?

yyx990803:
You should be using `createSSRApp()` in all cases when doing SSR, since the app creation code is shared between server and client.

PR #1471, opened 2022-01-29
fix: Inconsistent word "radius" vs "diameter" in doc

tanquar:
## Description of Problem

Inconsistent word "radius" vs "diameter" in docs - examples - circle.

## Proposed Solution

Use "radius".
The program source seems to handle the interfacing value as radius.
Diameter seems to be generated within the internal logic and it is not obvious to the user.


NataliaTepluhina (approved):
@tanquar good point, thank you 👍🏻

PR #1472, opened 2022-01-29
correct spoken language

tanquar:
## Description of Problem

In docs - about, a partner member who is located in Brazil and working for institutions in Brazil is marked as "Polish" speaker.
Most probably this could be a simple mistake in edit and it should be meant as "Portuguese".

## Proposed Solution

replace Polish -> Portuguese

## Additional Information

I've asked this member directly via email and awaiting for reply.

### DIFF
--- a/src/about/team/members-partner.json
+++ b/src/about/team/members-partner.json
@@ -335,7 +335,7 @@
     "company": "Fatec Taquaritinga",
     "companyLink": "http://www.fatectq.edu.br/",
     "location": "Taquaritinga, Brazil",
-    "languages": ["Polish", "English"],
+    "languages": ["Portuguese", "English"],
     "projects": [
       {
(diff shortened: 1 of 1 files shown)


skirtles-code:
It should be Portuguese. See https://vuejs.org/v2/guide/team.html and https://v3.vuejs.org/community/team.html. The old docs used country codes and `pt` has incorrectly been migrated as Polish.

tanquar:
Thank you for the old links!
Still awaiting from this member, but I can convince the correction.

tanquar:
I received reply from the member and confirmed Portuguese is correct 🇧🇷

ErickPetru:
As I talked to @tanquar, the current published version is correct, but somehow this file have this inconsistency and this merge is needed to avoid confusions on future deploys.

ErickPetru (approved):

PR #1473, opened 2022-01-30
Wrong flight logic

tanquar:
## Description of Problem

In example - flight,
1. condition for showing the warning message was logically flipped.
2. date comparison makes errors around end of month. Note that new Date() is zero-based.

## Proposed Solution

1. Add `!`
2. Use 0-11 for month value

## Additional Information

I also added "value" for option, for ease with translation projects.

### DIFF
--- a/src/examples/src/flight-booker/App/composition.js
+++ b/src/examples/src/flight-booker/App/composition.js
@@ -16,9 +16,9 @@ export default {
 
     function book() {
       alert(
-        `You have booked a ${flightType.value} leaving on ${
-          departureDate.value
-        }${isReturn.value ? ` and returning on ${returnDate.value}` : ``}`
+        isReturn.value
+          ? `You have booked a return flight leaving on ${departureDate.value} and returning on ${returnDate.value}.`
+          : `You have booked a one-way flight leaving on ${departureDate.value}.`
       )
     }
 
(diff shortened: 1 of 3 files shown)


yyx990803:
I'm not sure what you mean by "condition for showing the warning message was logically flipped." - your change to the message seems to do the exact same thing as existing code.

tanquar:
Thank you for your comment.

The logic correction is in
[2ef65ff](https://github.com/vuejs/docs/pull/1473/commits/2ef65ff2ff1db702eb7498b62e14a5e3ccda0742)
where `!this.isReturn` is correct, previously `this.isReturn`.

Previous `this.isReturn` will show the warning message in unwanted situation,
i.e. choosing single trip shows error message for return trip,
and choosing return trip shows no error message for return trip.

Correcting into `!this.isReturn` will fix this,
i.e. choosing single trip shows no error message for return trip,
and choosing return trip shows error message for return trip.

Following pieces of my commits do the same thing. These are adjustment for "ease with translation".
[eeaa0f2](https://github.com/vuejs/docs/pull/1473/commits/eeaa0f2978131e74b4f89e2a5d456c3d91df9163)
[08cd98c](https://github.com/vuejs/docs/pull/1473/commits/08cd98cc514ae73c8c60d5b05895d28eac7d0ad4)

skirtles-code (approved):
Just to confirm, there are two bugs fixed here:

1. Travel dates of `2022/01/31` and `2022/02/02` will incorrectly report that the first date is after the first date. This is because of the missing `m - 1`, leading to `2022/01/31` being treated as March 3rd and `2022/02/02` as March 2nd.
2. The logic for `canBook` is flipped in the Options API example. It is correct for the Composition API example.

PR #1483, opened 2022-02-03
add clarification about type exports in SFCs

theoephraim:
## Description of Problem

Latest vue / volar / vue-tsc complains when exporting additional types from SFCs (even though it worked previously). It seems this may be the intended behaviour (please confirm!) but was a bit unclear in these docs.

## Proposed Solution
This change specifically calls out exporting types and provides a short example.


NataliaTepluhina (approved):
@theoephraim great addition! Thank you 🙇🏻‍♀️

PR #1485, opened 2022-02-04
Options vs Composition API example uniformity fix & typo

xhuljanoduli:
While reading the new docs I saw these examples for options and composition differences, I think it would be clearer if the <style> tag was included in both examples to mimic the previous example and not cause any questions to people that might think that the <styles> have to be written somewhere else.

## Description of Problem

Examples for options API and composition API do not match the previous example

## Proposed Solution

Mimic the same pattern

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -143,8 +143,14 @@ export default {
 </script>
 
 <template>
-  <button @click="increment">count is: {{ count }}</button>
+  <button @click="increment">Count is: {{ count }}</button>
 </template>
+
+<style scoped>
+button {
+  font-weight: bold;
+}
+</style>
 ```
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

I think omitting the `<style>` is intentional. In the context of the API comparison, the `<style>` section just adds extra noise, making it trickier to compare the two APIs. The linked Playground examples also don't include the `<style>` section. The earlier example was trying to demonstrate SFC features, which justified the inclusion of a `<style>` section.

The inconsistency between `count` and `Count` in the text does need fixing, though. It also needs to be consistent with what's in the Playground example. It seems that both of those Playgrounds are currently using `count`, rather than `Count`. I don't have any casing preference myself, so long as we end up with it the same in all 4 places.

yyx990803 (changes requested):
The `<style>` tags are omitted intentionally since they are not relevant for the purpose of the examples. We can keep the change that makes the `count` casing consistent.

skirtles-code:
I've opened #1585 to address the case inconsistency.

PR #1488, opened 2022-02-06
add preinstall hook to allow only `pnpm`

dammy001:


skirtles-code (approved):
Looks good!

For reference, Vite does the same thing to enforce `pnpm` usage:

https://github.com/vitejs/vite/blob/5306234aad7e7432fb55c6033a63c6cf74493c3f/package.json#L16

PR #1489, opened 2022-02-06
docs: fix typo

KimYangOfCat:
## Description of Problem

When I proofread the Chinese translation document, I found that these two lines could be combined and there was no need to split them into two lines.
You can reference [this](https://github.com/vuejs-translations/docs-zh-cn/pull/17)


skirtles-code (approved):
Agreed. The rendered output is the same and having it written across two lines in the markdown is misleading.

Issue #1491, opened 2022-02-07
CSS issue w/ API Preferences question mark


JessicaSachs:
### Summary

Issue w/ the switch-link's `width` being compressed due to container's flex position.

<img width="500" alt="Screen Shot 2022-02-06 at 8 58 12 PM" src="https://user-images.githubusercontent.com/2801156/152713328-9532f12e-751f-4357-b517-7da2f8d3dcef.png">

Steps to reproduce:
1. Open navbar
2. Resize window between 960px and 1440px
3. See the `?` squish to the min content width

PR attached.

PR #1492, opened 2022-02-07
fix: API preferences link is getting squished

JessicaSachs:
## Description of Problem
Screen Shot 2022-02-06 at 8 58 12 PM

Details in #1491 

## Proposed Solution
Fix the width.


NataliaTepluhina (approved):
@JessicaSachs thanks for the fix! Looks much better 👍🏻

PR #1493, opened 2022-02-07
docs: adding information to testing docs

JessicaSachs:
### Docs update for Testing Guide

This is a pretty large change for the testing guide. Here's a list of proposed changes in order of appearance.

- Adds "Why Test?" intro section.
- Adds "Start Testing Early!" tip to "Why Test?" intro.
- Removes "testing concerns" in favor of describing specific issues that each testing type protects against. See the Unit, Component, and E2E sections.
- Adds Unit Testing code example for testing an `increment` function.
- Removes all mentions of "Visual testing" because that's a different testing type w/ a specific meaning (e.g. Percy or Happo snapshots).
- Establishes "Testing Types" terminology throughout the doc.
- Adds section about avoiding mocks in Component Testing section.
- Swaps "Interaction concern" to "Behavioral logic".
- Adds Component Testing example for a Stepper component.
- Recommends against using `.toMatchSnapshot()` as an assertion for a component's functionality.
- Recommends Cypress Component Testing for components that render in the browser. This is Vitest's guidance as well.
- Adds additional information about end-to-end tests. Mainly descriptions of what kinds of bugs end-to-end tests will catch.

### DIFF
--- a/src/guide/scaling-up/testing.md
+++ b/src/guide/scaling-up/testing.md
@@ -30,7 +30,7 @@ We will briefly discuss what each of these are, how they can be implemented for 
 
 Unit tests are written to verify that small, isolated units of code are working as expected. A unit test usually covers a single function, class, composable, or module. Unit tests focus on logical correctness and only concern themselves with a small portion of the application's overall functionality. They may mock large parts of your application's environment (e.g. initial state, complex classes, 3rd party modules, and network requests).
 
-Unit tests will catch issues with a function's business logic and logical correctness.
+In general, unit tests will catch issues with a function's business logic and logical correctness.
 
 Take for example this `increment` function:
 
@@ -67,12 +67,30 @@ describe('increment', () => {
 })
 ```
 
-As mentioned previously, unit testing is typically applied to self-contained business logic, classes, modules, or functions that do not involve UI rendering, network requests, or other environmental concerns.
+As mentioned previously, unit testing is typically applied to self-contained business logic, components, classes, modules, or functions that do not involve UI rendering, network requests, or other environmental concerns.
 
-These are typically plain JavaScript / TypeScript modules unrelated to Vue. In general, writing unit tests in Vue applications does not differ significantly from applications using other frameworks.
+These are typically plain JavaScript / TypeScript modules unrelated to Vue. In general, writing unit tests for business logic in Vue applications does not differ significantly from applications using other frameworks.
+
+There are two instances where you DO unit test Vue-specific features:
+
+1. Composables
+2. Components
+
+### Composables
 
 One category of functions specific to Vue applications are [Composables](/guide/reusability/composables.html), which may require special handling during tests.
 See [Testing Composables](#testing-composables) below for more details.
+
+### Components
+
+A component can be tested in two ways:
+
+1. Whitebox: Unit Testing
+
+Tests that are "Whitebox tests" are aware of the implementation details and dependencies of a component. Components be unit tested by using [`@vue/test-utils`'s](https://test-utils.vuejs.org) `shallowMount` command instead of the `mount` command. You can also make use of the `global.stubs` API. Please read the Vue Test Utils docs for some help on [how to decide](https://test-utils.vuejs.org/guide/advanced/stubs-shallow-mount.html#mount-shallow-and-stubs-which-one-and-when) if you want to Unit Test or Component Test your components. As stated above, unit tests may mock initial state and large parts of your application, when unit testing components, this includes 3rd party components, libraries, and all child components.
+
+2. Blackbox: Component Testing
+Tests that are "Blackbox tests" are unaware of the implementation details of a component. These tests do not mock anything. They render all child components and are considered more of an "integration test" in which Components are the Subject Under Test. We will cover this in the next section.
 
 ### Recommendation
 
@@ -96,7 +114,7 @@ In Vue applications, components are the main building blocks of the UI. Componen
 
 Component tests should catch issues relating to your component's props, events, slots that it provides, styles, classes, lifecycle hooks, and more.
 
-They should not mock child components, but instead test the interactions between your component and its children by interacting with the components as a user would. For example, a component test should click on an element like a user would instead of programmatically interacting with the component.
+Component tests should not mock child components, but instead test the interactions between your component and its children by interacting with the components as a user would. For example, a component test should click on an element like a user would instead of programmatically interacting with the component.
 
 Component tests should focus on the component's public interfaces rather than internal implementation details. For most components, the public interface is limited to: events emitted, props, and slots. When testing, remember to **test what a component does, not how it does it**.
 
--- a/src/guide/scaling-up/testing.md
+++ b/src/guide/scaling-up/testing.md
@@ -94,6 +94,7 @@ A component can be tested in two ways:
 Tests that are "Whitebox tests" are aware of the implementation details and dependencies of a component. They are focused on **isolating** the component under test. These tests will usually involve mocking some, if not all of your component's children, as well as setting up plugin state and dependencies (e.g. Vuex).
 
 2. Blackbox: Component Testing
+
 Tests that are "Blackbox tests" are unaware of the implementation details of a component. These tests mock as little as possible to test the integration of your component and the entire system. They usually render all child components and are considered more of an "integration test". See the [Component Testing recommendations](#component-testing) below.
 
 ### Recommendation
--- a/src/guide/scaling-up/testing.md
+++ b/src/guide/scaling-up/testing.md
@@ -85,16 +85,16 @@ There are two instances where you DO unit test Vue-specific features:
 One category of functions specific to Vue applications are [Composables](/guide/reusability/composables.html), which may require special handling during tests.
 See [Testing Composables](#testing-composables) below for more details.
 
-### Components
+### Unit Testing Components
 
 A component can be tested in two ways:
 
 1. Whitebox: Unit Testing
 
-Tests that are "Whitebox tests" are aware of the implementation details and dependencies of a component. Components mustmbe unit tested by using [`@vue/test-utils`'s](https://test-utils.vuejs.org) `shallowMount` command instead of the `mount` command. You can also make use of the `global.stubs` API. Please read the Vue Test Utils docs for some help on [how to decide](https://test-utils.vuejs.org/guide/advanced/stubs-shallow-mount.html#mount-shallow-and-stubs-which-one-and-when) if you want to Unit Test or Component Test your components. As stated above, unit tests may mock initial state and large parts of your application, when unit testing components, this includes 3rd party components, libraries, and all child components.
+Tests that are "Whitebox tests" are aware of the implementation details and dependencies of a component. They are focused on **isolating** the component under test. These tests will usually involve mocking some, if not all of your component's children, as well as setting up plugin state and dependencies (e.g. Vuex).
 
 2. Blackbox: Component Testing
-Tests that are "Blackbox tests" are unaware of the implementation details of a component. These tests do not mock anything. They render all child components and are considered more of an "integration test" in which Components are the Subject Under Test. We will cover this in the next section.
+Tests that are "Blackbox tests" are unaware of the implementation details of a component. These tests mock as little as possible to test the integration of your component and the entire system. They usually render all child components and are considered more of an "integration test". See the [Component Testing recommendations](#component-testing) below.
 
 ### Recommendation
 
--- a/src/guide/scaling-up/testing.md
+++ b/src/guide/scaling-up/testing.md
@@ -1,32 +1,75 @@
 # Testing
 
+## Why Test?
+
+Automated tests help you and your team build complex Vue applications quickly and confidently by preventing regressions and encouraging you to break apart your application into testable functions, modules, classes, and components. As with any application, your new Vue app can break in many ways, and it's important that you can catch these issues and fix them before releasing.
+
+In this guide, we'll cover basic terminology and provide our recommendations on which tools to choose for your Vue 3 application.
+
+There is one Vue-specific section covering composables. See [Testing Composables](#testing-composables) below for more details.
+
+:::tip Start testing early!
+We recommend you begin writing tests as soon as you can. The longer you wait to add tests to your application, the more dependencies your application will have, and the harder it will be to start.
+:::
+
+## Testing Types
+
+When designing your Vue application's testing strategy, you should leverage the following testing types:
+
+- **Unit**: Checks that inputs to a given function, class, or composable are producing the expected output or side effects.
+- **Component**: Checks that your component mounts, renders, can be interacted with, and behaves as expected. These tests import more code than unit tests, are more complex, and require more time to execute.
+- **End-to-end**: Checks features that span multiple pages and make real network requests against your production-built Vue application. These tests often involve standing up a database or other backend.
+
+Each testing type plays a role in your application's testing strategy and each will protect you against different types of issues.
+
 ## Overview
-
-When it comes to building reliable applications, automated tests can play a critical role in an individual or team's ability to build new features, refactor code, and fix bugs with confidence.
-
-In a Vue application, there are different types of concerns that can be covered by automated tests:
-
-- **Logic**: verify non-UI rendering logic is implemented correctly. This typically involves application-wide business logic (e.g. state management stores) and utility functions.
-
-- **Visual**: given certain input props / state, verify that an isolated component or component tree is rendering the correct visual output.
-
-- **Interaction**: simulate user behavior, such as a click or input, and verify that an isolated component or component tree renders correct, updated output.
-
-- **User Flow**: given a sequence of user interactions that are necessary to complete an actual user task, check whether the application as a whole is working as expected.
-
-For different testing concerns, we need to use different testing techniques:
-
-- **Unit Testing** covers **Logic** tests.
-- **Component Testing** covers **Visual** and **Interaction** tests.
-- **End-To-End (E2E) Testing** covers **User Flow** tests.
 
 We will briefly discuss what each of these are, how they can be implemented for Vue applications, and provide some general recommendations.
 
 ## Unit Testing
 
-Unit tests are written to verify that small, isolated units of code are working as expected. They focus on logical correctness that can be easily validated. The most common example of a unit test is testing whether a function returns expected values based on different input arguments.
+Unit tests are written to verify that small, isolated units of code are working as expected. A unit test usually covers a single function, class, composable, or module. Unit tests focus on logical correctness and only concern themselves with a small portion of the application's overall functionality. They may mock large parts of your application's environment (e.g. initial state, complex classes, 3rd party modules, and network requests).
 
-As mentioned previously, unit testing is typically applied to application-wide business logic or utility functions that do not involve actual UI rendering. These are typically plain JavaScript / TypeScript modules living outside of Vue components. Therefore, writing unit tests in Vue applications does not differ significantly from applications using other frameworks.
+Unit tests will catch issues with a function's business logic and logical correctness.
+
+Take for example this `increment` function:
+
+```js
+// helpers.js
+export function increment (current, max = 10) {
+  if (current < max) {
+    return current + 1
+  }
+  return current
+}
+```
+
+Because it's very self-contained, it'll be easy to invoke the increment function and assert that it returns what it's supposed to, so we'll write a Unit Test.
+
+If any of these assertions fail, it's clear that the issue is contained within the `increment` function.
+
+```js{4-16}
+// helpers.spec.js
+import { increment } from './helpers'
+
+describe('increment', () => {
+  it('increments the current number by 1', () => {
+    expect(increment(0, 10)).toBe(1)
+  })
+
+  it('does not increment the current number over the max', () => {
+    expect(increment(10, 10)).toBe(10)
+  })
+
+  it('has a default max of 10', () => {
+    expect(increment(10)).toBe(10)
+  })
+})
+```
+
+As mentioned previously, unit testing is typically applied to self-contained business logic, classes, modules, or functions that do not involve UI rendering, network requests, or other environmental concerns.
+
+These are typically plain JavaScript / TypeScript modules unrelated to Vue. In general, writing unit tests in Vue applications does not differ significantly from applications using other frameworks.
 
 One category of functions specific to Vue applications are [Composables](/guide/reusability/composables.html), which may require special handling during tests.
 See [Testing Composables](#testing-composables) below for more details.
@@ -49,14 +92,45 @@ Vitest is relatively new and is still undergoing rapid development. While it is 
 
 ## Component Testing
 
-In Vue applications, components are the main building blocks of the UI. Components are therefore the natural unit of isolation when it comes to visual and interaction tests. From a granularity perspective, component testing sits somewhere above unit testing and can be considered a form of integration testing. In some cases we will be testing a single component, but in other cases we could be testing a tree of components to make sure they are integrated correctly.
+In Vue applications, components are the main building blocks of the UI. Components are therefore the natural unit of isolation when it comes to validating your application's behavior. From a granularity perspective, component testing sits somewhere above unit testing and can be considered a form of integration testing. Much of your Vue Application should be covered by a component test and we recommend that each Vue component has its own spec file.
 
-Component tests should focus on the component's public interfaces rather than internal implementation details. Or, in other words, **test what a component does, not how it does it**.
+Component tests should catch issues relating to your component's props, events, slots that it provides, styles, classes, lifecycle hooks, and more.
 
-- **DO**
+They should not mock child components, but instead test the interactions between your component and its children by interacting with the components as a user would. For example, a component test should click on an element like a user would instead of programmatically interacting with the component.
 
-  - For **Visual** tests: assert correct render output based on input props and slots.
-  - For **Interaction** tests: assert correct render updates or emitted events in response to user input events.
+Component tests should focus on the component's public interfaces rather than internal implementation details. For most components, the public interface is limited to: events emitted, props, and slots. When testing, remember to **test what a component does, not how it does it**.
+
+**DO**
+
+- For **Visual** logic: assert correct render output based on inputted props and slots.
+- For **Behavioral** logic: assert correct render updates or emitted events in response to user input events.
+
+In the below example, we demonstrate a Stepper component that has a DOM element labeled "increment" and can be clicked. We pass a prop called `max` that prevents the Stepper from being incremented past `2`, so if we click the button 3 times, the UI should still say `2`.
+
+We know nothing about the implementation of Stepper, only that the "input" is the `max` prop and the "output" is the state of the DOM as the user will see it.
+
+```js
+const valueSelector = '[data-testid=stepper-value]'
+const buttonSelector = '[data-testid=increment]'
+
+mount(Stepper, {
+  props: {
+    max: 1
+  }
+})
+
+// Each of these commands does implicit assertions
+// that the button and values are visible
+// and that the button can be clicked.
+cy.get(valueSelector)
+  .should('be.visible')
+  .and('contain.text', '0')
+  .get(buttonSelector)
+  .click()
+  .get(valueSelector)
+  .should('contain.text', '1')
+  .click() // Should still be "1" because of the max prop
+```
 
 - **DON'T**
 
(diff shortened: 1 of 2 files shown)


src/guide/scaling-up/testing.md
```diff
@@ -49,30 +92,75 @@ Vitest is relatively new and is still undergoing rapid development. While it is
...
+In Vue applications, components are the main building blocks of the UI. Components are therefore the natural unit of isolation when it comes to validating your application's behavior. From a granularity perspective, component testing sits somewhere above unit testing and can be considered a form of integration testing. Much of your Vue Application should be covered by a component test and we recommend that each Vue component has its own spec file.
+
+Component tests should catch issues relating to your component's props, events, slots that it provides, styles, classes, lifecycle hooks, and more.
 
-Component tests should focus on the component's public interfaces rather than internal implementation details. Or, in other words, **test what a component does, not how it does it**.
+They should not mock child components, but instead test the interactions between your component and its children by interacting with the components as a user would. For example, a component test should click on an element like a user would instead of programmatically interacting with the component.
```
lmiller1990, line 99: :tada:

JessicaSachs: The advice against `.toMatchSnapshot` is personally my fave. 😂

src/guide/scaling-up/testing.md
```diff
@@ -49,30 +92,75 @@ Vitest is relatively new and is still undergoing rapid development. While it is
...
+// that the button and values are visible
+// and that the button can be clicked.
+cy.get(valueSelector)
+  .should('be.visible')
+  .and('contain.text', '0')
+  .get(buttonSelector)
+  .click()
+  .get(valueSelector)
+  .should('contain.text', '1')
+  .click() // Should still be "1" because of the max prop
```
lmiller1990, line 132: I wonder if we want to at least include an example of how it looks with Test Utils (in addition), since that's something a lot of people are familiar with at this point, and most of the information you'll find on Google will reference the VTU API.

src/guide/scaling-up/testing.md
```diff
@@ -49,30 +92,75 @@ Vitest is relatively new and is still undergoing rapid development. While it is
...
+
+- For **Visual** logic: assert correct render output based on inputted props and slots.
+- For **Behavioral** logic: assert correct render updates or emitted events in response to user input events.
+
+In the below example, we demonstrate a Stepper component that has a DOM element labeled "increment" and can be clicked. We pass a prop called `max` that prevents the Stepper from being incremented past `2`, so if we click the button 3 times, the UI should still say `2`.
+
+We know nothing about the implementation of Stepper, only that the "input" is the `max` prop and the "output" is the state of the DOM as the user will see it.
+
+```js
+const valueSelector = '[data-testid=stepper-value]'
```
lmiller1990, line 113: I think we should include "this example is written using Cypress" before we present it, at the very least, since the style is really different from the unit test example (eg the assertions) and that might be a little confusing/jarring.

src/guide/scaling-up/testing.md
JessicaSachs: Yeah I'm feeling weird about this, too. I actually had it written with Testing Library at first... but then I felt weird about recommending Cy, but not using Cy in the examples.

src/guide/scaling-up/testing.md
JessicaSachs: 100000% if we keep it as a Cy example, we should annotate it. It's disjointed and jarring like "Wtf is this syntax".

lmiller1990 (commented):
Cool, just some minor thoughts. I like this guide!

src/guide/scaling-up/testing.md
lmiller1990: I think this is a good opportunity to focus 'understand the concept, not the syntax'. Although not the most idiomatic Cypress, we can do something like:

___

Here's an example using Vue Test Utils, a low-level testing library that powers libraries like Testing Library and Cypress:

```js
const valueSelector = '[data-testid=stepper-value]'
const buttonSelector = '[data-testid=increment]'

const wrapper = mount(Stepper, {
  props: {
    max: 1
  }
})

// Each of these commands does implicit assertions
// that the button and values are visible
// and that the button can be clicked.
expect(wrapper.get(valueSelector).text()).toBe('0')
wrapper.get(buttonSelector).trigger('click')
expect(wrapper.get(valueSelector).text()).toBe('1')
wrapper.get(buttonSelector).trigger('click')

// Should still be "1" because of the max prop
expect(wrapper.get(valueSelector).text()).toBe('1')
```

This same test can be made more readable using Cypress; notice the ideas and concepts are the same; we are testing in a similar manner to a user, by clicking and observing the HTML. The only difference is the syntax; your testing library should not change the way you choose to test, just the syntax.

```js
mount(Stepper, {
  props: {
    max: 1
  }
})

// Each of these commands does implicit assertions
// that the button and values are visible
// and that the button can be clicked.
cy.get(valueSelector).should('contain.text', '0')
cy.get(buttonSelector).click()
cy.get(valueSelector).should('contain.text', '1')
cy.get(buttonSelector).click()

// Should still be "1" because of the max prop
cy.get(valueSelector).should('contain.text', '1'))
```

Far from perfect but this is what I was thinking. What do you think? It's a bit wordy and verbose, but I like the focus on ideas, not syntax and specific library APIs.

src/guide/scaling-up/testing.md
```diff
@@ -1,32 +1,75 @@
...
+Automated tests help you and your team build complex Vue applications quickly and confidently by preventing regressions and encouraging you to break apart your application into testable functions, modules, classes, and components. As with any application, your new Vue app can break in many ways, and it's important that you can catch these issues and fix them before releasing.
 
-When it comes to building reliable applications, automated tests can play a critical role in an individual or team's ability to build new features, refactor code, and fix bugs with confidence.
+In this guide, we'll cover basic terminology and provide our recommendations on which tools to choose for your Vue 3 application.
 
-In a Vue application, there are different types of concerns that can be covered by automated tests:
+There is one Vue-specific section covering composables. See [Testing Composables](#testing-composables) below for more details.
 
-- **Logic**: verify non-UI rendering logic is implemented correctly. This typically involves application-wide business logic (e.g. state management stores) and utility functions.
+:::tip Start testing early!
```
NataliaTepluhina, line 11: **suggestion**: we try to avoid "special" blocks like tips or warnings as much as possible as we consider them destructive for the reading flow. Is there any chance we could omit the `tip` part while keeping the valuable suggestion?

src/guide/scaling-up/testing.md
```diff
@@ -1,32 +1,75 @@
...
+:::
 
-- **Visual**: given certain input props / state, verify that an isolated component or component tree is rendering the correct visual output.
+## Testing Types
 
-- **Interaction**: simulate user behavior, such as a click or input, and verify that an isolated component or component tree renders correct, updated output.
+When designing your Vue application's testing strategy, you should leverage the following testing types:
 
-- **User Flow**: given a sequence of user interactions that are necessary to complete an actual user task, check whether the application as a whole is working as expected.
+- **Unit**: Checks that inputs to a given function, class, or composable are producing the expected output or side effects.
```
NataliaTepluhina, line 19: **note**: I am not quite sure we should have a separation here between unit and component testing as component tests are considered a part of unit testing.

src/guide/scaling-up/testing.md
```diff
@@ -1,32 +1,75 @@
...
+- **Component**: Checks that your component mounts, renders, can be interacted with, and behaves as expected. These tests import more code than unit tests, are more complex, and require more time to execute.
+- **End-to-end**: Checks features that span multiple pages and make real network requests against your production-built Vue application. These tests often involve standing up a database or other backend.
```
NataliaTepluhina, line 21: **suggestion**: I believe we're missing an `Integration testing` here

src/guide/scaling-up/testing.md
NataliaTepluhina: **question**: this part about not mocking child components is not completely clear to me. Should we avoid shallow mounting that replaces child components with stubs or we're speaking about something different here?

src/guide/scaling-up/testing.md
```diff
@@ -49,30 +92,75 @@ Vitest is relatively new and is still undergoing rapid development. While it is
...
+
+Component tests should catch issues relating to your component's props, events, slots that it provides, styles, classes, lifecycle hooks, and more.
 
-Component tests should focus on the component's public interfaces rather than internal implementation details. Or, in other words, **test what a component does, not how it does it**.
+They should not mock child components, but instead test the interactions between your component and its children by interacting with the components as a user would. For example, a component test should click on an element like a user would instead of programmatically interacting with the component.
 
-- **DO**
+Component tests should focus on the component's public interfaces rather than internal implementation details. For most components, the public interface is limited to: events emitted, props, and slots. When testing, remember to **test what a component does, not how it does it**.
```
NataliaTepluhina, line 101: **suggestion**: I believe we are missing side effects here like dispatching Vuex actions etc

src/guide/scaling-up/testing.md
NataliaTepluhina: I would go even further and suggest rewriting the test to Jest/having Jest code as an alternative here

src/guide/scaling-up/testing.md
```diff
@@ -49,30 +92,75 @@ Vitest is relatively new and is still undergoing rapid development. While it is
...
+```
 
 - **DON'T**
 
   Assert the private state of a component instance or test the private methods of a component. Testing implementation details makes the tests brittle, as they are more likely to break and require updates when the implementation changes.
 
   The component's ultimate job is rendering the correct DOM output, so tests focusing on the DOM output provide the same level of correctness assurance (if not more) while being more robust and resilient to change.
 
-  If a method needs to be tested, extract it into a standalone utility function and write a dedicated unit test for it. If it cannot be extracted cleanly, it should be tested as a part of an interaction test that invokes it.
+  Rely exclusively on snapshot tests. Asserting HTML strings does not describe correctness. Write tests with intentionality.
```
NataliaTepluhina, line 141: **praise**: great point that if often overlooked in tests!

NataliaTepluhina (commented):
@JessicaSachs great work! I've added a few comments here and there

src/guide/scaling-up/testing.md
xanf, line 99: This line feels very confusing for me. Description of **Unit**-tests never explicitly mentions that components could also be tested in this way (by using shallow rendering). So basically `shallow` style of testing is silently ignored

Additionally this requirement feels extremely **strict**  - considering this guide will be an intro to Vue testing for many people, it would be great to be both actionable and not biased towards such things

src/guide/scaling-up/testing.md
xanf: Any reason why this is written in Cy? Considering `vitest` as a first option why not start with "just pure vitest" - this will require people not to learn any additional syntax (Cy as an example) just to understand the test

src/guide/scaling-up/testing.md
xanf: @lmiller1990 exactly described my opinion which I mentioned one comment above. As I said - we can even go further and throw away everything, except a test runner (`vitest`) to help this example to be more focused

src/guide/scaling-up/testing.md
```diff
@@ -49,30 +92,75 @@ Vitest is relatively new and is still undergoing rapid development. While it is
...
+
+  If a method needs to be tested thoroughly, consider extracting it into a standalone utility function and write a dedicated unit test for it. If it cannot be extracted cleanly, it may be tested as a part of a component, integration, or end-to-end test that covers it.
 
 ### Recommendation
 
-- [Vitest](https://vitest.dev/) or the other unit testing frameworks mentioned above can be used for component testing by simulating the DOM in Node.js. See [Adding Vitest to a Project](#adding-vitest-to-a-project) for more details.
+- [Vitest](https://vitest.dev/) for components or composables that render headlessly. (e.g. the [`useFavicon`](https://vueuse.org/core/useFavicon/#usefavicon) function in VueUse.
+
+- [Cypress Component Testing](https://on.cypress.io/component) for components whose expected behavior depends on properly rendering styles or triggering native DOM events.
```
xanf, line 149: Sorry, I'm absolutely not familiar with how "recommendations" are being determined, but is there any reasoning to move recomendation away from testing-library for example? 

I won't mention all pros and cons of cypress / cypress component testing (like running time, etc.) because I really believe everyone is aware of that but this was like a main surprise of  this pull request for me - jumping from "pure vitest" to "full browser testing".

I bet this is a huge discussion point and not feeling safe to do such blind recommendation considering all possible options (promoting cypress as recommended way)

src/guide/scaling-up/testing.md
JessicaSachs: Yes, the guidance is to avoid `shallowMount`.

src/guide/scaling-up/testing.md
JessicaSachs: Jest and Vitest are the test runners and aren't really "visible" in this code example.

Vitest supports both Mocha and Jasmine matchers.

The debate is really:

<details>
<summary>What component mounting strategy are we recommending?</summary>

### Vue Test Utils

```js
const wrapper = mount(Stepper, {
  props: {
    max: 10
  }
})

```

### Cypress

```js
mount(Stepper, {
  props: {
    max: 10
  }
})
```

### Testing Library

```js
render(Stepper, {
  props: {
    max: 10
  }
})
```

</details>

<details>
<summary>What will we recommend for interacting with components? (Vue Test Utils, Testing Library, Cypress)</summary>

### Vue Test Utils

```js
await wrapper.get('[data-testid=increment]').trigger('click')
expect(wrapper.get('[data-testid=stepper]').text()).toEqual('1')
```

### Cypress

```js
cy.get('[data-testid=increment]').click()
cy.get('[data-testid=stepper]').should('contain.text', '1')
```

### Testing Library

```js
const button = screen.getByTestId('increment')
await fireEvent.click(button)
expect(screen.queryByText('1')).toBeTruthy()
```


</details>


<details>
<summary>What assertion library are we recommending? (Jasmine, Mocha, Cypress's wrapped Mocha)</summary>

### Jasmine

```js
expect('whatever').toBeSomething()
```

### Chai

```js
expect('whatever').to.be.something
```

### Cypress's Chai

```js
cy.get('button').should('be.something')
```


</details>

In this PR we recommend Vitest as the test runner for headless components and unit tests, which comes with both Mocha and Jasmine assertions (Jasmine is camel-case, Mocha is dots) and we recommend Cypress for mounting and clicking on headed components.

If we don't want users using `@vue/test-utils` directly, what syntax should we pick?

src/guide/scaling-up/testing.md
JessicaSachs: I don't believe we should encourage interacting with Vuex actions from within component tests. That's an implementation detail of the component.

A litmus test: If you replaced Vuex with Pinia, if the expected behavior of the component is the same, why should your test change?

src/guide/scaling-up/testing.md
JessicaSachs: Yeah it drives me nuts, lol. I remember auditing some major component libraries in our ecosystem and all of their tests were a render and a snapshot 😬.

src/guide/scaling-up/testing.md
JessicaSachs: > Sorry, I'm absolutely not familiar with how "recommendations" are being determined, but is there any reasoning to move recomendation away from testing-library for example?

Swapping to Cypress as the recommendation was the main point of the Pull Request and will (rightly) be a point of debate.

The short of it is that:

Vitest recommends Cypress for testing components that are rendered in the browser, so it feels weird to recommend people use Vitest and then Vitest says "No, use Cypress for testing your headed components". 🔁

src/guide/scaling-up/testing.md
JessicaSachs: Oh, we can add copy about `shallowMount` into Unit Tests if we'd like. I didn't mention it, but not on purpose.

src/guide/scaling-up/testing.md
JessicaSachs: I think there's two ways to test components:

1. By whitebox testing the component, where the `wrapper` is the subject under test. This is what I'd consider Unit Testing your components.
2. By blackbox testing the component, where the DOM is the subject under test. Testing Library and Cypress exclusively promote this blackbox-style testing (which we're referring to as "Component Testing")

src/guide/scaling-up/testing.md
JessicaSachs: After reading your other comments: If we're considering a section on "Unit Testing components", then executing Vuex actions makes perfect sense.

JessicaSachs:
Lots of comments (this is good!!!!).

So far what I'm hearing:

1. Representation of Unit Testing a component using `shallowMount` in Vitest by using `@vue/test-utils` directly. (Testing Library's philosophy explicitly against all mocking).
2. Addition of Integration Testing.
3. Rework the Component example. Perhaps we should remove it altogether? I think all choices are somewhat confusing in their own way.

src/guide/scaling-up/testing.md
JessicaSachs: Maybe we do that as a follow-up. I really do want to include the Testing Pyramid to explain how speed/complexity are all relative, but I don't want to rework the sections _again_ right now.

JessicaSachs:
How's this?

Screen Shot 2022-02-07 at 9 22 49 AM

src/guide/scaling-up/testing.md
JessicaSachs: I added _all the things_ in a tabbed component 😼. This should be helpful.

Screen Shot 2022-02-07 at 9 39 20 AM

src/guide/scaling-up/testing.md
NataliaTepluhina: I believe VTU (at least at the moment) is the most used approach for component testing across the user base, and don't think we should be limiting users who prefer `shallowMount` as it's a valid testing approach. I have read Kent's article before and I still don't completely agree with this article personally, and I know I am not the only one.

TL;DR: I  would prefer core docs not to dictate to users which approach they should take. We should allow people to choose their own approach.

src/guide/scaling-up/testing.md
NataliaTepluhina: I still disagree with this definition. VTU allows you to perform proper blackbox testing if you don't assert `wrapper.vm`  properties. Hence, the line between component and unit testing is still blurred for me.

src/guide/scaling-up/testing.md
NataliaTepluhina: @JessicaSachs replace both Vuex/Pinia with calling a side effect like Sentry or external helper. I click the button, I want to make sure my component fires this imported helper, how do I test this?

src/guide/scaling-up/testing.md
JessicaSachs: I think you misunderstood me. You are correct.

Blackbox testing: DOM is the subject under test. VTU (but w/o accessing wrapper.vm or major stubbing), VTL (does not let you access internals), Cy (does not let you access internals)

Whitebox testing: Wrapper is the subject under test, DOM is _sometimes_ used. Lots of stubbing. VTU supports whitebox via Shallow Mount. Not possible to perform with Cy or VTL.

src/guide/scaling-up/testing.md
JessicaSachs: There's two solutions: blackbox + whitebox.

**Blackbox testing:**
When you open up your browser and you're developing your app, how do you know if Sentry is working properly? I imagine you look at your network tab.

Use Mock Service Worker or cy.intercept(). This allows you to catch if your Sentry 3rd party library has broken or you are not using it correctly. It couples you to the "public API" of Sentry and requires you to setup an environment that allows network stubbing, but it does not couple you to the specific files, modules, methods used.

**Whitebox testing:**
Use module/file mocking, stubs with a spy, etc. This couples you to the way your source code is written and the library you use. It covers up possible integration failures. BUT it requires much less infrastructure and depending on the library, will make your tests faster/simpler.

JessicaSachs: This is a really helpful discourse for me, btw. I hope it's not frustrating for you, because it's very helpful for me to talk through these thought experiments. Thank you. <3

JessicaSachs:
I was thinking more regarding the `shallowMount` discussion.

I realized that Svelte and React exclusively recommend full `mount`/`render` instead of `shallowMount`.

https://reactjs.org/docs/testing.html
https://svelte.dev/faq#how-do-i-test-svelte-apps

I did re-write the Unit Testing section to cover `shallowMount`, but I believe we should remove it. Enzyme popularized `shallowMount` 4 years ago, but the majority of modern resources recommend against mocking the component hierarchy.

I do not think we should recommend it or mention it because the only way you can `shallowMount` is via VTU, and in the Component Testing section we tell users not to use VTU unless they have advanced components. It feels weird to mention something we're telling you not to use.

I'm happy to leave it in, I guess, but I feel that it's unnecessary given that the other frameworks do not recommend `shallowMount`.

xanf:
@JessicaSachs sorry, maybe I'm missing somethif, but I can't find any guidance **against**mocking trees in both links you've provided. React router ne even mentions how one could achieve this, and svelte uses "mount" as an operation of putting xomponnent into dom. 

Could you please point me to their opinion?

JessicaSachs:
No problem. In general, there will always be a layer of mocking. I mean, JSDom mocks the browser, yknow? The idea is that it is kept to a minimum and is not the default behavior.

>  React router ne even mentions how one could achieve this

React recommends React Testing Library with Jest. React Testing Library recommends using a real React Router in their examples. https://testing-library.com/docs/example-react-router/

React Testing Library author (Kent) [recommends against](https://kentcdodds.com/blog/why-i-never-use-shallow-rendering) mocking anything in the component hierarchy. That's the opinion of the React Testing Library, and is therefor the best practice for testing React components.

> svelte uses "mount" as an operation of putting xomponnent into dom.

Don't they use `render` via Svelte Testing Library? Either way, their recommendation is a blackbox test. You mount the component to the DOM and assert that the DOM contains the correct state. You don't have access to the internals of the component instance (a svelte-equivalent of `wrapper.vm`).

Both React and Svelte recommend blackbox style testing and a full mount with as little component hierarchy mocked as possible.

xanf:
@JessicaSachs sorry, it seems we're reading the same text, but understand it differently

I'm aware of Kent's position on mocking vs full-tree, but I can't say that for example svelte or even react socs says "as little as possible" (testing library obviously says that)

Based on my experience of maintaining gitlab as probsbly the biggest opensource app which is using vue - shallow-mounting was a huge time saver suring vue2 to vue3 migration  wheb 95% of mount tests are simply red due to nuances of compat build ajd focused shallow tests allowed me to precisely determine reason of failure. That's why I'm so concerned about blindly going down the full mount path

JessicaSachs:
I'm really happy Gitlab had a good experience with VTU's shallowMount. That's awesome. It's a great success story, and I'm glad that you have the debugging and development experience to use shallow mounting to its full potential.

One point of clarification: Please understand that nothing I am proposing is 
 done so blindly. I work full-time in the testing space. I've been building Open Source test runners full-time for years and I've worked as a Lead QA Eng building test infrastructure for a long time before that. I've worked with users and taught workshops. All of this experience gives me very good market awareness in the space. I rely on this experience when we discuss testing strategies to promote in the ecosystem. Nothing is done blindly.

---

I think we've gone past the point of being productive.

Let's call it, include `shallowMount`, and revisit this later if we'd like. There's really no harm in including it and it's very important to you, and not too important to me.

I had already updated the documentation to include it. I was mainly interested in the conversation.

---

What else is left? How do people feel about the Testing Syntax Switcher?

xanf:
@JessicaSachs thank you for your cooperation! I really feel proposing all options to the users is really important for a health of community

Regarding your clarification: please could tou assume the same level of good intentions and awareness from others. I'm frontend testing expert in GitLab for years, i actively contribute to @vue/test-utils (both versions) and last year more than 2000 people total attended my workshops on testing and vue testing. I believe this also gives me market awareness and understanding people's pain. 

I'm more than welcome for any diwcussions and I always have strong opinions weakly held, but the last message was not the friendliest one and feels more like social proofing than bringing arguments to discussion. I really believe this was not your intent and maybe the reason I read it in such way are just cultural differences, just wanted to give an honest feedback

JessicaSachs:
💔 I re-wrote that message a few times and strongly considered omitting talking about my background because I felt like it detracted from the convo, but wasn't sure how to best communicate that I **do** think about these things all the time.

I don't question your experience and expertise at all, and I'm really happy that you're willing to engage with me so deeply on this topic. A lot of people aren't able to and it's pretty refreshing (even during semi-conflict).

src/guide/scaling-up/testing.md
NataliaTepluhina: @JessicaSachs I like the discussions 😅 Sounds like a great summary! I am still not sure whether we should say that component's side effect is a part of the public component interface (because it's an internal part) but I still think we need to test these behaviors no matter what test runner we are using, so they worth a mention. What do you think?

bencodezen:
Thanks everyone for the spirited discussion! Clearly, people are coming from different experiences and there are a lot of different opinions in terms of what is best. That said, I think It's great to see everyone being able to share their views on "best practices" while still trying to keep in mind what's best for the community as a whole.

In terms of next steps, what do we think we need to wrap up this PR? I'd love to see us merge these changes so that people can benefit from all of this sooner rather than later. And then as the community establishes more patterns / best practice / guidance on testing, we can certainly iterate on it with another PR!

JessicaSachs:
Thanks Ben, you're awesome.

I think this is where we left off after the first review and I applied the relevant feedback.

**What I need**:
* Confirmation that we can skip Integration Testing for now (No. 2)
* Approval for the tabbed mount helpers (No. 3)

> So far what I'm hearing:
> 
> 1. Representation of Unit Testing a component using `shallowMount` in Vitest by using `@vue/test-utils` directly. (Testing Library's philosophy explicitly against all mocking... I guess this feels a _little_ awkward because we're calling Vue Test Utils a "low level" library?)
We finished this discussion and I rewrote the Unit Testing section and Component Testing section

> 2. Addition of Integration Testing. (another PR maybe?)
Another PR please.

> 3. Rework the Component example. Perhaps we should remove it altogether? I think all choices are somewhat confusing in their own way.
I created UI/UX to solve this. Please let me know if the tabbed "switcher" is good enough for now?
https://github.com/vuejs/docs/pull/1493#issuecomment-1031523097

image

NataliaTepluhina:
@bencodezen @JessicaSachs 

I believe multiple approaches are a good idea. However, I'd prefer pure VTU example over testing library there.

I'll bring a few more points within one comments so that the discussion is not spread:

- > Tests that are "Whitebox tests" are aware of the implementation details and dependencies of a component. Components be unit tested by using [@vue/test-utils's](https://test-utils.vuejs.org/) shallowMount command instead of the mount command.

  I believe this definition is not completely correct. We can perform whitebox testing with `mount` and `shallowMount` both, if we're checking internal component properties:

  ```js
  expect(wrapper.vm.something).toBe('Hello World')
  ```
  This (IMO) should be discouraged disregarding the mounting method.

  At the same time, we can perform blackbox testing with both `mount` and `shallowMount`:

  ```js
  expect(wrapper.find('#some-html-element').text()).tobe('Hello World')
  ```

  This does not correlate directly with the mounting method.

-   >  Vitest is relatively new and is still undergoing rapid development

  In all fairness, I am not sure that the only default recommended option should be something under active development 🤔 

- > We recommend using @testing-library/vue for testing components in applications, as its focus aligns better with the testing priorities of applications. Use @vue/test-utils only if you are building advanced components that require testing Vue-specific internals.

  It's not completely clear to me why the second case is more advanced 🤔 Maybe we need some examples here to understand why can't we achieve the same goal with VTU and why we recommend `@testing-library/vue` over it

Honestly, after trying to fit so many points within one guide, I start thinking that maybe we should limit this section to recommended testing frameworks/libraries with a short explanation and then let users read the comprehensive guides on the mentioned libraries docs. But this is just my 2 cents

lmiller1990:
I looked at how React solves this problem. They have three pages:

1. https://reactjs.org/docs/testing.html
2. https://reactjs.org/docs/testing-recipes.html
2. https://reactjs.org/docs/testing-environments.html

1. Testing - extremely short. "Here are some tools that we know people use and enjoy, check their docs". We'd link to VTU, Cypress, Testing Library. It's about a page long.
2. Recipes. They have some short examples, without too much ceremony. Basically they go "here is a scenario", "here's an example test". This is where we'd have the tabs, like we do for Options and Composition, if we want to include multiple frameworks. If we do this, I'd not include too much philosophy and opinions. You could link to external blog posts, talks, etc, if you want.
3. Environments. This page does not even say the world "React", you could literally copy and paste it here and it would be the same.

I wonder if we can do something similar here? The way they've done it in React does not **recommend** too specifically, it just presents information (kind of like the rest of the Vue docs - we don't recommend Options or Composition, we just list some of the trade-offs and leave the philosophy with the user.

JessicaSachs:
@lmiller1990 FYI this is the first paragraph in the first page of the react docs about testing.

Screen Shot 2022-02-12 at 7 03 53 PM
 
Like React, Vue does have recommendations. We **do** recommend Vitest as the default test runner for Vue and will continue to do so, because it's what makes sense for our recommended bundler: Vite.

Removing **all** recommendations and thought leadership will push the work onto the user to sift through 3rd party blog posts.

So far, we've added inclusive wording and a non-biased writeup of Whitebox/Blackbox testing and how to unit test components... I feel like we've rewritten the doc such that we're presenting all the information while actually writing content instead of just presenting links.

Aside from using Cypress's code example by default (I think it's out of place on its own), I'm happy with the content. Can we add the tabbed UI for the diff frameworks (just like we have for Composition vs Options) and merge?

NataliaTepluhina:
@JessicaSachs I believe the points from https://github.com/vuejs/docs/pull/1493#issuecomment-1033020045 are still not addressed (especially the connection between shallow/non-shallow mounting and whitebox/blackbox testing). Addressing this and adding tabbed UI remove the last blockers for merging the PR 👍🏻

JessicaSachs:
Okay I thiiiink I did it. Just give edits to the copy if it's not right -- I tried to adjust correctly.

Also, the Testing API Switcher should probably just be a Code Group inside of https://github.com/vuejs/theme. It's looking pretty good though, I think!

JessicaSachs:
Screen Shot 2022-02-13 at 3 30 40 AM
Screen Shot 2022-02-13 at 3 31 01 AM
Screen Shot 2022-02-13 at 3 31 28 AM

PR #1494, opened 2022-02-07
fix: special sponsors link on homepage is broken

JessicaSachs:
User reported that the Special Sponsors link for hbuilderx is broken. This fixes that.


JessicaSachs:
Fixed by #1495

PR #1501, opened 2022-02-08
Use `globals` setting in testing setup

paparent:
I've been starting a new project and follow this docs. Here's some suggestion:

Since https://github.com/vitest-dev/vitest/pull/565, `global` is deprecated. I propose a section to add a config for TypeScript.


NataliaTepluhina (approved):
@paparent great additions, thank you!

PR #1506, opened 2022-02-09
Update template-syntax.md

zhangenming:
add `v-text` for Text Interpolation section
Use the same name"Mustache"(There are several different ones in the text, I don't know if they are intentional);
add code `const rawHtml = '<span style="color: red">This should be red.</span>'`

### DIFF
--- a/src/guide/essentials/template-syntax.md
+++ b/src/guide/essentials/template-syntax.md
@@ -14,14 +14,38 @@ The most basic form of data binding is text interpolation using the "Mustache" s
 <span>Message: {{ msg }}</span>
 ```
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

I believe `v-text` is intentionally omitted from this page. It's only really used for handling edge cases and not something that newcomers needs to meet in the initial guides.

The `{{ ... }}` feature is known as 'text interpolation'. That is used as the heading on the section. The term 'mustache' refers specifically to the syntax. That same name is used in other frameworks and templating engines that use a similar syntax. Both terms get used frequently within the community, so there's no need to enforce consistency on one term or the other.

PR #1507, opened 2022-02-10
add codestyle to some words

wxsms:
## Description of Problem

we have some problems while translating this article to other language. It is, if a word (like 'props') means the `props` option in code, then it should have a code style, so we can keep the format of it (like 'props' with 's'). otherwise, by our translating guideline, we will omit the 's', so 'props' become 'prop'.

Anyway, it's just my reason to make this change. If this looks not good, please ignore it. And we will figure out a better solution.

ref https://github.com/vuejs-translations/docs-zh-cn/pull/11

### DIFF
--- a/src/guide/typescript/options-api.md
+++ b/src/guide/typescript/options-api.md
@@ -6,7 +6,7 @@
 While Vue does support TypeScript usage with Options API, it is recommended to use Vue with TypeScript via Composition API as it offers simpler, more efficient and more robust type inference.
 :::
 
-## Typing Component `props`
+## Typing Component Props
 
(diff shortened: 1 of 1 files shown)


src/guide/typescript/options-api.md
```diff
@@ -30,7 +30,7 @@ export default defineComponent({
 })
 ```
 
-However, the runtime props options only support using constructor functions as a prop's type - there is no way to specify complex types such as objects with nested properties or function call signatures.
+However, the runtime `props` options only support using constructor functions as a prop's type - there is no way to specify complex types such as objects with nested properties or function call signatures.
```
yyx990803, line 33: I think this is okay, but we should keep the original the headers.

PR #1514, opened 2022-02-11
Use counter variable, not undefined state

zangab:
## Description of Problem
The variable `state` is not defined in the context. On top you can find the `counter` variable though.
It's kind of irritating.

## Proposed Solution
Change variable.


NataliaTepluhina (approved):
@zangab nice catch! Thank you for fixing it 👍🏻

PR #1518, opened 2022-02-11
Add info about v-bind in CSS under CSSProperties

pleek91:
When using SFC it's often not necessary to do type augmentation if v-bind in CSS is used.

## Description of Problem
Adding custom properties in a template causes a type error because custom properties are not accounted for in `CSSProperties`. Searching how to fix that error frequently leads to the docs on [CSSProperties](https://vuejs.org/api/utility-types.html#cssproperties) and how to augment the type.
 
## Proposed Solution
Using `v-bind` in CSS is often the better solution or is at least relevant so linking to that documentation will help convey another solution to fixing the type error.


NataliaTepluhina (approved):
@pleek91 thank you for adding this!

PR #1522, opened 2022-02-12
Add diagram for scoped slots doc

NataliaTepluhina:
## Description of Problem

Currently, both slots and named slots have diagrams to illustrate the explanation. However, the scope slots chapter (which is the "hardest" to understand) lacks the visual explanation

## Proposed Solution

A diagram is added for scoped slots


skirtles-code:
I'm wondering about the small differences from the code that comes just before the picture. e.g. The use of `#default` rather than `v-slot` and the extra `<template>` tag. The text just before the example emphasises that the example is showing a single default slot with `v-slot`, but the code style in the picture is more akin to working with named scoped slots.

NataliaTepluhina:
@skirtles-code thank you for catching this! I've simplified a diagram to match the code, would you mind taking a second look? 🙏🏻

skirtles-code (approved):
Looks good!

PR #1528, opened 2022-02-13
Change 'e' to 'error'

MlinTec:
## Description of Problem
The 'error' is not undefined.

## Proposed Solution
Change the 'e' to 'error' or change the 'error' to 'e'.


skirtles-code (changes requested):
Looks good!

Would you be able to update the Playground links to match the code changes you've made?

Thanks.

skirtles-code:
Fixed in #1573.

PR #1536, opened 2022-02-13
fixed a typo

azizurrahmaan:
## Description of Problem
Found a typo 
https://github.com/vuejs/docs/blob/8dda0d8a04e8799821aa6d330a6b8199881b6abd/src/api/options-composition.md?plain=1#L89
## Proposed Solution
I think `is` should be replaced with `if`


skirtles-code (approved):
Thanks for the fix!

PR #1541, opened 2022-02-14
Add missing slash in closing header tag

tomek-ch:
There seems to be a "/" missing in the closing header tag in the code example, so I added one


skirtles-code (approved):
Thanks for the fix!

PR #1547, opened 2022-02-15
Update registration.md

wxsms:
## Description of Problem

`even` -> `evenly`

Just my understanding. if this feels not correct to you, feel free to close it. Thanks.

### DIFF
--- a/src/guide/components/registration.md
+++ b/src/guide/components/registration.md
@@ -49,7 +49,7 @@ Globally registered components can be used in the template of any component with
 <ComponentC/>
 ```
 
-This even applies to all subcomponents, meaning all three of these components will also be available _inside each other_.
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):

skirtles-code:
Thanks for the PR!

In this context, *even* is being used as an adverb. It doesn't have a specific meaning of its own in this usage, but it acts as a modifier, emphasising that the point being discussed is something that the reader might have not expected.

wxsms:
> Thanks for the PR!
> 
> In this context, _even_ is being used as an adverb. It doesn't have a specific meaning of its own in this usage, but it acts as a modifier, emphasising that the point being discussed is something that the reader might have not expected.

Sry, my bad. should I create another PR to revert this?

skirtles-code:
It's all good, Natalia's already reverted it. Thanks.

PR #1550, opened 2022-02-15
options-state.md typo

SquidwardHill:
## Description of Problem
Simple typo 
## Proposed Solution
explicit declared -> explicitly declared


skirtles-code (approved):
Thanks for the fix!

PR #1555, opened 2022-02-16
Fix unUnmounted to onUnmounted

BurkhalterY:
## Description of Problem
Error on exemple import (unUnmounted instead of onUnmounted)

## Proposed Solution
Replace unUnmounted by onUnmounted


skirtles-code (approved):
Thanks for the fix!

Issue #1548, opened 2022-02-15
Word breaks at inappropriate point in the Lifecycle Diagram


zetaraku:
The words **Mounted** and **Unmounted** in the [Lifecycle Diagram](https://github.com/vuejs/docs/blob/38d8292e74ea27d22dc58056a0a4cb171b664075/src/guide/essentials/images/lifecycle.png) are broken into two parts.

PR #1556, opened 2022-02-17
fix:  update lifecycle diagram (#1548)

aouos:
## Description of Problem
Update lifecycle diagram

## Proposed Solution
The words `Mounted` and `Unmounted` are broken into two parts.


bencodezen (approved):
Thanks for the PR!

bencodezen:
Note: This changes the typography a little bit, but fixing the orphan letters are a higher priority. We can re-standardize the typography later on.

zetaraku:
It seems that the picture is a manual replica of the diagram image.
As a link to the source Figma file is provided in the comment, should we just ask the author to edit it?

https://github.com/vuejs/docs/blob/38d8292e74ea27d22dc58056a0a4cb171b664075/src/guide/essentials/lifecycle.md?plain=1#L58-L66

PR #1557, opened 2022-02-17
chore(slots): fix typo

guillaumeduliscouet:
## Description of Problem

A confusion was made between header and footer


skirtles-code (approved):
Thanks for the fix!

PR #1560, opened 2022-02-17
add <label> tag for API Reference page

milos018:
## Description of Problem
The current API reference page has no <label> tag for the filter input, which is important for A11y issues.

## Proposed Solution
Inserted the <label> tag along with some CSS changes.
Wording "Filter" and "Enter keyword" may be changed for something better.

## Additional Information
My first PR (ever), please be gentle :)


skirtles-code (changes requested):
Thanks for the PR!

I think it might need a bit of tweaking for narrower viewport widths. There's a media query for that header, switching it from horizontal to vertical stacking when the viewport goes below 768px. I feel like the label and input should probably be left aligned and stacked vertically below that width.

What do you think?

milos018:
Thanks for the comment. I believe this is fixed now. Happy to make any other changes if needed.

src/api/ApiIndex.vue
```diff
@@ -204,6 +226,10 @@ h3 {
   .header {
     display: block;
   }
+
+  .api-filter {
+    justify-content: flex-start;
```
skirtles-code, line 231: I'm wondering whether we could replace the `justify-content: space-between` in the earlier rule with `flex-start`? From what I can tell it would look exactly the same, but without needing this extra tweak in the media query?

milos018: Yep, you are right. Will fix it. One more idea - is it OK to change the input type to "search" (instead of default "text) as we get the "x" (to cancel search) for free?

skirtles-code (approved):
Looks good!

Thanks.

PR #1561, opened 2022-02-18
fix the `v-for` attribute value

chenpeng991105:
## Description of Problem
the `v-for` attribute value is wrong at src/examples/src/cells/App/template.html
```html
<td v-for="c, j in cols">
```
## Proposed Solution
replaced with
```html
<td v-for="(c, j) in cols">
```


NataliaTepluhina (approved):
@chenpeng991105 good catch! Thank you for fixing it 👍🏻

Issue #1533, opened 2022-02-13
Spreadsheet in Guide/Extra Topics/Reactivity in Depth not readable in Dark Mode


c0lt:
I don't know if this issue is in the right place here or should go in the theme repo.
The Spreadsheet in the [Reactivity in Depth](https://vuejs.org/guide/extras/reactivity-in-depth.html#what-is-reactivity) page is not readable in Dark Mode. The contrast of the headers is too low and the content of a selected cell can't be seen at all:
![spreadsheet](https://user-images.githubusercontent.com/9382537/153761838-b5c6789c-cc3c-49e4-8d65-8099b1bcfb82.PNG)
![spreadsheet_cell](https://user-images.githubusercontent.com/9382537/153761840-c9f997c7-9321-4ff5-b91d-03343bf9f40a.PNG)

skirtles-code:
I believe this is fixed by #1564.

PR #1564, opened 2022-02-18
fix: SpreadSheet dark mode display (#1533) & add placeholder example (vuejs#1532)

aouos:
## Description of Problem
SpreadSheet dark mode text cannot be displayed

## Proposed Solution
Control the change of text and background color through CSS variable

## Additional Information
image

image


NataliaTepluhina (approved):
Thank you for the fix, @aouos!

PR #1567, opened 2022-02-21
Update reactivity-utilities.md

sombriks:
## Description of Problem

small clarification on toRef

## Proposed Solution

this small phrase


skirtles-code:
Thanks for the PR!

I'm afraid I don't understand this change. What do you mean by *'unless the source property is a component property'*?

skirtles-code:
Ah, OK, I understand what you meant.

I've tried to clarify the impact on mutating props in #1624.

PR #1571, opened 2022-02-21
Document v-bind.attr shorthand in API section.

LinusBorg:
## Description of Problem

We describe the vbind.prop shorthand, but not the v-bind.attr shorthand

## Proposed Solution

Add missing API documentation

### DIFF
--- a/src/api/built-in-directives.md
+++ b/src/api/built-in-directives.md
@@ -319,13 +319,16 @@ Dynamically bind one or more attributes, or a component prop to an expression.
   <svg><a :xlink:special="foo"></a></svg>
   ```
 
-  The `.prop` modifier also has a dedicated shorthand, `.`:
+  The `.prop` and `.attr` modifiers also have a dedicated shorthand, `.` and `^` respectively:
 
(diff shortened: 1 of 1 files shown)


src/api/built-in-directives.md
```diff
@@ -319,13 +319,16 @@ Dynamically bind one or more attributes, or a component prop to an expression.
...
 
   ```vue-html
   <div :someProperty.prop="someObject"></div>
-
   <!-- equivalent to -->
   <div .someProperty="someObject"></div>
+  
+  <div :someProperty.attr="someObject"></div>
+  <!-- equivalent to -->
+  <div ^someProperty="someObject"></div>
```
skirtles-code, line 331: I'm not exactly sure of the typical use case, but would it be more accurate for the example to use `someString` rather than `someObject` for the `.attr`/`^` example? I assume it would need to be a string to be set as an attribute?

```suggestion
  <div :someProperty.attr="someString"></div>
  <!-- equivalent to -->
  <div ^someProperty="someString"></div>
```

PR #1573, opened 2022-02-21
Watcher: Error catching fixed

m0shiurX:
## Description of Problem

There was a typo/mistake in the example try-catch block.
```
 try {
        .....
    } catch (e) {
      answer.value = 'Error! Could not reach the API. ' + error
    }
```

## Proposed Solution

It should either be `error` in the catch clause or `e` in the statement.
```
 try {
        .....
    } catch (error) {
      answer.value = 'Error! Could not reach the API. ' + error
    }
```


m0shiurX:
@skirtles-code Thanks for pointing it out, I have fixed the URL.

skirtles-code (approved):

PR #1574, opened 2022-02-22
Fix typo in tutorial step 2

NoelDeMartin:


skirtles-code:
Thanks for the PR!

I think that sentence could maybe be reworded to be a little easier to parse, but I believe `changed` is the correct word in the current sentence.

PR #1577, opened 2022-02-23
Change Vuex to Pinia and link it

vanillajonathan:
## Description of Problem
Vuex is the old store.

## Proposed Solution
Mention Pinia instead.

## Additional Information
Hyperlink it.


skirtles-code (approved):
Thanks for the improvement!

PR #1578, opened 2022-02-23
Update href of perfomance.mark

jairoblatt:
## Description of Problem
 href to `performance.mark` docs was empty
## Proposed Solution
  Updated to mozilla docs


skirtles-code (approved):
Thanks for the fix!

PR #1580, opened 2022-02-25
Update onBeforeUnmount type

florentinap:
## Description of Problem
There is a typo. The method name in Type section is onBeforeUnmounted (the correct name is onBeforeUnmount).

## Proposed Solution
Fix method name.


skirtles-code (approved):
Thanks for the fix!

PR #1583, opened 2022-02-27
docs: update computed.md code to follow given example

devzom:


skirtles-code:
Thanks for the PR!

I'm not really following what you're trying to do here. Could you explain the change in a bit more detail?

Variations on that template appear 3 times on that page, but they all seem to be consistent. They also seem to match their corresponding Playground examples.

skirtles-code:
I'm going to close this. If someone can provide further explanation about what this change is trying to achieve I'd be happy to reopen it.

PR #1591, opened 2022-03-02
Update watchers.md

pebutler3:
grammar fix

## Description of Problem
"This is useful when you need to conditional set up a watcher..."

## Proposed Solution
"This is useful when you need to conditionally set up a watcher"


skirtles-code (approved):
Thanks for the fix!

PR #1592, opened 2022-03-02
tutorial step 7: improve a11y, teach better practices

WickyNilliams:
## Description of Problem

this tutorial step teaches a bad practice, and is more complex than necessary as a result

## Proposed Solution

it's easier, and more accessible, to handle a single submit event on a form than it is to handle both click on button and enter keydown on input. so switch to that instead

## Additional Information

n/a


NataliaTepluhina (approved):
@WickyNilliams thank you, `submit` event is more semantic indeed

PR #1593, opened 2022-03-02
Update inconsistent naming in step-2 of the tutorial

jonksc:
## Description of Problem
The "state" object is inconsistent with the other parts of the tutorial, the object should be "counter" instead.
## Proposed Solution
Change "state" object to "counter".


NataliaTepluhina (approved):
@jonksc thank you for the fix!

PR #1594, opened 2022-03-02
changed variable name

modex98:
changed variable name, it sounds like my directory, while it's my directive actually


skirtles-code (approved):
Agreed. Non-standard abbreviations should be avoided.

PR #1595, opened 2022-03-02
fix example for consistency

Boldewyn:
The `<button>` element is already used in the example above. This commit updates the next example, where a `<div>` is used instead.


cesarkohl:
Agreed.

skirtles-code (approved):
Thanks for the fix!

Sorry for the delay getting this merged.

PR #1597, opened 2022-03-05
Improve clarity of a sentence in the "Lifecycle Hooks" section of the docs

jonksc:
## Description of Problem

> Do note this doesn't mean that the call must be placed lexically inside `setup()` or `<script setup>` - `onMounted()` can be called in an external function as long as the call stack is synchronous and originates from within `setup()`.

The dash in this sentence makes it look like it's `<script setup>` - `onMounted()`, when in fact they should be separate.
## Proposed Solution
To improve clarity and flow of the sentence, I suggest using a fullstop instead.

### DIFF
--- a/src/guide/essentials/lifecycle.md
+++ b/src/guide/essentials/lifecycle.md
@@ -51,7 +51,7 @@ setTimeout(() => {
 }, 100)
 ```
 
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Thanks!

PR #1598, opened 2022-03-05
Fix a grammatical error in the docs

jonksc:
## Description of Problem
There's a minor grammatical error in a sentence.
## Proposed Solution
Add the word "at".


skirtles-code (approved):
Thanks for the PR!

I don't think the original is wrong, but I do prefer your wording.

PR #1600, opened 2022-03-06
Update input bindings in Temperature converter 7GUIs example

anubra266:
## Description of Problem
In the [Temperature Converter (7GUIs) example](https://vuejs.org/examples/#temperature-converter), the inputs still use the `@change` directive, so it doesn't work.

## Proposed Solution
I updated them to use the `@input` directive.

## Additional Information
I just learnt vue between yesterday and today, I was practicing the examples when I encountered this, so I appreciate any feedback.

### DIFF
--- a/src/examples/src/temperature-converter/App/template.html
+++ b/src/examples/src/temperature-converter/App/template.html
@@ -1,2 +1,2 @@
-<input type="number" :value="c" @change="setC"> Celsius =
-<input type="number" :value="f" @change="setF"> Fahrenheit
+<input type="number" :value="c" @input="setC"> Celsius =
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

I saw the conversation about this on Vue Land. Based on that I believe this was simply a misunderstanding and no change is actually required.

ElMassimo:
@skirtles-code What was the discussion in Vue Land?

By using `change`, typing in the input does not reflect the value, making the example less interactive (and would look broken to most users).

For reference, the [Svelte demo](https://svelte.dev/examples/7guis-temperature) uses `input` instead of `change`.

PR #1602, opened 2022-03-07
menghapus text

smknj-muhammadsofyan:


skirtles-code:
I don't understand why this text has been removed.

PR #1603, opened 2022-03-07
fix: update typo in `web-components.md`

tomaszwylezek:
## Description of Problem
There was a typo, it should be `detail` instead of `details` 
## Proposed Solution
Adjusted it
## Additional Information
https://github.com/vuejs/core/blob/5898629d723e82b68e9b17b91bf8b1a8390a3912/packages/runtime-dom/src/apiCustomElement.ts#L341-L347

### DIFF
--- a/src/guide/extras/web-components.md
+++ b/src/guide/extras/web-components.md
@@ -151,7 +151,7 @@ document.body.appendChild(
 
 #### Events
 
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Thanks for the fix!

PR #1604, opened 2022-03-07
Remove redundancy in reactivity-fundamentals.md

sharonytlau:
## Description of Problem

> It also means that when we assign or destructure a reactive object's property into local variables, or when we pass that property into a function, or destructure properties from a reactive object, we will lose the reactivity connection:

In the above sentence, "or destructure properties from a reactive object"  is redundant and may cause confusion.

## Proposed Solution

Remove the redundancy:

> It also means that when we assign or destructure a reactive object's property into local variables, or when we pass that property into a function, we will lose the reactivity connection:

It will make the explanation better match the code example.

### DIFF
--- a/src/guide/essentials/reactivity-fundamentals.md
+++ b/src/guide/essentials/reactivity-fundamentals.md
@@ -352,7 +352,7 @@ The `reactive()` API has two limitations:
    state = reactive({ count: 1 })
    ```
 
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Agreed, this does appear to be repeating itself. Thanks for the fix!

PR #1605, opened 2022-03-07
, expected

rizimore:
## Description of Problem
, expected in example code

## Proposed Solution
simply, I put the comma, so if someone copy the example they don't face error


skirtles-code (approved):
Thanks for the fix!

PR #1606, opened 2022-03-07
Table is not correctly formatted

rizimore:
## Description of Problem
The table is not correctly formated according to MD guidelines

## Proposed Solution
I simply fixed it according to MD guideliens

## Additional Information
https://youtrack.jetbrains.com/issue/IDEA-283693


skirtles-code:
@rizimore Could you provide more information about which MD guidelines you mean? The link you provided seems to suggest that the current formatting is correct and that there's a bug with IntelliJ IDEA.

rizimore:
You can refer to this: https://www.jetbrains.com/help/phpstorm/markdown-table-doesn-t-have-side-borders.html

rizimore:
Its MD rule and Jetbrain code inspect as well

rizimore:
> You can refer to this: https://www.jetbrains.com/help/phpstorm/markdown-table-doesn-t-have-side-borders.html

This format comes under this (not exactly this)

skirtles-code:
I'm afraid I don't see anything there to suggest that there's a problem with the way this table is formatted.

The docs are using VitePress, which uses [markdown-it](https://github.com/markdown-it/markdown-it) for markdown parsing. That uses the same table rules as [GFM](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables). Those latter docs show the same formatting as we have on this table. This is also the style used by Prettier. JetBrains seems to be the odd-one-out in insisting that the pipes have to go all the way to the edge.

PR #1608, opened 2022-03-09

WangNingning1994:
In the DOM Update Timing section's example, it seems we should use state.count++(which has been shown above) instead of count.value++(which comes later).


skirtles-code (approved):
Thanks for the fix!

I've a feeling this part was originally further down the page, after the bits about `ref`. In that context, `count.value` would have made sense. In its present location it makes much more sense to use `state.count`.

PR #1615, opened 2022-03-12
fix: missing comma at example src/guide/essentials/component-basics.md

RioChndr:
## Description of Problem
While copying example code at section component basic, I found missing comma at example `src/guide/essentials/component-basics.md`.

## Proposed Solution
Add an comma 😁


skirtles-code (approved):
Thanks for the fix!

PR #1616, opened 2022-03-12
update Ref type import

jeff-fe:
In TypeScript 4.5+, This practice appears  error message(`TS1444: 'Ref' is a type and must be imported using a type-only import when 'preserveValueImports' and 'isolatedModules' are both enabled.`). We should change Type import mode.

image

Reference linking: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta

### DIFF
--- a/src/guide/typescript/composition-api.md
+++ b/src/guide/typescript/composition-api.md
@@ -162,7 +162,7 @@ year.value = '2020'
 Sometimes we may need to specify complex types for a ref's inner value. We can do that by using the `Ref` type:
 
 ```ts
-import { ref, Ref } from 'vue'
+import { ref, type Ref } from 'vue'
 
 const year: Ref<string | number> = ref('2020')
 


dammy001 (commented):
LGTM @skirtles-code

skirtles-code:
Thanks for the PR. Sorry for the delay. I don't use TS myself, so I didn't really feel qualified to review it, and when I reached out for help it turned out there were various other complications.

The syntax currently used in the docs is valid, but it isn't compatible with the default settings used by `npm init vue`, giving the warning mentioned above. Whether Vite needs to use those settings is up for discussion, but as it doesn't seem likely it'll change in the near future we do need to address this in the docs.

There are two different ways this could be addressed using the `type` modifier. The one shown in this PR would be my preferred option, but it doesn't currently work with the SFC Playground.

To ensure compatibility with both Vite and the Playground, I've opened PR #1656 to apply the alternative fix.

I've updated some of the other examples in that PR too, as there are a handful of other cases where types are being imported.

jeff-fe:
> Thanks for the PR. Sorry for the delay. I don't use TS myself, so I didn't really feel qualified to review it, and when I reached out for help it turned out there were various other complications.
> 
> The syntax currently used in the docs is valid, but it isn't compatible with the default settings used by `npm init vue`, giving the warning mentioned above. Whether Vite needs to use those settings is up for discussion, but as it doesn't seem likely it'll change in the near future we do need to address this in the docs.
> 
> There are two different ways this could be addressed using the `type` modifier. The one shown in this PR would be my preferred option, but it doesn't currently work with the SFC Playground.
> 
> To ensure compatibility with both Vite and the Playground, I've opened PR #1656 to apply the alternative fix.
> 
> I've updated some of the other examples in that PR too, as there are a handful of other cases where types are being imported.

Thank you very much

PR #1617, opened 2022-03-13
fix: correct Markdown syntax and broken links

luvejo:
Fixed this formatting error on [SFC Syntax Specification](https://vuejs.org/api/sfc-spec.html#pre-processors) page:

Screenshot from 2022-03-12 20-22-37


skirtles-code (approved):
Thanks for the fix!

PR #1624, opened 2022-03-17
Update API entry for toRef to mention mutating props

skirtles-code:
This is my attempt at addressing the concern raised in #1567.

The API entry for `toRef` mentions using it to create refs from props. This change attempts to clarify that the usual restriction on not mutating a prop still applies, using `toRef` is not a fix for that.

I've also made a few other tweaks to this entry:

* It said `string` when it meant `number`.
* An example wasn't importing `toRef`. Usually I wouldn't show the imports, but given the `<script setup>` tag was shown, I felt the import should be shown too.
* The final note about using `toRef` with optional properties was added back when props did not include omitted values. That has since been changed, so props now includes all props, even if they aren't passed. I've reworded that note to avoid mentioning props.

### DIFF
--- a/src/api/reactivity-utilities.md
+++ b/src/api/reactivity-utilities.md
@@ -80,12 +80,14 @@ Can be used to create a ref for a property on a source reactive object. The crea
   const fooRef = ref(state.foo)
   ```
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
Thank you @skirtles-code! Looks good to me 👍🏻

Issue #1625, opened 2022-03-18
Fix broken links on Reactivity Fundamentals page


ivangermanov:
1. Navigate to the https://vuejs.org/guide/essentials/reactivity-fundamentals.html page.
2. Keep your eyes on the URL and start scrolling down.
3. You will first see `https://vuejs.org/guide/essentials/reactivity-fundamentals.html#script-setup` (this shouldn't be the case)
4. Then, you will see `https://vuejs.org/guide/essentials/reactivity-fundamentals.html#declaring-reactive-state` (this **should** be the case, since it matches the header name).
5. Continue scrolling down where the header is _Reactive Proxy vs. Original_. The URL there is again `https://vuejs.org/guide/essentials/reactivity-fundamentals.html#script-setup`, but it should be `https://vuejs.org/guide/essentials/reactivity-fundamentals.html#reactive-proxy-vs-original`.
6. The same problem with incorrect links appears for the _Deep Reactivity_ (which gets the https://vuejs.org/guide/essentials/reactivity-fundamentals.html#ref-unwrapping-in-reactive-objects) link and _Stateful Methods_  (which gets the https://vuejs.org/guide/essentials/reactivity-fundamentals.html#reactivity-transform link) headers.

This also causes the Algolia search component to not correctly link to the subsections:
![image](https://user-images.githubusercontent.com/42522864/158982453-0d782a93-9234-4168-9356-e8ea307df201.png)

I could not find the cause why these links (such as https://vuejs.org/guide/essentials/reactivity-fundamentals.html#ref-unwrapping-in-reactive-objects) exist. I performed a solution-wide search for them, and some of them don't even exist. Others do exist, but have nothing to do with the _Reactivity Fundamentals_ page.

I first discovered this when reading https://vuejs.org/api/composition-api-setup.html#basic-usage and clicking on the _automatically shallow unwrapped_ link in the same section. This links to https://vuejs.org/guide/essentials/reactivity-fundamentals.html#ref-unwrapping-in-templates, but it should probably link to https://vuejs.org/guide/essentials/reactivity-fundamentals.html#deep-reactivity. I'll link a PR to fix that link.

skirtles-code:
It seems to be a consequence of the API Preference switch. The hidden sections from the other API are being matched.

NataliaTepluhina:
@ivangermanov I cannot reproduce the described bug anymore 🤔 Please let me know if it's still reproducible for you (if not, let's close the issue)

PR #1626, opened 2022-03-18
Fix broken 'ref-unwrapping-in-templates' link to

ivangermanov:
## Description of Problem
Problem described in https://github.com/vuejs/docs/issues/1625

## Proposed Solution
The solution is partial—fix the link. The other problems from the issue will still remain.


NataliaTepluhina (approved):
@ivangermanov thank you for the fix!

PR #1627, opened 2022-03-21
fix: add h3 in src/guide/typescript/composition-api.md

jCodeLife:
Add 'Using <script setup>' as H3，Corresponding to the h3 below --Without <script setup>

### DIFF
--- a/src/guide/typescript/composition-api.md
+++ b/src/guide/typescript/composition-api.md
@@ -4,6 +4,8 @@
 
 ## Typing Component Props
 
+### using `<script setup>`
+
 When using `<script setup>`, the `defineProps()` macro supports inferring the props types based on its argument:
 
 ```vue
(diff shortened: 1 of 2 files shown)


src/guide/typescript/composition-api.md
```diff
@@ -4,6 +4,8 @@
 
 ## Typing Component Props
 
+### using `<script setup>`
```
NataliaTepluhina, line 7: @jCodeLife that's a nice addition! Could we please capitalize the first letter to be consistent with other headings? 🙏🏻 
```suggestion
### Using `<script setup>`
```

jCodeLife: Yes,you are right

jCodeLife (commented):
It is no problem

skirtles-code (approved):

PR #1631, opened 2022-03-22
update example code to include return

azyth:
## Description of Problem
missing return statement in code example


skirtles-code:
Thanks for the PR, but the code is using `<script setup>`, so a `return` would not be valid here.

PR #1637, opened 2022-03-24
fix typo

demahom18:
## Description of Problem
A little typo error
## Proposed Solution
Fixed it


skirtles-code (approved):
Thanks for the fix!

PR #1640, opened 2022-03-25
fix: toRef is used without import

mammad2c:
## Description of Problem
we used `toRef` in Accessing Props section, but we didn't import it.

## Proposed Solution
add `toRef` to list of imported properties from `vue` package.


skirtles-code (approved):
Thanks for the fix!

PR #1644, opened 2022-03-26
this's value in setup is undefined

Huauauaa:
## Description of Problem
this's value in setup is undefined


NataliaTepluhina (approved):
@Huauauaa thank you for the contribution!

PR #1645, opened 2022-03-26
Add a section on README about the recommended Node.js version.

StephenChips:


StephenChips:
At the 12th step of the tutorial, the example code snippet has a typo. If user use vue in html, they should import JavaScript file instead of Vue file.

### Updated Webpage
cap

### Old Webpage
cap2

README.md
```diff
@@ -20,3 +20,7 @@ pnpm run dev
 ## Working on the theme
 
 If changes need to made for the theme, check out the [instructions for developing the theme alongside the docs](https://github.com/vuejs/vue-theme#developing-with-real-content).
+
+## Node.js Version Requirement
+
+This project uses JavaScript's new features like optional chaining, so it is recommended to use `v14.0.0` or higher.
```
skirtles-code, line 26: I'm not sure we need a separate section for this. Personally I think it would be sufficient just to mention the required version at the end of the `Contributing` section, just after the `pnpm` commands.

I also wonder whether we should include an `engines` section in `package.json`, to enforce the version explicitly rather than relying on documenting it?

skirtles-code (changes requested):
Thanks for the PR! I've made some suggestions, let me know what you think.

README.md
StephenChips: I generally approve your suggestion, It is briefer that just appending the the required version at the end of the `Contributing` section. However, since I never used the `engines` field before, I do not know whether it will print out warning or error messages or not, when an user hits `pnpm i`. According to the *npm* document:

> Unless the user has set the engine-strict config flag, this field is advisory only and will **only produce warnings when your package is installed as a dependency**.

[https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines)

Since this project is a standalone website instead of a third-party library, it will never be a dependency of other projects, so I am afraid that warnning will never be generated.

If the `engines` field is just a statement that will have any effects, I think it would be better to write down the requirement on the document. IMO, people like me is such lazy that probably won't take a look at the `packages.json` file after cloning the repository, and maybe have never thought that the the required version has been written in the `engines` field somewhere in the `packages.json` file. What I would do is to follow the installation guide and hope I can kick off the project ASAP, which will end up with a runtime error. If the required version have been written in the document, I will be informed and will get my Node.js updated first, since I have to visit the repository's home page before cloning.

src/tutorial/src/step-11/description.md
```diff
@@ -45,7 +45,7 @@ Then, we can use the component in the template as:
 <div class="html">
 
 ```js
-import ChildComp from './ChildComp.vue'
```
bencodezen, line 48: Not sure why you're suggesting `.js` instead. The tutorial includes a file for `.vue` file.

StephenChips: Because there is not a SFC transformer if you use Vue.js in a `.html` file directly. Any `.vue` file must be processed by the building system (vite, webpack etc.), then will be transformed into component's JavaScript and CSS file. If you use Vue.js in HTML, without a building system, you would rather import the JavaScript file that define the *Vue* object directly. If you click the `reset` button, you may see that the "CORRECT" code imports a `.js` file instead of a `.vue` file.

skirtles-code: I believe this change is correct.

From a process perspective, generally it's preferred to put unrelated changes in separate PRs so they can be reviewed and merged independently. I'd be happy to merge this extension change, but I can't because it's in the same PR as another change that needs further tweaking.

StephenChips: Thannks for your advice. I'll pay attention next time.

PR #1646, opened 2022-03-28
fix: remove obtrusive punctuation marks

jCodeLife:
## Description of Problem
obtrusive punctuation marks

Because there's nothing in front or back
## Proposed Solution
remove obtrusive punctuation marks


skirtles-code (approved):
Thanks for the fix!

PR #1651, opened 2022-03-31
Update component-basics.md

BrManga:
Changed the code example to fit with the render example. It was 'Error' in code but rendered as 'This is an Error for Demo Purposes'.

## Description of Problem
Code example was not fitting with rendered result. 
## Proposed Solution
Changing 'Error!' text to  'This is an Error for Demo Purposes' to avoid confusion.
 ## Additional Information


skirtles-code (approved):
Thanks for the fix!

PR #1655, opened 2022-04-10
docs: src/api/render-function.md

changshou83:
## Description of Problem
missing words on line 96


skirtles-code (approved):
Thanks for the PR!

Agreed, that link does seem to be inconsistent with the other links on that same page.

PR #1656, opened 2022-04-10
Use 'import type' in TypeScript examples

skirtles-code:
This PR attempts to address the problem raised in #1616.

The TS settings currently used by `npm init vue` require the use of the `type` modifier when importing types. So the following won't work:

```ts
import { ref, Ref } from 'vue'
```

This could be fixed using:

```ts
import { ref, type Ref } from 'vue'
```

However, this won't currently work in the SFC Playground. Details at https://github.com/vuejs/repl/pull/28.

The only syntax that seems to work in both cases is:

```ts
import { ref } from 'vue'
import type { Ref } from 'vue'
```

I'm not convinced this is ideal either. While it does work in all cases, I feel this is probably the least likely of the 3 options to be used in a real project.

In this PR I've switched the examples to use that last syntax. There aren't many examples that are affected. Once the Playground supports the second syntax I'd be inclined to switch to that instead.

### DIFF
--- a/src/guide/typescript/composition-api.md
+++ b/src/guide/typescript/composition-api.md
@@ -164,7 +164,8 @@ year.value = '2020'
 Sometimes we may need to specify complex types for a ref's inner value. We can do that by using the `Ref` type:
 
 ```ts
-import { ref, Ref } from 'vue'
+import { ref } from 'vue'
+import type { Ref } from 'vue'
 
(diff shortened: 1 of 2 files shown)


dammy001 (approved):
LGTM 🤌🏾

NataliaTepluhina (approved):

PR #1659, opened 2022-04-12
Update introduction.md

amwebdev86:
word was missing and it read " the things we build on the web may vary drastically in form and scale. " added 'grow'
 change: " the things we build on the web may grow vary drastically in form and scale.


skirtles-code:
Thanks for the PR, but I believe the original sentence is correct. I think you may have misread 'vary' as 'very'.

amwebdev86:
My mistake 😅

PR #1660, opened 2022-04-12
Update state-management.md

enoughtal:
in line 200, is there a typo?


enoughtal:
> in line 200, is there a typo?
> 
> ## Description of Problem
> ## Proposed Solution
> ## Additional Information

PR #1676, opened 2022-04-20
Fixed typo (plural s missing)

GingaaBread:
When you mutate reactive state, it [...] should be either When you mutate a reactive state, it [...] or When you mutate reactive states, it [...]


skirtles-code:
Thanks for the PR, but I believe the current wording is correct. If it said 'reactive properties' then I would agree that it needs to be plural, but I don't think 'states' works here.

PR #1677, opened 2022-04-20
Update quick-start without building

brignolij:
## Description of Problem
I spend a lot of time making vue 3 quick start without build tool working. The second example is not working on Firefox.
I also spend time searching how to import and use _ref_.

## Proposed Solution
I propose a new example in the documentation showing how to import ref as an example of how to import Vue stuff.

### DIFF
--- a/src/guide/quick-start.md
+++ b/src/guide/quick-start.md
@@ -83,8 +83,28 @@ To get started with Vue without a build step, simply copy the following code int
   }).mount('#app')
 </script>
 ```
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR, I think a full code block would be too repetitive, so I adjusted it to be more minimal in https://github.com/vuejs/docs/commit/4560bc1e3684152f4703ba01b7e3a1045970ebd4

PR #1679, opened 2022-04-22
docs: fix typo

Yorksh1re:

### DIFF
--- a/src/api/built-in-components.md
+++ b/src/api/built-in-components.md
@@ -29,7 +29,7 @@ Provides animated transition effects to a **single** element or component.
   interface TransitionProps {
     /**
      * Used to automatically generate transition CSS class names.
-     * e.g. `name: fade'` will auto expand to `.fade-enter`,
+     * e.g. `name: fade` will auto expand to `.fade-enter`,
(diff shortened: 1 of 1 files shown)


src/api/built-in-components.md
```diff
@@ -29,7 +29,7 @@ Provides animated transition effects to a **single** element or component.
   interface TransitionProps {
     /**
      * Used to automatically generate transition CSS class names.
-     * e.g. `name: fade'` will auto expand to `.fade-enter`,
+     * e.g. `name: fade` will auto expand to `.fade-enter`,
```
skirtles-code, line 32:
```suggestion
     * e.g. `name: 'fade'` will auto expand to `.fade-enter`,
```

skirtles-code (changes requested):
Thanks for the PR!

I think the problem here is a missing quote, rather than a superfluous quote. Please could you review my proposed change below and merge it into your branch if you're happy with it?

Yorksh1re:
I think you're right, I'm translating the Chinese version of the new vue3 documentation, and I'll sync corrections there.

PR #1682, opened 2022-04-23
Fix wording about component use with <script setup>

### LINKED ISSUE #1638: docs wrongly say that in script setup components are automatically registered locally when they are not
## Docs

https://vuejs.org/guide/components/registration.html#local-registration

> When using SFC with `<script setup>`, imported components are automatically registered locally:

If they are registered locally it should be possible to do `<component is="comp" />`

## Plaground

https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cD5cbiAgaW1wb3J0IENvbXAgZnJvbSAnLi9Db21wLnZ1ZSc7XG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8Y29tcCAvPlxuPC90ZW1wbGF0ZT4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL3NmYy52dWVqcy5vcmcvdnVlLnJ1bnRpbWUuZXNtLWJyb3dzZXIuanNcIlxuICB9XG59IiwiQ29tcC52dWUiOiI8dGVtcGxhdGU+Zm9vPC90ZW1wbGF0ZT4ifQ==

## Compiled JS

```js
/* Analyzed bindings: {
  "Comp": "setup-const"
} */
import { openBlock as _openBlock, createBlock as _createBlock } from "vue"

import Comp from './Comp.vue';

const __sfc__ = {
  setup(__props) {

  
return (_ctx, _cache) => {
  return (_openBlock(), _createBlock(Comp))
}
}

}
__sfc__.__file = "App.vue"
export default __sfc__
```

leopiccionia:
Fix #1638.

The first paragraph mentions that:

> A Vue component needs to be "registered" so that Vue knows where to locate its implementation when it is encountered in a template. There are two ways to register components: global and local.

Now that the content contradicts it, I wonder if we need to reword it, too.

### DIFF
--- a/src/guide/components/registration.md
+++ b/src/guide/components/registration.md
@@ -65,7 +65,7 @@ Local registration scopes the availability of the registered components to the c
 
 <div class="composition-api">
 
-When using SFC with `<script setup>`, imported components are automatically registered locally:
+When using SFC with `<script setup>`, imported components can be locally used without registration:
 
 ```vue
 <script setup>


dammy001 (approved):
@skirtles-code

NataliaTepluhina (approved):

PR #1685, opened 2022-04-24
Update template-syntax.md

Sepi1211:


skirtles-code:
Thanks for the PR, but I believe 'to' is correct in this context.

PR #1686, opened 2022-04-25
fix: banner height error

Jiangmenghao:
## Description of Problem
Wrong banner height(Vue School Banner) leads to wrong nav-bar position

## Proposed Solution
VueSchoolBanner.vue:
```
.has-top-banner {
  --vt-banner-height: 80px;
}
```


skirtles-code (approved):
Thanks!

PR #1692, opened 2022-04-30
Added falsy values mdn docs link

imadatyatalah:
## Description of Problem

falsy values docs link is missed.

## Proposed Solution

added falsy value mdn docs link to `falsy values` text.


NataliaTepluhina (approved):
@imadatyatalah thank you for the addition!

PR #1698, opened 2022-05-04
Update component-instance.md

kenneth-saey:
Fixed a typo in `$nextTick()` details.

## Description of Problem
The details of the `$nextTick()`-section use a wrong word. It starts with "The only **different** from the global version...". This should be "The only **difference** from the global version...".

## Proposed Solution
Change "different" into "difference".


skirtles-code (approved):
Thanks for the fix!

Issue #1691, opened 2022-04-28
Undefined path in the last tutorial page (#step-15)


chrisnaadhi:
The last tutorial page ([#step-15](https://vuejs.org/tutorial/#step-15)) has some undefined link path in their anchor tag :
- Quick Start link
- Main Guide link
- Examples link

Expecting each of the link go to their path in docs, but got 404 (i know i can access it via Docs dropdown for those links, but still expect some quick link in the tutorial😄)

Screenshot :
![image](https://user-images.githubusercontent.com/29543851/165768308-353c041d-1d04-4bc4-86a9-e0708ca2dbca.png)

PR #1699, opened 2022-05-04
Fix for #1691

BilleLind:
## Description of Problem

404 errors on the links in /tutorials/#step-15 
- Quick Start
- Main Guide
- Examples

## Proposed Solution

Changing the links from markdown based links `[]() to <a href="">` like in the previous steps

## Additional Information
Tough unrelated, it is my first time committing to an open source.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: 5.3
+lockfileVersion: 5.4
 
 specifiers:
   '@types/markdown-it': ^12.2.3
--- a/src/tutorial/src/step-15/description.md
+++ b/src/tutorial/src/step-15/description.md
@@ -4,10 +4,10 @@ You have finished the tutorial!
 
 At this point, you should have a good idea of what it's like to work with Vue. However, we covered a lot of things really fast and glossed over the details, so definitely keep learning! As a next step, you can:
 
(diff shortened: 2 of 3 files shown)


BilleLind:
I would wager it is due to createMarkdownRenderer from  vitepress, since i dont understand most of the code, i think it is because of a missing base string =>

in https://github.com/vuejs/docs/blob/main/src/tutorial/tutorial.data.ts
it utilizes "createMarkdownRenderer" from vitepress,

which accepts 3 params: srcDir, options and **base**
https://github.com/vuejs/vitepress/blob/main/src/node/markdown/markdown.ts

BilleLind:
By changing the const md in /tutorial/tutorial.data.ts (added , '/' )

`const md = createMarkdownRenderer(process.cwd(), {
      // @ts-ignore
      highlight: await createHighlighter()
    }, '/')
`
it makes it possible to utilize [example](/examples/), though with a flash of 404 before going to the correct page, which I am uncertain if it is caused by it being run locally. 

Should I commit it along with the current pull request?

yyx990803:
I think simply adding the missing base to `createMarkdownRenderer` is good enough. I'm not seeing flash of 404 when testing locally.

BilleLind:
Deployed the fix for createMarkdownRenderer and the links are working, also reverted the prior changes made in step 15

PR #1706, opened 2022-05-09
Missing closing parentheses in `computed` function.

r2abreu:
## Description of Problem

Missing closing parentheses.

## Proposed Solution

Adding parentheses.

## Additional Information

Nothing else to comment.

### DIFF
--- a/src/guide/extras/reactivity-in-depth.md
+++ b/src/guide/extras/reactivity-in-depth.md
@@ -176,7 +176,7 @@ import { ref, computed } from 'vue'
 
 const A0 = ref(0)
 const A1 = ref(1)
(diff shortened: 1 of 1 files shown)


dammy001 (approved):

skirtles-code:
Thanks for the PR, but I believe the existing code is correct. There are already two closing `)` on those lines, the first one comes immediately after the two opening `(`.

r2abreu:
> Thanks for the PR, but I believe the existing code is correct. There are already two closing `)` on those lines, the first one comes immediately after the two opening `(`.

Yup you're right, I should wear glasses. 

Thanks.

PR #1713, opened 2022-05-13
Add a v-bind example to match what the guide states

JamShady:
## Description of Problem
The current documentation refers to a non-existent v-bind example. This just adds the example in (and slightly adjusts the text to be more contextually accurate)

## Proposed Solution
Added a code block example using v-bind

### DIFF
--- a/src/guide/essentials/component-basics.md
+++ b/src/guide/essentials/component-basics.md
@@ -297,7 +297,14 @@ Then want to render a component for each one, using `v-for`:
 
 </div>
 
-Notice how we can use `v-bind` to pass dynamic props. This is especially useful when you don't know the exact content you're going to render ahead of time.
+We can also use `v-bind` to pass props dynamically. This is especially useful when you don't know the exact content you're going to render ahead of time.
+
+```vue-html
+<BlogPost
+  v-for="post in posts"
+  v-bind="post"
+ />
+```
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

I believe that the existing wording is referring to the preceding example. `v-bind` is being used in that example via the shorthand `:`. I don't think that sentence was intended to refer to the ability to use `v-bind="obj"`.

JamShady:
@skirtles-code OK, I see where you're getting at, but the wording is ambiguous and unclear.

The current wording says "Notice how we can use `v-bind` to pass dynamic props. This is especially useful when you don't know the exact content you're going to render ahead of time"

The example not only does not use `v-bind` (as you say, it uses a short-hand form), but it also explicitly defines the properties to bind, i.e. 'key/id', and 'title'.... which is not 'dynamic' (if you understand the dynamic part to be the variable props, as opposed to props that are variables)

If it is supposed to mean, as you say, you can use a shorthand form to pass in dynamic properties (i.e. variables instead of hard-coded strings), then it should state something like "Notice how we can use `v-bind` to pass variables as dynamic props. This is especially useful when you don't know what the content values will be ahead of time"

skirtles-code:
Reopening so we don't lose track of this.

I agree that the current wording 'dynamic props' is ambiguous and should probably be tweaked, maybe to something like 'dynamic prop values'.

PR #1718, opened 2022-05-15
use proper punctuation in text formatting

hossein13m:

### DIFF
--- a/src/guide/essentials/class-and-style.md
+++ b/src/guide/essentials/class-and-style.md
@@ -1,6 +1,6 @@
 # Class and Style Bindings
 
-A common need for data binding is manipulating an element's class list and its inline styles. Since they are both attributes, we can use `v-bind` to handle them. we only need to calculate a final string with our expressions. However, meddling with string concatenation is annoying and error-prone. For this reason, Vue provides special enhancements when `v-bind` is used with `class` and `style`. In addition to strings, the expressions can also evaluate to objects or arrays.
+A common need for data binding is manipulating an element's class list and its inline styles. Since they are both attributes, we can use `v-bind` to handle them since we only need to calculate a final string with our expressions. However, meddling with string concatenation is annoying and error-prone. For this reason, Vue provides special enhancements when `v-bind` is used with `class` and `style`. In addition to strings, the expressions can also evaluate to objects or arrays.
 
 ## Binding HTML Classes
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

Could you explain why you think the current punctuation is wrong? A colon seems preferable to me. I'm not sure the part after the punctuation mark really makes sense as a separate sentence.

hossein13m:
@skirtles-code Thanks for your time.

After the sentence, if we provide some list, it would be better to use `:`, but since the sentence is over, we can use a period and move on to the following sentence(s), which we can explain more.

e.g., there are two types of something: 1. type one, 2. type 2.
e.g., we can choose between multiple approaches. We can use approach one, which has the description or prefer the next approach with the following description.

src/guide/essentials/class-and-style.md
```diff
@@ -1,6 +1,6 @@
 # Class and Style Bindings
 
-A common need for data binding is manipulating an element's class list and its inline styles. Since they are both attributes, we can use `v-bind` to handle them: we only need to calculate a final string with our expressions. However, meddling with string concatenation is annoying and error-prone. For this reason, Vue provides special enhancements when `v-bind` is used with `class` and `style`. In addition to strings, the expressions can also evaluate to objects or arrays.
+A common need for data binding is manipulating an element's class list and its inline styles. Since they are both attributes, we can use `v-bind` to handle them. we only need to calculate a final string with our expressions. However, meddling with string concatenation is annoying and error-prone. For this reason, Vue provides special enhancements when `v-bind` is used with `class` and `style`. In addition to strings, the expressions can also evaluate to objects or arrays.
```
bencodezen, line 3:
```suggestion
A common need for data binding is manipulating an element's class list and its inline styles. Since they are both attributes, we can use `v-bind` to handle them since we only need to calculate a final string with our expressions. However, meddling with string concatenation is annoying and error-prone. For this reason, Vue provides special enhancements when `v-bind` is used with `class` and `style`. In addition to strings, the expressions can also evaluate to objects or arrays.
```

bencodezen: This might resolve the punctuation issue and resolve the odd transition between the two parts. What do people think?

skirtles-code, line 3: I'm wondering about a slightly bigger rewording. I'm not overly keen on the existing use of the word 'only' in that sentence. I feel like that usage contradicts the docs guidelines about not invalidating the struggle.

```suggestion
A common need for data binding is manipulating an element's class list and inline styles. Since `class` and `style` are both attributes, we can use `v-bind` to assign them a string value dynamically, much like with other attributes. However, trying to generate those values using string manipulation can be annoying and error-prone. For this reason, Vue provides special enhancements when `v-bind` is used with `class` and `style`. In addition to strings, the expressions can also evaluate to objects or arrays.
```

bencodezen: Sorry for missing this @skirtles-code. I like your suggestion here better. Please go ahead and override my mistaken commit!

skirtles-code:
This should be addressed by #1771.

PR #1720, opened 2022-05-15
Add more specific example code to Type Augmentation docs.

connorshea:
## Description of Problem

I found the prose-only description a bit confusing and got stuck for a solid ten minutes trying to figure it out.

## Proposed Solution

I've added example code of what does and doesn't work, since I find that more immediately understandable.


NataliaTepluhina (approved):
@connorshea thank you for adding the example!

PR #1722, opened 2022-05-16
Be explicit about Composition API style

letroot:
## Description of Problem
The Composition API is fairly new and folks who are more familiar with the Options API may feel a bit lost when `npm unit vue/@latest` generates SFCs (HelloWorld.vue, etc.) written in Composition style.

## Proposed Solution
I think it should be explicitly made clear that the init Vue scaffolding Single File Components are in Composition API style. I think explicit is better than implicit. Helps beginners feel more safe.

### DIFF
--- a/src/guide/quick-start.md
+++ b/src/guide/quick-start.md
@@ -48,7 +48,7 @@ If you are unsure about an option, simply choose `No` by hitting enter for now. 
 <span class="line"><span style="color:var(--vt-c-green);">&gt; </span><span style="color:#A6ACCD;">npm run dev</span></span>
 <span class="line"></span></code></pre></div>
 
-You should now have your first Vue project running! Here are some additional tips:
+You should now have your first Vue project running! You should note here that the components in the generated app are written in [Composition API](guide/extras/composition-api-faq.html) style. Here are some additional tips:
 
 - The recommended IDE setup is [Visual Studio Code](https://code.visualstudio.com/) + [Volar extension](https://marketplace.visualstudio.com/items?itemName=Vue.volar). [WebStorm](https://www.jetbrains.com/webstorm/) is also viable.
 - More tooling details, including integration with backend frameworks, are discussed in the [Tooling Guide](/guide/scaling-up/tooling.html).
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@letroot while I agree with your concern, I am not sure where exactly in the core docs this belongs. Do you believe we should add it to https://vuejs.org/guide/scaling-up/tooling.html#project-scaffolding ?

letroot:
> @letroot while I agree with your concern, I am not sure where exactly in the core docs this belongs. Do you believe we should add it to https://vuejs.org/guide/scaling-up/tooling.html#project-scaffolding ?

@NataliaTepluhina It could feature there as well.

Although, my main concern is with it featuring early enough in the docs. Somewhere around where the reader is told that they can "toggle between the Options API and Composition API docs mode; this section: https://vuejs.org/guide/introduction.html#api-styles

Does this make sense?

NataliaTepluhina:
@letroot thank you! I believe the `Quick Start` section would work better because we're mentioning the build tools there. Would you prefer to create a pull request to fix this or would it be better for docs team to handle this?

src/guide/quick-start.md
```diff
@@ -48,7 +48,7 @@ If you are unsure about an option, simply choose `No` by hitting enter for now.
 <span class="line"><span style="color:var(--vt-c-green);">&gt; </span><span style="color:#A6ACCD;">npm run dev</span></span>
 <span class="line"></span></code></pre></div>
 
-You should now have your first Vue project running! Here are some additional tips:
+You should now have your first Vue project running! You should note here that the components in the generated app are written in [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html) style. Here are some additional tips:
```
skirtles-code, line 51: What do you think about this?

```suggestion
You should now have your first Vue project running! Note that the example components in the generated app are written using the [Composition API](/guide/introduction.html#composition-api) and `<script setup>`, rather than the [Options API](/guide/introduction.html#options-api). Here are some additional tips:
```

skirtles-code, line 51: What do you think about this?

```suggestion
You should now have your first Vue project running! Note that the example components in the generated app are written using the [Composition API](/guide/introduction.html#composition-api) and `<script setup>`, rather than the [Options API](/guide/introduction.html#options-api). Here are some additional tips:
```

skirtles-code:
This seems to have stagnated, so I've opened #1802 to replace it.

PR #1724, opened 2022-05-16
Update component-basics.md

Danny-Devs:
## Description of Problem
Previous example didn't show how to use the new script setup defineEmits pattern in the template. This is important because there is no $ before the emit with the new emit pattern. 

## Proposed Solution
I changed the code snippet to be demonstrate how 'emit' using defineEmits would be used in markup.


bencodezen (approved):
Thanks for expanding the example to be included in the template!

PR #1730, opened 2022-05-22
Fix reference to another page on Fallthrough Attributes

tanquar:
## Description of Problem

"The next section" does not seem to exist within the page.
I guess it is located in the same submenu -> Components in depth -> Fallthrough Attributes.
(/guide/components/attrs)

## Proposed Solution

Add link and link text.
```
(which will be discussed in [its dedicated section](/guide/components/attrs)).
```

### DIFF
--- a/src/guide/components/props.md
+++ b/src/guide/components/props.md
@@ -8,7 +8,7 @@
 
 ## Props Declaration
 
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Agreed. Thanks for the fix!

PR #1732, opened 2022-05-23
Help beginners understand .mount() method better

letroot:
## Description of Problem
While I was working through this section of the tutorial, I couldn't figure out where the app's root component was. I understood that if this was a plain HTML file Vue app (that is, without the build step), then the root component would be whatever div/element I gave the #app id. But I couldn't figure out where it was for build-step mode.

## Proposed Solution
I think it helps the beginner to have the docs be explicit about the index.html file at the root of the project. It engenders experimentation and lessens the feeling of implicit magic going on.

### DIFF
--- a/src/guide/essentials/application.md
+++ b/src/guide/essentials/application.md
@@ -53,7 +53,7 @@ An application instance won't render anything until its `.mount()` method is cal
 app.mount('#app')
 ```
 
-The content of the app's root component will be rendered inside the container element. The container element itself is not considered part of the app. If you're running Vue with a build step, the container element is the div in the `index.html` file at the root of your project. You can change the id of that div from 'app' to any other name. You would have to change the argumnet passed to the `createApp.mount()` method to match the id of the div.
+The content of the app's root component will be rendered inside the container element. The container element itself is not considered part of the app. If you're running Vue with a build step, the container element is the div in the `index.html` file at the root of your project. You can change the id of that div from 'app' to any other name. You would have to change the argument passed to the `createApp.mount()` method to match the id of the div.
 
 The `.mount()` method should always be called after all app configurations and asset registrations are done. Also note that its return value, unlike the asset registration methods, is the root component instance instead of the application instance.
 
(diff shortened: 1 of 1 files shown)


src/guide/essentials/application.md
```diff
@@ -53,7 +53,7 @@ An application instance won't render anything until its `.mount()` method is cal
 app.mount('#app')
 ```
 
-The content of the app's root component will be rendered inside the container element. The container element itself is not considered part of the app.
+The content of the app's root component will be rendered inside the container element. The container element itself is not considered part of the app. If you're running Vue with a build step, the container element is the div in the `index.html` file at the root of your project. You can change the id of that div from 'app' to any other name. You would have to change the argumnet passed to the `createApp.mount()` method to match the id of the div.
```
Jothinayagan, line 56: Good work. Please check, there is a typo error in the word 'Argument'.

letroot: You're right. I'll fix right away.

letroot (commented):
I patched up a typo

skirtles-code:
Thanks for the PR!

I'm not sure I understand this change.

I think what you're calling 'a build step' is specifically a default Vite project. There are various other options, e.g. Vue CLI, Nuxt, Laravel, Inertia.

Whether a build step is used doesn't change how `mount` behaves; it's grabbing the specified element from the HTML. The main difference here is that `index.html` has been created by a scaffolding tool, rather than creating it yourself by hand. Explaining where the HTML file lives in Vite is a bit of a tangent from explaining `mount`.

skirtles-code:
Thanks for the PR!

I'm not sure I understand this change.

I think what you're calling 'a build step' is specifically a default Vite project. There are various other options, e.g. Vue CLI, Nuxt, Laravel, Inertia.

Whether a build step is used doesn't change how `mount` behaves; it's grabbing the specified element from the HTML. The main difference here is that `index.html` has been created by a scaffolding tool, rather than creating it yourself by hand. Explaining where the HTML file lives in Vite is a bit of a tangent from explaining `mount`.

skirtles-code:
I'm going to close this.

I don't feel able to merge this for the reasons I gave previously. I'd be happy to re-open it if those concerns could be addressed.

PR #1734, opened 2022-05-24
Fixed rendering about ending lines in example desc

Jinjiang:
## Description of Problem

The common IDE option like "Files: Insert Final Newline" in VSCode which add ending lines for files automatically when you save them, causes an additional empty line in the description comment from `description.txt` in examples.

## Proposed Solution

Just trim the `desc` string in `src/examples/util.ts`

## Additional Information

Original source file:
image

After saving it with the newline option in VSCode:
image

Then the render result:
image

After trimming the desc in `util.ts`:
image

Thanks.

### DIFF
--- a/src/examples/utils.ts
+++ b/src/examples/utils.ts
@@ -131,9 +131,9 @@ export function resolveSFCExample(
     raw,
     files,
     (filename, { template, composition, options, style }) => {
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
@Jinjiang thank you for the improvement 🙇🏻

PR #1737, opened 2022-05-27
Show how to call the emit function

jonlemmon:
It wasn't immediately clear to me how to call the emit function when using `<script setup>`. This should hopefully make things more clear.


skirtles-code:
Thanks for the PR!

I've got some changes to the events docs in progress that should address this in a slightly different way. I should have that up as a PR in the next day or so, but I'll leave this PR open until that one is ready so I don't forget to address the concern raised here.

jonlemmon:
Awesome, thanks @skirtles-code! 

PS - You were a massive help the other day in the Discord channel, linking me to your tabs example. What a legend! 🏆

PR #1739, opened 2022-05-28
Fix a link

xfq:
## Description of Problem

There's a broken URI fragment in the docs.

## Proposed Solution

Update the link.


skirtles-code:
Thanks for the PR!

Could you elaborate on what problem you're seeing? When I tested it the existing link seemed to work fine. There's a case mismatch in the fragment, but it still took me to the right place.

NataliaTepluhina (approved):
@xfq thank you for the fix 🙇🏻 

@skirtles-code for me, the previous link brought me to the API page but not `options` section of it, so I believe it'd be an improvement

xfq:
@skirtles-code  Sorry, I just tried opening the old URL several times and found that sometimes it would jump to the correct section, and sometimes it would only stay at the top of the page, and it might be a bug in the browser or MDN...

@NataliaTepluhina I was going to say it could be closed, but since you have already merged it, thank you!

PR #1740, opened 2022-05-29
Fixed mistype

crisarji:
From `An SFC` to `A SFC`

## Description of Problem
There is one mistype, using `An SFC` in the beginning of the sentence.

## Proposed Solution
Change it from `An SFC` to `A SFC`

## Additional Information
Nothing else


skirtles-code:
Thanks for the suggestion, but I think this is correct as is. SFC is pronounced 'ess eff see', which would be proceeded with *an* rather than *a*.

PR #1754, opened 2022-06-06
Remove @next tag from @test-library/vue package

dholde:
## Description of Problem
As of [this commit](https://github.com/testing-library/testing-library-docs/commit/ac0737e57af1b3b86cea039beef631e7e52b67f7) the testing library dependency for vue3 is the default one, so no need for the `next` tag.

## Proposed Solution
Remove the `@next` tag from the the npm install command

### DIFF
--- a/src/guide/scaling-up/testing.md
+++ b/src/guide/scaling-up/testing.md
@@ -288,7 +288,7 @@ When end-to-end (E2E) tests are run in continuous integration / deployment pipel
 In a Vite-based Vue project, run:
 
 ```sh
-> npm install -D vitest happy-dom @testing-library/vue@next
+> npm install -D vitest happy-dom @testing-library/vue
 ```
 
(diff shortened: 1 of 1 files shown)


dholde:
Closed as the issue seems to be in the testing-library/vue documentation. There the `next` tag was removed even it points to version 6 which supports vue3. See also [this PR for the testing-library docs](https://github.com/testing-library/testing-library-docs/pull/1083).

dholde:
Reopening: testing-library/vue released a new version and with that the`latest` tag now points to version 6 ([See this testing-library docs PR](https://github.com/testing-library/testing-library-docs/pull/1083)).

bencodezen (approved):
Thanks for the update! I've verified this is also the recommended command in the [testing library docs](https://testing-library.com/docs/vue-testing-library/intro/)

PR #1755, opened 2022-06-07
Reword passing of prop values to components

skirtles-code:
#1713 raised a problem with the current wording, which refers to 'dynamic props'. It isn't clear what exactly that means, as both prop names and their values could be dynamic.

I've reworded that sentence to mention 'prop values', as well as making it slightly more explicit that the sentence is talking about the example above.

### DIFF
--- a/src/guide/essentials/component-basics.md
+++ b/src/guide/essentials/component-basics.md
@@ -297,7 +297,7 @@ Then want to render a component for each one, using `v-for`:
 
 </div>
 
-Notice how we can use `v-bind` to pass dynamic props. This is especially useful when you don't know the exact content you're going to render ahead of time.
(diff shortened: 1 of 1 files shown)


yyx990803 (approved):

src/guide/essentials/component-basics.md
```diff
@@ -297,7 +297,7 @@ Then want to render a component for each one, using `v-for`:
 
 </div>
 
-Notice how we can use `v-bind` to pass dynamic props. This is especially useful when you don't know the exact content you're going to render ahead of time.
+Notice how `v-bind` is used  to pass dynamic prop values. This is especially useful when you don't know the exact content you're going to render ahead of time.
```
bencodezen, line 300:
```suggestion
Notice how `v-bind` is used to pass dynamic prop values. This is especially useful when you don't know the exact content you're going to render ahead of time.
```

bencodezen (approved):
Looks great!

PR #1756, opened 2022-06-07
Rename v-model example variables to avoid confusion

skirtles-code:
This PR comes out of a question on Vue Land. Someone was confused by one of the examples in the docs, which sets a prop using:

```
:first-name="firstName"
```

and then updates it using the event:

```
update:firstName
```

The cause of confusion was the two uses of `firstName` in camelCase. There's actually no need for these two things to have matching names, and having them match makes it seem like that is somehow significant. To add to the confusion, it is `first-name` from the first line that has to match `firstName` in the second line.

I've changed `firstName` and `lastName` to `name1` and `name2`, to avoid the naming coincidence being a potential source of confusion. The Playgrounds have been updated to match. Other examples on the same page have avoided this problem already, so didn't need to be changed.


src/guide/components/events.md
```diff
@@ -444,8 +444,8 @@ Each v-model will sync to a different prop, without the need for extra options i
 
 ```vue-html
 <UserName
-  v-model:first-name="firstName"
-  v-model:last-name="lastName"
+  v-model:first-name="name1"
```
yyx990803, line 447: `name1` and `name2` feels a bit too contrived, how about just `first` and `last`?

bencodezen:
Love the separation of v-model variable from the actual variable! 👍

src/guide/components/events.md
skirtles-code: Agreed. Never really liked `name1` and `name2` but I was having a bit of a mental block trying to think of something better.

bencodezen (approved):
Looks good to me. Thanks for updating this!

PR #1757, opened 2022-06-07
Supported custom anchors in API index

Jinjiang:
## Description of Problem

The API index page looks like this if we use custom anchors in pages, especially for translation websites that want to keep the English anchors with the translated text:

image

## Proposed Solution

Extend `headers` from `string[]` to `{ text, anchor }[]` and process the text and the anchor separately.

## Additional Information

https://github.com/vuejs-translations/docs-zh-cn/pull/294

### DIFF
--- a/src/api/ApiIndex.vue
+++ b/src/api/ApiIndex.vue
@@ -31,7 +31,7 @@ const filtered = computed(() => {
             return item
           }
           // filter headers
-          const matchedHeaders = item.headers.filter(matches)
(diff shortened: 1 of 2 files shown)


Justineo (approved):

wxsms (approved):

skirtles-code:
I haven't had chance to investigate yet, but is it possible that this PR has caused #1759? The API filter/search is very broken.

Jinjiang:
Let me take a look.

PR #1758, opened 2022-06-08
highlighted the important point in computed properties

Jothinayagan:
## Description of Problem
Beginners or someone who is not familiar with the concept should be able to notice the critical points.

## Proposed Solution
Add the highlighted information in the best practices section on the computed page.


bencodezen (approved):
I like this change! Thanks for the suggestion @Jothinayagan!

PR #1764, opened 2022-06-09
Various tweaks to how events are documented

skirtles-code:
Over the last few weeks I've encountered various examples of people misunderstanding some of the basics of component events. I've tried to address some of the potential sources of confusion in the two pages of the guide that introduce this topic: `component-basics.md` and `events.md`.

One specific problem is understanding the difference between `emit` and `$emit`: I've even encountered people who thought one or the other was a typo in the docs. As I understand it, when Evan rewrote the docs he chose to introduce `$emit` in the  `<template>` first, irrespective of which API people were using. `emit` was then introduced later, in the context of `defineEmits()`, as a way to emit events "in JavaScript", i.e. in the `<script setup>` section. This same approach to introducing the topic was used in both `component-basics.md` and `events.md`. This was lost somewhat when PR #1724 changed an existing example to use `emit` in the `<template>` section. That updated example isn't necessarily wrong, but we lose the attempt to motivate the design of `defineEmits()`. I've rolled back that example as part of this PR and instead I've expanded the text to try to explain the relationship between `$emit` and `emit` a bit more explicitly.

Some other misunderstandings I've encountered that I've tried to address:

* Trying to use `defineEmits()` inside a function.
* Not understanding how to use the `emit` function returned by `defineEmits()` (it's obvious if you understand it's the same as `$emit`, but not everyone gets that).
* Thinking that the `const emit =` part is required, even if you aren't using it. They then can't figure out how to solve the linter error about `emit` being unused. I've tried to address this implicitly, by making the first example in `events.md` use `defineEmits()` without the `const emit =` part.

### DIFF
--- a/src/guide/components/events.md
+++ b/src/guide/components/events.md
@@ -8,7 +8,7 @@
 
 ## Emitting and Listening to Events
 
-A component can emit custom events directly in template expressions (e.g. in a `v-on` handler) using the built-in `$emit` function:
+A component can emit custom events directly in template expressions (e.g. in a `v-on` handler) using the built-in `$emit` method:
 
 ```vue-html
 <!-- MyComponent -->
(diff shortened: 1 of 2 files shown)


bencodezen (approved):
Love the improvement on this. Let's ship it!

PR #1765, opened 2022-06-09
fix: typo

mordechairoth:
removed an extra (misspelled) word.

## Description of Problem
There was an extra and misspelled word "a11y" in the docs.
## Proposed Solution
Remove those extra characters.

### DIFF
--- a/src/guide/extras/rendering-mechanism.md
+++ b/src/guide/extras/rendering-mechanism.md
@@ -54,7 +54,7 @@ Vue templates are compiled into virtual DOM render functions. Vue also provides 
 
 So why does Vue recommend templates by default? There are a number of reasons:
 
-1. Templates are closer to actual HTML. This makes it easier to reuse existing HTML snippets, apply a11y best practices, style with CSS, and for designers to understand and modify.
+1. Templates are closer to actual HTML. This makes it easier to reuse existing HTML snippets, apply best practices, style with CSS, and for designers to understand and modify.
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
This is intentional. `a11y` is an abbreviation of `accessibility`. I wonder whether it would be better to write out the word in full though, especially given the lack of surrounding context.

mordechairoth:
@skirtles-code I think it might be helpful to write out the word `accessibility` in full, as you mentioned. It might make this doc more 'accessibly' to people like me and other 😉.

src/guide/extras/rendering-mechanism.md
```diff
@@ -54,7 +54,7 @@ Vue templates are compiled into virtual DOM render functions. Vue also provides
 
 So why does Vue recommend templates by default? There are a number of reasons:
 
-1. Templates are closer to actual HTML. This makes it easier to reuse existing HTML snippets, apply a11y best practices, style with CSS, and for designers to understand and modify.
+1. Templates are closer to actual HTML. This makes it easier to reuse existing HTML snippets, apply best practices, style with CSS, and for designers to understand and modify.
```
NataliaTepluhina, line 57:
```suggestion
1. Templates are closer to actual HTML. This makes it easier to reuse existing HTML snippets, apply accessibility best practices, style with CSS, and for designers to understand and modify.
```

NataliaTepluhina:
@mordechairoth @skirtles-code I believe you both are right about replacing `a11y` with accessibility here! I've added this change to the PR

PR #1766, opened 2022-06-10
Create SECURITY.md

sutisna06:


skirtles-code:
Without an explanation, I'm unclear what this PR is supposed to be.

It appears to be a file with generic text in it, not related to Vue. My best guess is that this PR was opened by mistake.

PR #1767, opened 2022-06-10
settle a source of confusion

Erichy-dev:
## Description of Problem
The transitions and animations of inner elements in Transition elements will only work if there is a duration attribute. Otherwise the whole animation doesn't work.
## Proposed Solution
So to avoid confusion I thought it much better that the code should be complete before adding embellishments.
I was confused at first since the code wouldn't work yet I couldn't figure out the issue.

### DIFF
--- a/src/guide/built-ins/transition.md
+++ b/src/guide/built-ins/transition.md
@@ -286,6 +286,14 @@ Although the transition classes are only applied to the direct child element in 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


yyx990803:
The source of the confusion is because the first two snippets are not meant to be copied as a complete example. They are not showing the styles for the outer element for brevity. You were supposed to keep reading until the playground example shows the full code.

I think it's better to add a comment in the CSS saying `/* ...other necessary CSS omitted */`.

Erichy-dev:
yah agreed please add the CSSS comment

Erichy-dev:
hello.
using "interface pp{}" and withDefaults(defineProps<pp>(), {}) raises an issue in the terminal saying that "default export of the module has or is using private..."

skirtles-code:
Opened #1801 to add the suggested comment.

PR #1771, opened 2022-06-11
Reword the introduction to class and style bindings

skirtles-code:
This change was first proposed in discussions on #1718, and replaces that PR.

The original suggestion was to replace the `:` with a `.`, but I felt that we could improve the wording to avoid the punctuation altogether.


hossein13m (approved):
based on the discussion at #1718, this enhances the documentation

NataliaTepluhina (approved):

PR #1774, opened 2022-06-14
clarify that Object.keys is stable in modern ECMAScript

Josh-Cena:
## Description of Problem

The docs claimed that `Object.keys()` is inconsistent across JS engines; this is not true in modern ECMAScript. The ordering is well-defined [in spec](https://tc39.es/ecma262/#sec-ordinaryownpropertykeys).

## Proposed Solution

Made it clearer that the ordering is only inconsistent in legacy engines (especially IE).

## Additional Information

I've done similar work in MDN docs. e.g. https://github.com/mdn/content/pull/17127

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -199,7 +199,7 @@ And another for the index:
 </div>
 
 :::tip Note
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
@Josh-Cena thank you for the fix!

PR #1777, opened 2022-06-16
Replaced codepen in Accessibility -> Placeholder with playground link

### LINKED ISSUE #1532: Example in Guide/Best Practices/Accessibility#Placeholder is missing
I just read the new guide and I think the example in the [Placeholder](https://vuejs.org/guide/best-practices/accessibility.html#placeholder) section of the accessibility page is missing. The text says:
`Looking at the following example, you can see that the Last Name placeholder which meets the color contrast criteria looks like pre-populated data:`, but there is nothing to be seen.

NataliaTepluhina:
## Description of Problem

Currently, Accessibility guide is missing code examples. This PR fixes the issue for the placeholder example

## Additional Information

I'll create a follow-up PR to replace all codepens with playground links

Close #1532


skirtles-code (commented):
I'm wondering whether we actually need to show the code for this example directly in the docs? Isn't the point of the example just to show that black placeholder text looks like pre-populated data?

NataliaTepluhina:
@skirtles-code do you recommend removing the code and the snippet and leaving only a screenshot?

Issue #233, opened 2020-07-18
Enforce styling across the repo


bencodezen:
I'm noticing that we are having some different auto-formatting styles in some of the PRs. I'm thinking we should have an opinionated styling formatting for contributors going forward so we don't have large style changes going forward.

## Proposal

This might be two pronged:

- Include an opinionated `.prettierrc` file so people's autoformat follows the team's standard consistently
- Add a pre-commit hook for linting and standardizing styles

What do you think @NataliaTepluhina @phanan @sdras?

NataliaTepluhina:
@bencodezen up for both options 👍🏻

phanan:
I'd vote for a pre-commit hook. Husky and lint-stage work pretty well for me.

sdras:
Agree with @phanan :)

schalkneethling:
Is this still on the to-do list? If so, is someone working on this or can the community contribute this? Thanks.

PR #1778, opened 2022-06-16
chore: enforce styling across the repo

xfq:
## Description of Problem

There are some different auto-formatting styles in some PRs.

## Proposed Solution

Add a pre-commit hook.

## Additional Information

See https://github.com/vuejs/docs/issues/233

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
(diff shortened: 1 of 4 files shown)


skirtles-code:
We've experimented with using automatic formatting before, but it's always proven to be problematic. The examples in the docs sometimes need the flexibility to deviate from strict formatting rules to allow for more effective communication.

I just tried running Prettier over the existing files and it does find some inconsistencies that need to be fixed. But it also changes things that really shouldn't be changed. While there are some weaker styling rules that could theoretically be applied, I don't think we'd be able to achieve that using Prettier.

I'm going to leave this open for now, as we need to fix some of the actual mistakes that Prettier finds.

yyx990803:
For the reason explained - some examples are easier to understand when hand-formatted.

PR #1782, opened 2022-06-17
fix(async-components): add example of how use it

DaniAcu:
## Description of Problem

The official docs don't provide a quick look of how use async components in the template.

## Proposed Solution

I just add a simple code example for that.

## Additional Information

For any devs, it's better be explicit with the code examples instead of just explaining in the text, because they could find what they need in the docs quickly.

### DIFF
--- a/src/guide/components/async.md
+++ b/src/guide/components/async.md
@@ -49,6 +49,11 @@ export default {
 
 </div>
 
+```vue-html{2}
+<AsyncComponent :prop="value"/>
+```
+
+
 ## Loading and Error States
 
 Asynchronous operations inevitably involve loading and error states - `defineAsyncComponent()` supports handling these states via advanced options:


NataliaTepluhina:
@DaniAcu thank you for your contribution! Maybe it would be better to move the example of component usage to line 17, right after the first JS snippet?

skirtles-code:
I'm wondering whether the example from line 37 to 48 could be converted to a full SFC, including the template? It's already most of the way there, so only a small change is required to make it a complete component. Without that extra context I'm not sure whether writing `<AsyncComponent />` would actually help to clarify anything.

The earlier examples could be interpreted as vanilla `.js` files, or as the `<script setup>` section of an SFC. I think they're intentionally written that way, but it does make it more difficult to include the template in those examples.

DaniAcu:
@skirtles-code @NataliaTepluhina Thanks for answer. It's helpful to anyone that need a quick view of the docs, to have a wrote example. If you read everything is completely understandable, but it's not exposing this information in a quick way to check it.

We also have Suspense, so maybe it's confused about if we need to use it in template or not. A example is just a way to be more explicit about.

I will move it as Natalia recommend btw 😊

DaniAcu:
Oh you already do it better in the other PR! I think it's perfect, i will close this

Thanks 😊

PR #1784, opened 2022-06-18
Fix minor language typos/improve language

peterhauke:
I removed a hyphen from "Server-Side Rendering" and hyphens from "Static Site Generation" to be consistent with other vuejs.org pages


NataliaTepluhina (approved):
@peterhauke thank you! These all are nice improvements 👍🏻

PR #1786, opened 2022-06-20
Remove hyphen from Single-File Component

peterhauke:
For an explanation and justification, see: https://github.com/vuejs/docs/pull/1785


skirtles-code:
I think we should change it the other way.

Prior to #1785 there were only about 5 uses without a hyphen and over 30 with a hyphen.

I also think that hyphenating is more consistent with the usual rules for English hyphenation.

PR #1787, opened 2022-06-20
Fix minor language typo/improve language

peterhauke:
Remove hyphen from Single-File Component, for an explanation and justification, see: #1785

There is almost no difference between "in order to" and "to". "in order" can be removed almost always without changing the meaning.

"in full details" should be "in full detail"


peterhauke:
Removing the hyphen from Single-File Component is no longer correct, see: #1788, so this PR is no longer necessary.

PR #1791, opened 2022-06-22
Fix typos

peterhauke:


src/guide/essentials/template-syntax.md
```diff
@@ -198,7 +198,7 @@ Another example is the `v-on` directive, which listens to DOM events:
 <a @click="doSomething"> ... </a>
 ```
 
-Here the argument is the event name to listen to: `click`. `v-on` is one of the few directives that also have a corresponding shorthand, with its shorthand character being `@`. We will talk about event handling in more detail too.
+Here the argument is the event name to listen to: `click`. `v-on` is one of the few directives that also has a corresponding shorthand, with its shorthand character being `@`. We will talk about event handling in more detail too.
```
skirtles-code, line 201: I'm not sure this change is correct. I believe the verb here should match 'the few directives', which is plural.

skirtles-code (changes requested):
Thanks for the PR! I've included my feedback below.

src/guide/essentials/template-syntax.md
peterhauke: Thanks, @skirtles-code

Maybe the following is clearer and sufficient in this case:

v-on has a corresponding shorthand, which is the @ character.

I think "is one of the few directives that" is redundant in the original sentence.

skirtles-code (approved):

PR #1798, opened 2022-06-25
Add: Callback Flush Timing - About the flush: sync

niceplugin:
Add content that watchers.md : Callback Flush Timing - About the flush: sync option

Among the options used for watch, `flush: 'sync'` was not explained in the guide or API documentation, so I added it.

It need to check because my English is not good.

Hope this helps.

### DIFF
--- a/src/guide/essentials/watchers.md
+++ b/src/guide/essentials/watchers.md
@@ -342,6 +342,57 @@ watchPostEffect(() => {
 
 </div>
 
+The `flush: 'pre'|'post'` option buffers the callback so that it is only called once at the end, even if there are multiple state changes in the same "tick".
+
+The `flush: 'sync'` option should be used if the callback needs to be invoked synchronously on every state change multiple times within the same tick.
+
+<div class="options-api">
+
+```js
+export default {
+  data: () => ({ count: 0 }),
+  watch: {
+    count: {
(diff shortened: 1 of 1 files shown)


niceplugin:
It looks fine, but I only got the review label.
Is there anything more I should do?

skirtles-code:
Thanks for the PR and sorry for the delay in getting back to you.

`flush: 'sync'` poses a documentation challenge. It's an 'escape hatch' feature, intended to be used in specific edge cases by advanced users. Such features are typically not covered in the main guides, or at least not the guides aimed at beginners. Usually such features are only documented in the API reference, or even omitted from the docs entirely in extreme cases. In this specific case we have the added complication that `flush: 'pre'` and `flush: 'post'` are mainstream, commonly used features that very much need to be covered in the basic guides.

I did raise this internally and the feeling seemed to be that we want to keep `'sync'` out of the main guide. It's already mentioned in the API reference, but as you correctly pointed out there's no explanation about what it does.

Adding an explanation in the API reference is also a bit of a pain because the `flush` setting is relevant to various entries.

I've tried to strike a balance in #1895 (now merged), providing a brief introduction to `flush: 'sync'` and also trying to warn people away from using it. The explanation is under `watchEffect`, with links added from other places that have a `flush` setting.

PR #1804, opened 2022-06-27
Setting indents for CompositionAPI code block

Reza-Javadpour:
Just for better code styles :)


skirtles-code:
Thanks for the suggestion, but I believe we consistently don't indent the `<script>` section throughout the docs.

PR #1783, opened 2022-06-18
added dynamic component v-model note

calebwaldner:
## Description of Problem
The documentation states using native HTML elements works with the dynamic `<component>` element, but doesn't state that `v-model` may not behave as expected.

## Proposed Solution
Added a brief note stating that `v-model` may not work as expected and also added a brief explanation on why.

## Additional Information
My explanation of why `v-model` doesn't work was based on how I understood the comments made on this issue. https://github.com/vuejs/core/issues/5020. If my explanation is not accurate I apologize.

PR #1807, opened 2022-06-27
Explain v-model with :is and native elements

skirtles-code:
This is an attempt to address the problem raised in #1783.

In brief, `<component :is="'input'" v-model="...">` won't work because the template compiler assumes it is a component `v-model`, not an `<input>` `v-model`.

#1783 attempted to address this in the guide. However, I think it would be better to address this in the API Reference instead. The API Reference already covers some other edge cases involving `<component :is>` that aren't directly mentioned in the guide. It also gives us the freedom to discuss the problem at greater length, without worrying about derailing the narrative flow of the guide.

In particular, the section of the guide that covers `<component :is>` comes before the section on component `v-model`, so it's trying to explain something that the reader isn't expected to have encountered yet.

I don't think this edge case is common enough to justify inclusion in the **Essentials** section of the guide.

### DIFF
--- a/src/api/built-in-special-elements.md
+++ b/src/api/built-in-special-elements.md
@@ -90,6 +90,24 @@ A "meta component" for rendering dynamic components or elements.
 
   Registration is not required if you pass the component itself to `is` rather than its name, e.g. in `<script setup>`.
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
I agree that this approach is better - `Essentials` isn't the best place to have this deep dive into an edge case

PR #1810, opened 2022-06-28
docs: update `inject` API types

webfansplz:
## Description of Problem

Related to [vuejs/core/issues/6194](https://github.com/vuejs/core/issues/6194)

## Proposed Solution

Update docs

## Additional Information

https://github.com/vuejs/core/blob/8edf4b344179d0a8d3c1090ce63f466f4e787976/packages/runtime-core/src/apiInject.ts#L31

### DIFF
--- a/src/api/composition-api-dependency-injection.md
+++ b/src/api/composition-api-dependency-injection.md
@@ -52,12 +52,16 @@ Injects a value provided by an ancestor component or the application (via `app.p
   function inject<T>(key: InjectionKey<T> | string): T | undefined
 
   // with default value
-  function inject<T>(key: InjectionKey<T> | string, defaultValue: T): T
+  function inject<T>(
+    key: InjectionKey<T> | string,
+    defaultValue: T,
+    treatDefaultAsFactory?: false
+  ): T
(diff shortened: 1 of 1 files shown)


src/api/composition-api-dependency-injection.md
```diff
@@ -52,12 +52,16 @@ Injects a value provided by an ancestor component or the application (via `app.p
   function inject<T>(key: InjectionKey<T> | string): T | undefined
 
   // with default value
-  function inject<T>(key: InjectionKey<T> | string, defaultValue: T): T
+  function inject<T>(
```
NataliaTepluhina, line 55: **question**: I am not sure why do we add a third parameter here if we listed the factory below 🤔

NataliaTepluhina:
@webfansplz thank you for creating the PR! However, you have probably noticed in https://github.com/vuejs/core/issues/6194 that even when provided default value is a function, we still evaluate the `treatDefaultAsFactory`as `false` so I am not sure this addition resolves the reported issue. I believe the current documentation is correct and doesn't require an addition.

PR #1812, opened 2022-07-01
Move security warning to the top

mvikrammenon:
## Description of Problem
Present placement of the warning message at the bottom, does not help in highlighting the risk of its usage upfront. 

## Proposed Solution
Moving this section to the top would do better job of highlighting the risk.


NataliaTepluhina:
Thank you for your contribution! When moving the warning to the top, we start with saying `use X` before X is actually explained. I believe we should leave the warning where it currently is.

PR #1814, opened 2022-07-05
Update accessibility.md

valcosmos:


NataliaTepluhina:
@valcosmos thank you for the contribution! We use a singular `id` here as a specific term, and the syntax of `id`s is intended. It's similar to uppercased cases: ID - IDs (note that the last "s" is not uppercased)

PR #1825, opened 2022-07-10
Add vuemastery banner

pieer:


.vitepress/theme/index.ts
```diff
@@ -24,6 +27,5 @@ export default Object.assign({}, VPTheme, {
     app.provide('prefer-composition', preferComposition)
     app.provide('prefer-sfc', preferSFC)
     app.provide('filter-headers', filterHeadersByPreference)
-    app.component('VueSchoolLink', VueSchoolLink)
```
yyx990803, line 27: FYI this should not be removed as this component is needed in many pages. Fixed in 5281fd3

yyx990803:
@pieer I've noticed that the banner's dismiss button isn't working as intended (it fails to close the banner and instead always goes to the link). See live deployment: https://62cce165f7ad1f000ab5fc04--vuejs.netlify.app/

I've pinned the deployment to the version without the banner for now until this is fixed.

PR #1826, opened 2022-07-11
Fixed a typo (changed an to a, due to consonant)

llamington:
## Description of Problem
The docs read 'Vue is able to detect when an reactive array's mutation methods are called and trigger necessary updates'.

## Proposed Solution
This should read 'Vue is able to detect when *a* reactive array's mutation methods are called and trigger necessary updates'


skirtles-code (approved):
Thanks!

PR #1834, opened 2022-07-16
Add key for v-for

ruofee:
## Description of Problem

A v-for example is missing key

## Proposed Solution

Add key for v-for


skirtles-code:
Thanks for the suggestion.

I think the `key` has been intentionally omitted here to avoid it being a distraction.

While there is a best practice for including a `key` in this scenario, it isn't strictly required. You can see from the Playground examples that it works fine without it.

PR #1850, opened 2022-07-22
Fix vue mastery banner z-order

### LINKED ISSUE #1846: Navbar on top of Vue mastery banner on mobile
Only happens when I scroll

hichemfantar:
Looks like there's already a CSS variable to properly handle banner z-index so I switched it to that.
closes #1846


NataliaTepluhina (approved):
Thank you for the fix!

hichemfantar:
Anytime!

PR #1856, opened 2022-07-24
update introduction.md

salimzy:


skirtles-code:
As far as I can tell this change doesn't affect the rendered output and it doesn't improve the source. No explanation has been provided on the PR to justify the change.

Closing.

PR #1858, opened 2022-07-25
Vuchl patch crud example

vuchl:
## Description of Problem
there are unnecessary semicolons in the example code

## Proposed Solution
remove semicolons


yyx990803:
These semicolons are added automatically by prettier - no need to change them manually.

PR #1860, opened 2022-07-26
push always the max +1 number

sys-tps:
when you push an item the list will increment the max number

## Description of Problem
previously when pushing a item in the list then reversing the list and popping an item and finally pushing again resulting incorrect increment, being that length would be less than the max number

## Proposed Solution
added Math.max() function


skirtles-code:
Thanks for the PR!

I'm not sure I understand the motivation behind this change. The current implementation can lead to the same number being included in the list multiple times, but as far as I can tell there isn't a problem with having repeated numbers.

I don't think there's anything wrong with the `Math.max` version, but I don't currently see any reason why it would be preferable to the existing functionality.

PR #1862, opened 2022-07-26
docs(cookbook): unbound breakpoint

stfujnkk:
Debugging cannot be performed normally with previous configuration files. There will be an unbound breakpoint exception


skirtles-code:
This PR is against an old branch, which should probably be deleted. Making a change to that branch isn't going to impact the deployed docs. The current documentation no longer includes the cookbook.

PR #1863, opened 2022-07-27
Update overview.md

Joakal:
The plugin is NOT recommended for take-over as per original maintainer of Volar https://github.com/johnsoncodehk/volar/discussions/471#discussioncomment-1361405

## Description of Problem

This plugin is NOT needed for Vue TS Takeover mode as per original maintainer of Volar

## Proposed Solution

Suggest to use the correct plugin only.

### DIFF
--- a/src/guide/typescript/overview.md
+++ b/src/guide/typescript/overview.md
@@ -57,7 +57,7 @@ See also:
 
 > This section only applies for VSCode + Volar.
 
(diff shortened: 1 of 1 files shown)


sodatea:
Please read on… The next paragraph immediately says that Takeover Mode is a better solution.

We suggest TypeScript Vue Plugin (Volar) by default because it's easier to set up. And the whole paragraph in the documentation is about its mechanism and why it is inferior to Takeover Mode.

Meanwhile, once Takeover Mode is on, TypeScript Vue Plugin (Volar) would be ignored.
So it's better to keep it because it doesn't hurt the performance of Takeover Mode, while still improves the overall experience when the user doesn't take the extra steps to turn on Takeover Mode in the workspace.

PR #1864, opened 2022-07-27
These code snippets don't match their description

3-2:
## Description of Problem
The document says 'When using Vue _without_ a build step' under [the section 'In-DOM Root Component Template'](https://vuejs.org/guide/essentials/application.html#in-dom-root-component-template). But the first line in the js code snippet seems to require a build step such as Vite?
`import { createApp } from 'vue'`
Since the module name 'vue' doesn't start with a path sign, the code snippet should require a build step such as Vite. ([Reference](https://stackoverflow.com/search?q=Relative+references+must+start+with+either))
## Proposed Solution
Therefore, I edited these two code snippets with reference to [the section 'Without Build Tools'](https://vuejs.org/guide/quick-start.html#without-build-tools) in the article 'Quick Start' in Vue documentation.
I think my edit is proper for documentation readers regards the consistence.
## Additional Information
By the way, I am also a Vue beginner. Let me know if I am wrong.
[See also](https://forum.vuejs.org/t/am-i-right-there-must-be-a-build-step-such-as-vite-as-long-as-you-see-this-code-wherever-import-from-vue/131526).

### DIFF
--- a/src/guide/essentials/application.md
+++ b/src/guide/essentials/application.md
@@ -62,13 +62,15 @@ The `.mount()` method should always be called after all app configurations and a
 When using Vue without a build step, we can write our root component's template directly inside the mount container:
 
 ```html
+<script src="https://unpkg.com/vue@3"></script>
+
 <div id="app">
   <button @click="count++">{{ count }}</button>
 </div>
 ```
 
 ```js
-import { createApp } from 'vue'
+const { createApp } = Vue
 
 const app = createApp({
   data() {


skirtles-code:
Thanks for the PR!

I believe the current use of an `import` is intentional. It's assuming that you're using an import map, as introduced in <https://vuejs.org/guide/quick-start.html#without-build-tools>. I can see that as being potentially confusing, though.

I'm not sure that section in `application.md` needs to say 'without a build step'. Using an in-DOM template isn't limited to applications without a build step. Instead I think it should probably mention that it requires a build of Vue that supports template compilation at runtime.

Thoughts?

3-2:
> I believe the current use of an import is intentional.

Yes. I know the solution of `importmap`, this feature is mentioned in 'Quick Start', too. With `importmap`, the code `import { createApp } from 'vue'` can run up without error.
The section description is correct technically, while it's a bit confusing for Vue learners, maybe. I think we could consider editing the article to make it clearer.
> Using an in-DOM template isn't limited to applications without a build step.

Please excuse my lack of understanding on the term 'in-DOM template'. Does it refer to write Vue project in plain HTML, as opposed to SFC? ([Reference](https://stackoverflow.com/questions/68578348/what-exactly-is-a-dom-template-in-vue-js)) That's to say the build step such as Vite and Webpack can be used in plain HTML? (I'm sorry that I don't understand much about bundler)
> Instead I think it should probably mention that it requires a build of Vue that supports template compilation at runtime.

Do you know [which Vue build](https://unpkg.com/browse/vue@3.2.37/dist/) supports this feature?

skirtles-code:
I've reworded this section in #2332 to try to make it a bit clearer.

Thanks again for pointing out this problem.

PR #1866, opened 2022-07-27
Add Fig as an autocomplete method

brendanfalk:
[Fig](https://fig.io) provides autocomplete for 300+ CLI tools including Vue. It looks like this:

image

We think this will help users become more efficient with Vue CLI, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!


bencodezen:
Thanks for the PR @brendanfalk and love that Fig supports Vue CLI in the terminal!

At this time though, the docs are focused on highlighting official tools and libraries that have become standard use in the Vue ecosystem. As a result, we cannot accept this PR at this time. 

That said, I'm sure the community would love to see how Fig can benefit their Vue workflow through videos or what not, so I'd encourage you and the team to share that in the official Vue Discord to help the community. Thanks again!

Issue #1621, opened 2022-03-16
Typescript with Composition API: Document use of PropType


LinusBorg:
### Preface 

Our Typescript docs are split into two pages for Composition API and Options API, which is fine as there are different things to be aware about when using these APIs with TS.

However at the same time, we should document similarities in both sections equally.

### We explain the use of `PropType` only in the Options API TS guide.

We explain how to use it here in the second code snippet: https://vuejs.org/guide/typescript/options-api.html#typing-component-props

However we don't mention it on the [Composition API Version of that page](https://vuejs.org/guide/typescript/composition-api.html#typing-component-props) anywhere. If one were to read this page alone, as they are not using Options API, they would not be aware they need to do this to use an external type:

```html
<script setup> 
import type { Categories } from './types'
import { PropType } from 'vue'

defineProps({
  category: {
    type: String as PropType<Categories>
  }
})
</script>
```

## Proposed solution

Right after the first simple code snippet explaining runtime props declaration, we should insert and explanation of `PropType` before we go on to dive into `defineProps<{ }>()`

julian-eckhardt:
I created a PR (https://github.com/vuejs/docs/pull/1868) matching this issue and am open for any comments.
I'd really like to get this merged as it would IMHO be an improvement over the current situation.

PR #1868, opened 2022-07-29
TS Composition API: Mention explicit PropType on prop runtime declaration

julian-eckhardt:
## Description of Problem

Right now, the guide does not mention a way to have explicit prop type declarations and default values at the same time.
The only (close) solution, Reactivity Transform, is experimental and opt-in.

## Proposed Solution

Mentioning the `as PropType <Foo>` syntax, as in this PR.

## Additional Information

I'm filing this PR not only to contribute to the docs and hopefully help others, but also to verify my own approach on the problem. I'd be keen on maintainers' input on this.

### DIFF
--- a/src/guide/typescript/composition-api.md
+++ b/src/guide/typescript/composition-api.md
@@ -6,6 +6,7 @@
 
 ### Using `<script setup>`
 
+#### Runtime Declaration
 When using `<script setup>`, the `defineProps()` macro supports inferring the props types based on its argument:
 
 ```vue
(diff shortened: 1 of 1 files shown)


src/guide/typescript/composition-api.md
```diff
@@ -22,6 +22,22 @@ props.bar // number | undefined
 
 This is called "runtime declaration", because the argument passed to `defineProps()` will be used as the runtime `props` option.
 
+Prop types can also be explicitly declared when using runtime declaration:
+
+```vue
+<script setup lang="ts">
+const props = defineProps({
+  foo: {
+    type: String as PropType<string>,
```
tony19, line 31: This is a redundant typing as the type `string` can already be inferred from the `String` constructor (same for `number` and `Number` below). I recommend showing an example where `PropType` is necessary. For example:

```typescript
type: Array as PropType<string[]>
```

skirtles-code:
There is a related issue for this: <https://github.com/vuejs/docs/issues/1621>.

src/guide/typescript/composition-api.md
skirtles-code: Maybe reuse the `PropType` example from the Options API page, ported to `<script setup>`?

julian-eckhardt: @tony19 I know, but for `Object` that wouldn't be the case. I just wanted to show how it works, as a 1:1 copy of the previous example.

tony19 (approved):

skirtles-code:
I gave this some more thought and decided it might be better to take a slightly different approach to documenting this. There are more details on #1621, which has now been merged.

I'm going to close this PR, but if anyone has any further thoughts on this topic then feel free to share them on #1621 or open a new issue.

Thanks.

PR #1870, opened 2022-07-29
Adding `<meta name="theme-color" />` for docs

sanjaiyan-dev:
Adding  `<meta name="theme-color" />` for the docs. For more info https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

I choosed #3c8772 as a theme-color hence it looks nice in dark mode and white mode too, and that colour in presents in VUE JS logo too.

_Sorry if I made any mistakes :(_

### DIFF
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -560,6 +560,7 @@ export default defineConfigWithTheme<ThemeConfig>({
   scrollOffset: 'header',
 
   head: [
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
@sanjaiyan-dev thank you, I agree the color looks nice on both themes 👍🏻

PR #1871, opened 2022-07-30
fix: Minor grammatical error in props.md

mst101:
## Description of Problem
...there isn't as much practical benefits for...

## Proposed Solution
...there isn't as much practical benefit in...


skirtles-code (approved):
Looks good, thanks!

PR #1877, opened 2022-08-01
Update event-handling.md

thiensubs:
Add demo capture Key - to show which key press and details of KeyboardEvent Object.

image

## Description of Problem
We have a detail of keys in this documentation, but it is hard to understand at all, and how to implement it in detail.
## Proposed Solution
Easy to understand how to handle the key.
## Additional Information
N/A


skirtles-code:
Thanks for the PR! Sorry for the delay in getting back to you.

I've discussed this with the team and we feel that the example you've added is showing native browser functionality, not anything specific to Vue. We try to limit the scope of the docs to Vue itself where possible.

PR #1883, opened 2022-08-03
Change v-model for multiline example

patryk406:
## Description of Problem

The multiline example uses the message ref for the v-model, so the example above "message" changed if we wrote something in the multiline section input.

## Proposed Solution

change message ref in v-model and expression for multilineText(this one is not used anywhere so probably should be used there)


skirtles-code (approved):
Looks good. Thanks!

PR #1880, opened 2022-08-03
Shuffle team members

Jinjiang:
## Description of Problem

Shuffle team members like the v2 doc site.

## Proposed Solution

Provide a `shuffleMembers()` function to process the data ahead.

## Additional Information

It's on the TODO list of `@vue/theme`. However, now I feel it's better done here. Thanks.
https://github.com/vuejs/theme/issues/36

PR #1886, opened 2022-08-04
fix(member): fix the shuffle bug

KiritaniAyaka:
fix [the member shuffle bug](https://github.com/vuejs/docs/pull/1880#issuecomment-1204698307)


KiritaniAyaka:
This method will fix the bug but caused the first member of  EmeritiMember and PartnerMember to be pinned.

Sorry... 😢

PR #1889, opened 2022-08-04
Fix hiring link of Vehikl

J2TEAM:
## Description of Problem
The hiring link show error 404 due to missing protocol.

image

image

## Proposed Solution
Add protocol to the link.


skirtles-code (approved):
Thanks for the fix!

PR #1553, opened 2022-02-16
Update custom-elements-everywhere score

rictic:
See https://github.com/vuejs/core/issues/5401

PR #1890, opened 2022-08-05
Web Components: Add Event Handling Caveat

### LINKED ISSUE #1708: Vue does not perform 100% anymore at Web Components test
On this page: https://vuejs.org/guide/extras/web-components.html#using-custom-elements-in-vue it is said that vue performs 100% at Custom Elements Everywhere test. 
It seems that the test has changed now. 3 tests failed. Until vue comes back to 100% this line should probably be updated or removed.

aztalbot:
## Description of Problem

The custom elements section currently includes an incorrect score from Custom Elements Everywhere. Even if Custom Elements Everywhere decided to accept custom directives as satisfactory (which they have not), I don't think the score is all that relevant. Bottom line: custom elements work, and there are some caveats.

The caveat around listening for events with capital letters currently is not documented, and neither is the recommended solution.

## Proposed Solution

- Remove the sentence referring to Custom Elements Everywhere
- Document the caveat around listening to events with names other than lowercase and kebab-case names
- Document the recommended solution of using a custom directive

## Additional Information

Related issue: https://github.com/vuejs/core/issues/5401
Related closed PR: https://github.com/vuejs/docs/pull/1553
Related RFC discussion: https://github.com/vuejs/rfcs/discussions/451
Closes #1708

### DIFF
--- a/src/guide/extras/web-components.md
+++ b/src/guide/extras/web-components.md
@@ -6,7 +6,7 @@ We consider Vue and Web Components to be primarily complementary technologies. V
 
 ## Using Custom Elements in Vue
 
-Vue [scores a perfect 100% in the Custom Elements Everywhere tests](https://custom-elements-everywhere.com/libraries/vue/results/results.html). Consuming custom elements inside a Vue application largely works the same as using native HTML elements, with a few things to keep in mind:
+Consuming custom elements inside a Vue application largely works the same as using native HTML elements, with a few things to keep in mind:
 
 ### Skipping Component Resolution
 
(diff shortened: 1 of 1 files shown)


src/guide/extras/web-components.md
```diff
@@ -75,6 +75,30 @@ However, there could be rare cases where the data must be passed as a DOM proper
 <my-element .user="{ name: 'jack' }"></my-element>
 ```
 
+### Handling Events with Capital Letters
+
+The `v-on` directive handles lowercase and kebab-case events dispatched from custom elements. However, the `v-on` directive does not support listening for events whose names include capital letters. For example, adding `v-on:myInput` to a custom element adds a listener for events with the name `my-input`, which does not handle an event with the name `myInput`.
+
+If your app uses custom elements that dispatch events with capital letters, you can use a custom directive to add the correct event listener to the element.
+
+#### Example Custom Directive to Handle Specific Events
```
skirtles-code, line 84: I'm inclined to go straight from the previous paragraph into the example, without the extra heading.

I think we can also shorten **If your app uses** to just **If you use**

```suggestion
If you use custom elements that dispatch events with capital letters, you can use a custom directive to add the correct event listener to the element:
```

skirtles-code (commented):
Thanks for the PR! Sorry for the delay getting back to you.

LinusBorg:
LGTM

yyx990803:
Hold it for now - I believe this should be considered a bug and is fixable.

LinusBorg:
> Hold it for now - I believe this should be considered a bug and is fixable.

@yyx990803 For camelCase it mighr be, i think i even have an open PR for that? Not sure 😅

But we map `onUpperCase` props to `upperCase` events, so we cant real differentiate between a camelCase and PascalCase Listener.

PR #1899, opened 2022-08-10
Update animation.md

Yechuanjie:
in Animating with Watchers, the template code  `tweened.number.toFixed(0)` is wrong, excepted `tweened.toFixed(0)`.

### DIFF
--- a/src/guide/extras/animation.md
+++ b/src/guide/extras/animation.md
@@ -190,7 +190,7 @@ export default {
 
 ```vue-html
 Type a number: <input v-model.number="number" />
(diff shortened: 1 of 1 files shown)


skirtles-code (changes requested):
Thanks for the PR!

This PR fixes it for the Options API, but it's now broken for the Composition API. I think we need to move the template code inside the `<div>` and create a separate snippet for each API.

skirtles-code:
I've opened #1938 to apply the fix I proposed above.

PR #1904, opened 2022-08-10
Fix Testing Library example

NoelDeMartin:
## Description of Problem
I think the current example is wrong. It seems to render the component twice and uses different variable names for it (`Component` vs `Stepper`).

## Proposed Solution
I guess the proper solution is to refer to the component as `Stepper`, like in the rest of the examples, and render only once.


skirtles-code (approved):
Thanks for the PR!

The change looks good to me, but I have limited experience with this library, so I'll hold off merging for now so other people can take a look.

PR #1914, opened 2022-08-14
CRUD Example Patch

SparklingWater45:
## Description of Problem
A semicolon is in-front of the watch function for the CRUD example.

## Proposed Solution
Remove semicolon in options and composition API examples.


skirtles-code:
Thanks for the PR, but these semicolons have been included intentionally. Please see #1858 for details.

PR #1915, opened 2022-08-15
chore: add spaces around slashes

jay-es:
Changed `getter/setters` to `getter / setters` as in the following commit:
https://github.com/vuejs/docs/commit/c4ba9cb#diff-69907cc80676ccbf5f96eaeafae0365ca25493210a1fcaa5d6325cf02d59f834


xfq:
If the slash divides two words, there is usually no space (see [CMOS](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Punctuation/faq0096.html), for example).

If this is the house style of Vue.js, maybe it can be mentioned in https://github.com/vuejs/docs/blob/main/.github/contributing/writing-guide.md

PR #1919, opened 2022-08-17
fix: update reactivity-in-depth.md

Hazel-Lin:
Vue3 -> Vue2


skirtles-code:
Thanks for the PR, but I believe the existing text is correct. It does mean Vue 3.

Hazel-Lin:
So, Is it mean use Options API in Vue3?

skirtles-code:
I'm not sure what you're asking. The Options API is available with Vue 3 and it is built on top of the Composition API.

PR #1920, opened 2022-08-17
fix withDefaults demo

liuhanqu:
The  `labels`  is  `string[]`  in  Prop Interface, but the default value is a function

### DIFF
--- a/src/api/sfc-script-setup.md
+++ b/src/api/sfc-script-setup.md
@@ -302,7 +302,7 @@ export interface Props {
 
 const props = withDefaults(defineProps<Props>(), {
   msg: 'hello',
-  labels: () => ['one', 'two']
+  labels: ['one', 'two']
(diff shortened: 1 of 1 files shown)


src/api/sfc-script-setup.md
```diff
@@ -302,7 +302,7 @@ export interface Props {
 
 const props = withDefaults(defineProps<Props>(), {
   msg: 'hello',
-  labels: () => ['one', 'two']
+  labels: ['one', 'two']
```
tony19, line 305: The original text is correct. The default value for an array prop must be a factory function. You'll notice an [ESLint error](https://eslint.vuejs.org/rules/require-valid-default-prop.html) about this as well.

liuhanqu: Appreciate for you replay

PR #1924, opened 2022-08-19
Make the interaction more reasonable.

zhangenming:


skirtles-code:
Thanks for the PR and sorry for the delay in responding.

Related: #1600

The main advantage of `@change` is that it handles emptying a field better. With `@input` it jumps to `0`, interfering with whatever editing the user was performing. I've tried various ways to address this but I'm struggling to come up with one that makes all the interactions clean without cluttering up the example code.

skirtles-code:
I don't think we can merge this, but I've opened #2143 if anyone wants to make suggestions for how we could improve this example.

PR #1929, opened 2022-08-19
Fix confusing and unnecessary method access in comments

mrkhosravian:
## Description of Problem
Confusing and unnecessary method access `vm.submit()` in Key Modifiers section.

## Proposed Solution
Replace `vm.submit()` with simple `submit` method name.


skirtles-code (approved):
Thanks for the fix!

PR #1934, opened 2022-08-25
Correct docs

TheJaredWilcurt:
Example: https://jsfiddle.net/b4cyeht6

This works the same in Vue 2 and Vue 3.

### DIFF
--- a/src/guide/components/props.md
+++ b/src/guide/components/props.md
@@ -610,4 +610,4 @@ export default {
 
 </div>
 
-The casting rules for `Boolean` will apply if it occurs prior to `String`.
+If the array of types includes `Boolean` then the casting rules for `Boolean` will apply, unless `String` appears earlier in the array.
--- a/src/guide/components/props.md
+++ b/src/guide/components/props.md
@@ -593,7 +593,7 @@ When a prop is declared to allow multiple types, e.g.
 
 ```js
 defineProps({
-  disabled: [String, Boolean, Number]
+  disabled: [Boolean, Number]
 })
 ```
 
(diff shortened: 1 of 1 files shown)


src/guide/components/props.md
```diff
@@ -593,7 +593,7 @@ When a prop is declared to allow multiple types, e.g.
 
 ```js
 defineProps({
-  disabled: [Boolean, Number]
+  disabled: [String, Boolean, Number]
```
skirtles-code, line 596: I'm not sure why these examples need to change. They seem fine to me as they were.

src/guide/components/props.md
```diff
@@ -603,11 +603,11 @@ defineProps({
...
-    disabled: [Boolean, Number]
+    disabled: [String, Boolean, Number]
   }
 }
 ```
 
 </div>
 
-The casting rules for `Boolean` will apply regardless of type appearance order.
+The casting rules for `Boolean` will apply if it occurs prior to `String`.
```
skirtles-code, line 613: I agree that the wording needs to change, but I think we need to be a little more precise here. The proposed change would imply that `disabled: [Boolean, Number]` would not use the casting rules, because `Boolean` is not prior to `String`. Perhaps something like this?

```suggestion
If the array of types includes `Boolean` then the casting rules for `Boolean` will apply, unless `String` appears earlier in the array.
```

skirtles-code (changes requested):
Thanks for the PR! You're right, the current wording is incorrect.

I've provided some feedback below on your proposed changes.

PR #1943, opened 2022-08-29
Refine internal links

itsyuxuan:
## Description of Problem
Those links are supposed to be internal, but they are written in URL, which affects the click action and the performance.

## Proposed Solution
Change the URLs to internal router links.


skirtles-code (approved):
Thanks!

PR #1946, opened 2022-08-31
docs: document mixing named and default slots

noook:
https://github.com/vuejs/core/issues/6549

## Description of Problem

Regarding [this issue](https://github.com/vuejs/core/issues/6549), the compiler doesn't inform what's the real problem with mixing named slot and default. This PR updates the documentation to explain how to avoid this error. The compiler error still needs to be clearer.

## Proposed Solution

Provide an example on how to use default slots along named slots

## Additional Information

—

### DIFF
--- a/src/guide/components/slots.md
+++ b/src/guide/components/slots.md
@@ -415,6 +415,7 @@ Note the `name` of a slot won't be included in the props because it is reserved 
 If you are mixing named slots with the default scoped slot, you need to use an explicit `<template>` tag for the default slot. Attempting to place the `v-slot` directive directly on the component will result in a compilation error. This is to avoid any ambiguity about the scope of the props of the default slot. For example:
 
 ```vue-html
+<!-- This template won't compile -->
 <template>
   <MyComponent v-slot="{ message }">
     <p>{{ message }}</p>
--- a/src/guide/components/slots.md
+++ b/src/guide/components/slots.md
@@ -276,36 +276,6 @@ Now everything inside the `<template>` elements will be passed to the correspond
 
 </div>
 
-Note that if you are mixing named slots with the default slot, you need to explicitly define the default template. Otherwise, it would hint that the data of the default slot would be available in the other slots scopes.
-
-```vue-html
-<template>
-  <BaseLayout v-slot="{ message }">
-    <p>{{ message }}</p>
-    <template #footer>
-      <!-- message belongs to the default slot, and is not available here -->
-      <p>{{ message }}</p>
-    </template>
-  </BaseLayout>
-</template>
-```
-
-Instead, you will need to write it as follows:
-```vue-html
-<template>
-  <MyComponent>
-    <!-- Use explicit default slot -->
-    <template #default="{ message }">
-      <p>{{ message }}</p>
-    </template>
-
-    <template #footer>
-      <p>Here's some contact info</p>
-    </template>
-  </MyComponent>
-</template>
-```
-
 Again, it may help you understand named slots better using the JavaScript function analogy:
 
 ```js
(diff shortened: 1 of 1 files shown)


skirtles-code (commented):
Thanks for the PR!

I think there may be a couple of problems with explaining it this way:

1. It appears to contradict the example that immediately precedes it, which mixes named slots and the default slot quite happily, without using `#default`. The problem only occurs when `v-slot`/`#default` is explicitly used on the parent tag. If it is omitted then everything works fine.
2. This new example is using a scoped slot before the concept of scoped slot has been introduced.

I think we may need to move this explanation further down, because scoped slots are the key reason why somebody would run into this problem in the first place. Maybe in the section about **Named Scoped Slots**?

noook:
That's what I did in my first commit, I finally decided to move it earlier after mixing the slots.

I'll do that !

src/guide/components/slots.md
```diff
@@ -412,6 +412,36 @@ Passing props to a named slot:
 
 Note the `name` of a slot won't be included in the props because it is reserved - so the resulting `headerProps` would be `{ message: 'hello' }`.
 
+Also note that if you are mixing named scoped slots with the default scoped slot, you need to explicitly define the default template. Otherwise, it would hint that the data of the default scoped slot would be available in the other slots scopes.
```
skirtles-code, line 415:
```suggestion
If you are mixing named slots with the default scoped slot, you need to use an explicit `<template>` tag for the default slot. Attempting to place the `v-slot` directive directly on the component will result in a compilation error. This is to avoid any ambiguity about the scope of the props of the default slot. For example:
```

src/guide/components/slots.md
```diff
@@ -412,6 +412,36 @@ Passing props to a named slot:
...
+
+```vue-html
```
skirtles-code, line 417: It's always a bit dangerous to have examples of incorrect code. People tend to skim over the text and just copy the examples. I think a note at the top of the example can help to mitigate that.

```suggestion
```vue-html
<!-- This template won't compile -->
```

src/guide/components/slots.md
```diff
@@ -412,6 +412,36 @@ Passing props to a named slot:
...
+    <p>{{ message }}</p>
+    <template #footer>
+      <!-- message belongs to the default slot, and is not available here -->
+      <p>{{ message }}</p>
+    </template>
+  </BaseLayout>
+</template>
+```
+
+Instead, you will need to write it as follows:
```
skirtles-code, line 429:
```suggestion
Using an explicit `<template>` tag for the default slot helps to make it clear that the `message` prop is not available inside the other slot:

 ```

File: src/guide/components/slots.md [hunk 4]
```diff
@@ -412,6 +412,36 @@ Passing props to a named slot:
...
+
+```vue-html
+<template>
+  <BaseLayout v-slot="{ message }">
+    <p>{{ message }}</p>
+    <template #footer>
+      <!-- message belongs to the default slot, and is not available here -->
+      <p>{{ message }}</p>
+    </template>
+  </BaseLayout>
```
skirtles-code, line 425: I think the two examples should use the same component name. As the preceding example also uses `MyComponent` I decided to go with that, but we could make them both `BaseLayout` if you prefer.

```suggestion
  <MyComponent v-slot="{ message }">
    <p>{{ message }}</p>
    <template #footer>
      <!-- message belongs to the default slot, and is not available here -->
      <p>{{ message }}</p>
    </template>
  </MyComponent>
```

skirtles-code (changes requested):
I've made some suggestions below. Let me know if anything isn't clear or if you have any questions. Thanks.

src/guide/components/slots.md
noook: I'm sorry, it was clearly a failed copy paste from my local example, it's normal to have `MyComponent`

noook:
@skirtles-code it's totally clear, I used your suggestions.

skirtles-code (approved):
Looks good to me!

NataliaTepluhina (approved):
@noook thank you for the contribution and thanks to @skirtles-code for the detailed review! PR looks good to me, let's get it merged

PR #1948, opened 2022-09-06
update index.md

RobsonRomeuRizzieri:
add translation portuguese brazil


SkyaTura:
You should start by reading the [starting a new translation](https://github.com/vuejs/docs/blob/main/src/translations/index.md#starting-a-new-translation) before submitting a new PR.

The change you made would only send users to a page that doesn't exist yet.

skirtles-code:
I don't think this can be merged, as the URL does not appear to link to a valid repository. If you were trying to initiate a new translation then I suggest joining the discussions at <https://github.com/vuejs-translations/guidelines/discussions>.

PR #1951, opened 2022-09-08
fix special sponsor layout problem on zh-cn doc.

wxsms:
## Description of Problem

There's no description in platinum_china data.  Which caused the layout push to right by the `lead` span. We should keep the span next to img in order to keep the logo always center.

Also added `align-items: center` to keep texts in middle vertically.

image

ref: https://github.com/vuejs-translations/docs-zh-cn/pull/536


skirtles-code (approved):
Looks good.

PR #1968, opened 2022-09-23
add notes in lifecycle hooks

jay-es:
I added notes to `renderTracked` and `renderTriggered` as same as Composition API.

- [`onRenderTracked()`](https://vuejs.org/api/composition-api-lifecycle.html#onrendertracked)
- [`onRenderTriggered()`](https://vuejs.org/api/composition-api-lifecycle.html#onrendertriggered)


skirtles-code (approved):
Thanks for the PR!

I agree, we should have the same notes for both APIs.

PR #1969, opened 2022-09-23
Add a virtual list library

lixiang810:
Since [vue-virtual-scroller](https://github.com/Akryum/vue-virtual-scroller) don't have types, it's necessary to introduce a new library.

## Description of Problem

[vue-virtual-scroller](https://github.com/Akryum/vue-virtual-scroller) don't have types.

## Proposed Solution

Introduce a new library.

### DIFF
--- a/src/guide/best-practices/performance.md
+++ b/src/guide/best-practices/performance.md
@@ -138,6 +138,7 @@ Implementing list virtualization isn't easy, luckily there are existing communit
 
 - [vue-virtual-scroller](https://github.com/Akryum/vue-virtual-scroller)
 - [vue-virtual-scroll-grid](https://github.com/rocwang/vue-virtual-scroll-grid)
+- [vueuc/VVirtualList](https://github.com/07akioni/vueuc)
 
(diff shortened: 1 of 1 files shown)


bencodezen (approved):

skirtles-code (approved):
I haven't used this library myself, but I would be in favour of adding it.

My reasoning is:

* `vueuc` isn't just a virtual scrolling library, but that is one of its prominent features.
* Its author, 07akioni, is also the creator of Naive UI, which seems to be a popular component library. We already suggest using Naive UI elsewhere in the docs.
* The [npm downloads](https://www.npmjs.com/package/vueuc) for `vueuc` suggest it is being widely used in the community. It compares favourably with the other two libraries mentioned in the docs, [vue-virtual-scroller](https://www.npmjs.com/package/vue-virtual-scroller) (check the v2 downloads) and [vue-virtual-scroll-grid](https://www.npmjs.com/package/vue-virtual-scroll-grid).
* `vueuc` seems to be actively maintained and has been for some time.

PR #1986, opened 2022-10-03
Reword absent Boolean prop customization note

thedamon:
Reworded sentence about customizing the default behaviour of boolean props. 

## Description of Problem
Previous sentence ends up communicating that the built in behaviour is not desirable, which one would assume is not the case or it would not have been built in.

## Proposed Solution
Mention how to customize the behaviour without implying that the default default behaviour is undesirable

### DIFF
--- a/src/guide/components/props.md
+++ b/src/guide/components/props.md
@@ -484,7 +484,7 @@ Additional details:
 
 - An absent optional prop other than `Boolean` will have `undefined` value.
 
-- The `Boolean` absent props will be cast to `false`. You should set a `default` value for it in order to get desired behavior.
+- The `Boolean` absent props will be cast to `false`. (You can change this by setting a `default` for it — ie: `default: undefined` to behave as a non-Boolean prop).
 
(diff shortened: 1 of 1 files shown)


src/guide/components/props.md
```diff
@@ -484,7 +484,7 @@ Additional details:
 
 - An absent optional prop other than `Boolean` will have `undefined` value.
 
-- The `Boolean` absent props will be cast to `false`. You should set a `default` value for it in order to get desired behavior.
+- The `Boolean` absent props will be cast to `false`. (You can change this by setting a `default` for it — ie: `default: undefined` to behave as a non-Boolean prop).
```
NataliaTepluhina, line 487: **suggestion**: a few little punctuation fixes
```suggestion
- The `Boolean` absent props will be cast to `false`. You can change this by setting a `default` for it — i.e.: `default: undefined` to behave as a non-Boolean prop.
```

NataliaTepluhina (commented):
@thedamon thank you for your PR! I've added a tiny fix but overall the change looks good to me 👍🏻

skirtles-code (approved):
I think that entire section needs reworking, but this change is a step in the right direction so I'm happy to merge it.

PR #1988, opened 2022-10-04
Add vue toronto banner

jillztom:
As discussed with @yyx990803, adding the VueConf Toronto banner.


yyx990803:
Can you resolve the conflict? We just removed the VueSchool banner.

jillztom:
@yyx990803 done ✅

PR #1994, opened 2022-10-06
fix(ts-warning):Property does not exist on type 'never'

Saeed-Paris:
## Description of Problem
hi. saw a warning from TS in VueJobs Page.
sometimes TS gets confused when analyzing our code and yell at us with this message.
```
Property 'x' does not exist on type 'never' in TypeScript
```
## Proposed Solution
simply just use square brackets instead of dot notation to access the property
## Additional Information
not so familiar with PR messages so sry about that.
image of that warning
unknown

### DIFF
--- a/.vitepress/theme/components/VueJobs.vue
+++ b/.vitepress/theme/components/VueJobs.vue
@@ -1,13 +1,6 @@
 <script lang="ts">
 // shared data across instances so we load only once
-const base = `https://app.vuejobs.com/feed/vuejs/docs?format=json`
-
-</script>
-
-<script setup lang="ts">
-import { onMounted, computed } from 'vue'
-
-let vuejobs = $ref<HTMLElement>()
+const base = 'https://app.vuejobs.com/feed/vuejs/docs?format=json'
 
 let items = $ref<Jobs[]>([])
 
@@ -15,14 +8,20 @@ type Jobs = {
   organization: Organization;
   title: string;
   link: string;
-  locations: Array<string>;
-  remote: false | "ALLOWED" | "ONLY";
+  locations: string[];
+  remote: false | 'ALLOWED' | 'ONLY';
 };
 
 type Organization = {
   name: string;
-  avatar: string; 
+  avatar: string;
 };
+</script>
+
+<script setup lang="ts">
+import { onMounted, computed } from 'vue'
+
+let vuejobs = $ref<HTMLElement>()
 
 const openings = computed(() =>
   items.sort(() => 0.5 - Math.random()).slice(0, 2)
@@ -44,8 +43,9 @@ onMounted(async () => {
         target="_blank"
       >
         <div class="vj-company-logo">
-          <img :src="job.organization.avatar"
-            :alt="`${job.organization.name}'s logo for hiring purposes on VueJobs.com.`" />
+          <img 
+            :src="job.organization.avatar" 
+            :alt="`Logo for ${job.organization.name}`" />
         </div>
         <div
           style="
@@ -57,7 +57,7 @@ onMounted(async () => {
           <div class="vj-job-title">{{ job.title }}</div>
           <div class="vj-job-info">
             {{ job.organization.name }} <span>· </span>
-            <span>{{job.remote ? "Remote" : job.locations[0]}}</span>
+            <span>{{ job.remote ? 'Remote' : job.locations[0] }}</span>
           </div>
         </div>
       </a>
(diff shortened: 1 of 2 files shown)


skirtles-code (commented):
Thanks for the PR!

I'm wondering whether it might be better to fix this by introducing a type rather than suppressing the warnings?

I was thinking something like this:

```ts
let items = $ref<Job[]>([])
```

Then define a type called `Job` that has all the relevant properties used in the template.

.vitepress/theme/components/VueJobs.vue
```diff
@@ -1,14 +1,29 @@
...
+  organization: Organization;
+  title: string;
+  link: string;
+  locations: Array<string>;
+  remote: false | "ALLOWED" | "ONLY";
+};
+
+type Organization = {
+  name: string;
+  avatar: string;
```
Saeed-Paris, line 24: it can be null so we should choose an avatar to show whenever user didn't upload any image.

.vitepress/theme/components/VueJobs.vue
```diff
@@ -29,35 +44,32 @@ onMounted(async () => {
...
-        >
+          ">
           <div class="vj-job-title">{{ job.title }}</div>
           <div class="vj-job-info">
             {{ job.organization.name }} <span>· </span>
-            <span v-if="['ONLY', 'ALLOWED'].includes(job.remote)"
-              >Remote</span
-            >
-            <span v-else>{{ job.locations[0] }}</span>
+            <span>{{job.remote ? "Remote" : job.locations[0]}}</span>
```
Saeed-Paris, line 60: I checked on Vuejobs.com that user can only choose between remote or just location so it won't be anything except those three that i'd mentioned at line 19.

Saeed-Paris:
@skirtles-code I was completely wrong. Sorry about that.

skirtles-code:
@IsraelOrtuno @jillztom Would either of you be able to confirm that the type changes here are correct? Do the proposed types match what the VueJobs API can return?

IsraelOrtuno:
> @IsraelOrtuno @jillztom Would either of you be able to confirm that the type changes here are correct? Do the proposed types match what the VueJobs API can return?

As of now I do not see any other case where the API could cause a conflict with that type.

Users could skip uploading image logo but it's very rare and in those cases, we find the logo ourselves before publishing.

Thanks for checking this.

.vitepress/theme/components/VueJobs.vue
skirtles-code, line 60:
```suggestion
            <span>{{ job.remote ? 'Remote' : job.locations[0] }}</span>
```

.vitepress/theme/components/VueJobs.vue
```diff
@@ -29,35 +44,32 @@ onMounted(async () => {
         target="_blank"
       >
         <div class="vj-company-logo">
-          <img :src="job.organization.avatar" alt="" />
+          <img :src="job.organization.avatar"
+            :alt="`${job.organization.name}'s logo for hiring purposes on VueJobs.com.`" />
```
skirtles-code, line 48: The attributes need to be consistent, either all of the same line or all on different lines. I think I'd also go for a shorter `alt` text:
```suggestion
          <img
            :src="job.organization.avatar"
            :alt="`Logo for ${job.organization.name}`"
          />
```

.vitepress/theme/components/VueJobs.vue
```diff
@@ -1,14 +1,29 @@
...
 let vuejobs = $ref<HTMLElement>()
 
+let items = $ref<Jobs[]>([])
+
+type Jobs = {
+  organization: Organization;
+  title: string;
+  link: string;
+  locations: Array<string>;
+  remote: false | "ALLOWED" | "ONLY";
```
skirtles-code, line 19: I suggest using `string[]` instead of `Array<string>`, and single quotes for strings. I think that'd be more consistent with the usual code style used in the docs.

```suggestion
  locations: string[];
  remote: false | 'ALLOWED' | 'ONLY';
```

.vitepress/theme/components/VueJobs.vue
```diff
@@ -1,14 +1,29 @@
...
-let items = $ref([])
+
 </script>
 
 <script setup lang="ts">
 import { onMounted, computed } from 'vue'
 
 let vuejobs = $ref<HTMLElement>()
 
+let items = $ref<Jobs[]>([])
```
skirtles-code, line 12: As noted in the comment at the start of the `<script>` section, `items` needs to be in that section to ensure the data only gets loaded once. We lose that if we move it into the `<script setup>`.

skirtles-code (changes requested):
Thanks for your work on this. Here are my suggestions...

Saeed-Paris:
@skirtles-code  thanks for great tips. I really appreciate it.

skirtles-code (approved):

PR #1995, opened 2022-10-07
chore: consistent headings for API docs

jay-es:
## Proposed Solution
With colon if there is text on the same line.
Otherwise, no colon.

## Additional Information
Or, should I add a colon to all headings for more clarity?


yyx990803:
Sorry for letting this sit for so long! There were some conflict due to heavy content changes, so I landed this in 202b311

PR #1999, opened 2022-10-12
fix: add vs dot unified writing paradigm

zhousg:
## Description of Problem

add vs dot unified writing paradigm


skirtles-code (approved):
Thanks for the PR!

The use of `vs.` seems to be standard in US English, whereas `vs` would be considered correct in British English. There are various instances of `vs.` elsewhere in the docs, so these two instances of `vs` do appear to be inconsistent.

PR #2003, opened 2022-10-12
Fix typo composition api lifecycle

devisasari:


skirtles-code:
Already included in #2010. Closing.

PR #2005, opened 2022-10-12
Fix typo guide components registration

devisasari:

PR #2001, opened 2022-10-12
Fix typo rendering mechanism

devisasari:

PR #2006, opened 2022-10-12
Fix typo composition api dependency injection

devisasari:


brc-dd:
It would be better if you can just merge #2000, #2001, #2002, #2003, #2004, #2005, #2006 into a single PR.

devisasari:
> Already included in #2010. Closing.

Thank you

PR #2010, opened 2022-10-12
Fix typo: change descendent to descendant

devisasari:


skirtles-code:
@devisasari I have closed the other PRs as this one seems to include all of the commits. I'm not sure why you opened so many PRs with the same changes, but it was only necessary to open this one PR.

skirtles-code (approved):
Both *descendant* and *descendent* are valid English words. The two spellings are used interchangeably in common usage, though *descendant* is more common.

More formally, *descendant* would be considered the correct spelling in this context. In the stricter, formal definition, the word *descendent* would be considered an adjective referring to something moving downwards.

PR #2015, opened 2022-10-14
Fix: remove extraneous reference to v-bind

ig3:
## Description of Problem
This leads the reader to search for the use of `v-bind` but it isn't used in the examples or anywhere in the page so the search is a waste of time and the note is unhelpful.

## Proposed Solution
Remove the note about `v-bind`.

### DIFF
--- a/src/guide/essentials/component-basics.md
+++ b/src/guide/essentials/component-basics.md
@@ -297,8 +297,6 @@ Then want to render a component for each one, using `v-for`:
 
 </div>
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

`v-bind` is the name of the directive that is commonly abbreviated to the `:` prefix. This is explained earlier in the guide and is assumed knowledge by this point. It is used in the example that immediately precedes this paragraph.

ig3:
Thank you for the quick response and explanation.

PR #2021, opened 2022-10-15
Fix typo reactivity fundamentals

devisasari:


src/guide/essentials/template-syntax.md
```diff
@@ -255,7 +255,7 @@ The above will be converted to `:[someattr]` in in-DOM templates. If your compon
 
 ### Modifiers
 
-Modifiers are special postfixes denoted by a dot, which indicate that a directive should be bound in some special way. For example, the `.prevent` modifier tells the `v-on` directive to call `event.preventDefault()` on the triggered event:
+Modifiers are special postfixes denoted by a dot, which indicates that a directive should be bound in some special way. For example, the `.prevent` modifier tells the `v-on` directive to call `event.preventDefault()` on the triggered event:
```
skirtles-code, line 258: I believe **indicate** is correct here, as the relevant noun is **modifiers**, not **dot**.

src/guide/introduction.md
```diff
@@ -89,7 +89,7 @@ If you find these concepts intimidating, don't worry! The tutorial and guide onl
...
 
-Despite the flexibility, the core knowledge about how Vue works is shared across all these use cases. Even if you are just a beginner now, the knowledge gained along the way will stay useful as you grow to tackle more ambitious goals in the future. If you are a veteran, you can pick the optimal way to leverage Vue based on the problems you are trying to solve, while retaining the same productivity. This is why we call Vue "The Progressive Framework": it's a framework that can grow with you and adapt to your needs.
+Despite the flexibility, the core knowledge about how Vue works is shared across all these use cases. Even if you are just a beginner now, the knowledge gained along the way will stay useful as you grow to tackle more ambitious goals in the future. If you are a veteran, you can pick the optimal way to leverage Vue based on the problems you are trying to solve while retaining the same productivity. This is why we call Vue "The Progressive Framework": it's a framework that can grow with you and adapt to your needs.
```
skirtles-code, line 92: I find the original version easier to read. I think the comma helps to avoid ambiguous parsing of the sentence.

For example, consider the following:

> ... the problems you are trying to solve while coding.

The use of 'solve while' here is very different to the sense it is used in the docs. Without the comma, I think the sentence in the docs can be mentally parsed this way and it only becomes apparent later in the sentence that this isn't the intended meaning (what's known as a [garden-path sentence](https://en.wikipedia.org/wiki/Garden-path_sentence)). The comma avoids that ambiguity.

skirtles-code (changes requested):
Thanks for the fixes!

I've reviewed the proposed changes and made a couple of suggestions below.

On a separate note, you've opened a lot of PRs over the last few days, many of which contain the same commits. Some of the changes in this PR have already been merged, which just makes it extra work to review. Please can you ensure that you're adding commits to a suitable branch rather than continuing the branch from your previous PR? Each PR should contain separate changes, branching from `main`. How many commits/changes should go into each PR is often subjective, but there shouldn't be overlap between PRs. GitHub gives you a preview of the changes when you create a PR, so I suggest checking that to ensure that it only contains the changes you're expecting. Thanks.

devisasari:
> Thanks for the fixes!
> 
> I've reviewed the proposed changes and made a couple of suggestions below.
> 
> On a separate note, you've opened a lot of PRs over the last few days, many of which contain the same commits. Some of the changes in this PR have already been merged, which just makes it extra work to review. Please can you ensure that you're adding commits to a suitable branch rather than continuing the branch from your previous PR? Each PR should contain separate changes, branching from `main`. How many commits/changes should go into each PR is often subjective, but there shouldn't be overlap between PRs. GitHub gives you a preview of the changes when you create a PR, so I suggest checking that to ensure that it only contains the changes you're expecting. Thanks.

Thank you for your suggestions and contrubitions, I consider what you say.

src/guide/essentials/template-syntax.md
devisasari: Thanks you're right

PR #2024, opened 2022-10-15
Update utility-types.md

wxsms:
looks like it should not be included in the code tag?

ref: vuejs-translations/docs-zh-cn/pull/581


wxsms:
close as duplicate #2023

Issue #2041, opened 2022-10-21
Should add a link to production deployment doc


aklipfel:
# The Problem
While translating `production-deployment.md` file in French, I came across "Reactivity debugging" that was tricky to translate because this concept was not clear to me and had to look for it on the internet to find the corresponding article.

## Proposed Solution
I would suggest to add a link to the Reactivity Debugging doc, for users that didn't know about this concept like me. 

Linked to #2042

skirtles-code:
Closed by #2042.

PR #2042, opened 2022-10-21
fix: add reactivity debugging link

aklipfel:
# The Problem
While translating `production-deployment.md` file in French, I came across "Reactivity debugging" that was tricky to translate because this concept was not clear to me and had to look for it on the internet to find the corresponding article.

## Proposed Solution
I would suggest to add a link to the Reactivity Debugging doc, for users that didn't know about this concept like me. 

Linked to #2041


skirtles-code (approved):

skirtles-code:
Good idea, thanks for the PR!

PR #2043, opened 2022-10-22
fix: TeamList sticky info overlap

sadeghbarati:
## Description of Problem
Vue docs **`banner`** caused overlapping TeamList sticky info title ( e.g. Core Team Members ) under header ( VPNav )

https://vuejs.org/about/team.html

Screenshot 2022-10-22 120434

## Proposed Solution

- `html.banner-dismissed`
- Use predefined `--vt-banner-height` CSS variable and CSS `calc()` for calculate correct **`top`** value for `sticky` position

### DIFF
--- a/src/about/team/TeamList.vue
+++ b/src/about/team/TeamList.vue
@@ -61,18 +61,26 @@ defineProps<{
 @media (min-width: 768px) {
   .info {
     position: sticky;
-    top: 32px;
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

I need to give this a closer look, but my first impression is that this looks good.

skirtles-code (approved):

PR #2046, opened 2022-10-23

skuhnow:
Fixed the nesting of "Ref Unwrapping in Arrays and Collections", it was depth 4, should be depth 3

## Description of Problem

"Ref Unwrapping in Arrays and Collections" is h4 (**####**), so it's not in the navigation. Should be h3 (**###**)

## Proposed Solution

Removed one "#"


skirtles-code (approved):
Thanks for the PR!

I'm not entirely sure whether the intention was for that section be level 3 or level 4, as I can see an argument for it being a subsection under **Ref Unwrapping in Reactive Objects**. That said, I think I prefer having it as level 3.

I'm not going to merge this immediately, so that others can weigh in, but I agree with the proposed change.

PR #2049, opened 2022-10-26
feat: add en anchors

### LINKED ISSUE #1998: Add english anchors
As stated in https://github.com/vuejs-translations/guidelines#retaining-original-anchors, translated versions of the repo need to add the original english anchors.
But this takes time to add for each translation. I think it would be great to have those anchors in the source repo so they are already ready to go when we are translating.

What do you think?

wxsms:
## Description of Problem

close #1998 

added a script to generate anchors. It can also re-run for updates.


wxsms:
Looks like that `reactivity-fundamentals.md` has 2 headings with same name, which will generate same slug id and cause vitepress complain. Any idea?

edimitchel:
> Looks like that `reactivity-fundamentals.md` has 2 headings with same name, which will generate same slug id and cause vitepress complain. Any idea?

Seems to be a duplicate title for composition api / option api..

Jinjiang:
I suggest we just follow the current rule to add a incremental number after the duplicated ids. That's also what you can get on the current doc site.

image

For example, the second `#reactive-proxy-vs-original` should become `#reactive-proxy-vs-original-1` automatically.

Thanks.

skirtles-code:
I'm wondering whether we need to include the script in the PR? The script is useful for the initial bulk adding of headings, but do we really need to use it beyond that?

Perhaps we could just manually correct the handful of duplicate headers rather than trying to get the script to do it automatically?

edimitchel:
> I'm wondering whether we need to include the script in the PR? The script is useful for the initial bulk adding of headings, but do we really need to use it beyond that?
> 
> Perhaps we could just manually correct the handful of duplicate headers rather than trying to get the script to do it automatically?

It could be put into a gist instead of directly in the source of docs yes!

Jinjiang:
> > I'm wondering whether we need to include the script in the PR? The script is useful for the initial bulk adding of headings, but do we really need to use it beyond that?
> > Perhaps we could just manually correct the handful of duplicate headers rather than trying to get the script to do it automatically?
> 
> It could be put into a gist instead of directly in the source of docs yes!

Sounds a good idea.

wxsms:
I just removed the script related things from my branch. And fixed the duplicated anchor as @Jinjiang suggested.

the script posted here, and you may use it at will.

<details>

```js
/**
 * slugify function taken from vitepress
 */
// string.js slugify drops non ascii chars so we have to
// use a custom implementation here
const { remove: removeDiacritics } = require('diacritics')
// eslint-disable-next-line no-control-regex
const rControl = /[\u0000-\u001f]/g
const rSpecial = /[\s~`!@#$%^&*()\-_+=[\]{}|\\;:"'<>,.?/]+/g

const slugify = (str) => {
  return (
    removeDiacritics(str)
      // Remove control characters
      .replace(rControl, '')
      // Replace special characters
      .replace(rSpecial, '-')
      // Remove continuous separators
      .replace(/\-{2,}/g, '-')
      // Remove prefixing and trailing separators
      .replace(/^\-+|\-+$/g, '')
      // ensure it doesn't start with a number (#121)
      .replace(/^(\d)/, '_$1')
      // lowercase
      .toLowerCase()
  )
}

let glob = require('glob');
let path = require('path');
let fs = require('fs');

glob('../../src/**/*.md', { cwd: __dirname }, function (err, files) {
  for (let f of files) {
    let filePath = path.join(__dirname, f);
    let content = fs.readFileSync(filePath, { encoding: 'utf-8' });
    let headings = content.match(/^#+ .+$/gm);
    for (let h of headings ?? []) {
      let _h = h.replace(/{#.+$/, '').trim();
      content = content.replace(h, `${_h} {#${slugify(_h)}}`);
    }
    fs.writeFileSync(filePath, content, { encoding: 'utf-8' });
  }
});
```

</details>

src/api/options-lifecycle.md
```diff
@@ -209,7 +209,7 @@ Called when an error propagating from a descendant component has been captured.
 
   - An `errorCaptured` hook can return `false` to prevent the error from propagating further. This is essentially saying "this error has been handled and should be ignored." It will prevent any additional `errorCaptured` hooks or `app.config.errorHandler` from being invoked for this error.
 
-## renderTracked <sup class="vt-badge dev-only" />
+## renderTracked <sup class="vt-badge dev-only" /> {#rendertracked-sup-class-vt-badge-dev-only}
```
Jinjiang, line 212: The original hash names like this one on the website (`#rendertracked`) actually doesn't include the `-sup-*` things.

https://vuejs.org/api/options-lifecycle.html#rendertracked

I think we'd better still follow this rule to avoid breaking changes.

Thanks.

wxsms: fixed

edimitchel (approved):

Jinjiang (approved):

src/api/composition-api-dependency-injection.md
```diff
@@ -1,6 +1,6 @@
-# Composition API: <br>Dependency Injection
+# Composition API: <br>Dependency Injection {#composition-api-br-dependency-injection}
```
skirtles-code, line 1: I believe this has changed relative to the live docs, due to the `<br>`.

```suggestion
# Composition API: <br>Dependency Injection {#composition-api-dependency-injection}
```

skirtles-code (approved):

hymair:
Seems like this introduced a regression:

image

cc @wxsms @Jinjiang @skirtles-code @edimitchel

Jinjiang:
@hymair thanks for figuring it out. I created a PR to fix this. 🙏

PR #2050, opened 2022-10-26
Change emit example function name for clarity

bbaustin:
## Description of Problem
As-is, I did not immediately see the connection between `this.$emit('submit')` and the following code, `<MyComponent @some-event="callback" />` in the "component instance" example

## Proposed Solution
I suggest changing the function name from `submit` to `someEvent` (and also possibly adding some extra HTML example code for added clarity)

## Additional Information
I'm just learning Vue and was a little confused by this section in the docs. It's very possible I'm still confused, in which case, please disregard :)

### DIFF
--- a/src/guide/components/events.md
+++ b/src/guide/components/events.md
@@ -19,11 +19,6 @@ A component can emit custom events directly in template expressions (e.g. in a `
 
 The `$emit()` method is also available on the component instance as `this.$emit()`:
 
-```vue-html
-<!-- MyComponent -->
-<button @click="submit">click me</button>
-```
-  
 ```js
 export default {
   methods: {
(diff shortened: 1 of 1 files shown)


skirtles-code (changes requested):
Thanks for the PR!

I agree that the event name should be changed to match the surrounding examples. The sudden switch to `submit` just makes it confusing.

I don't think we need the extra template snippet though. How the `submit()` method gets called isn't really relevant to the point being made.

bbaustin:
> I don't think we need the extra template snippet though. How the `submit()` method gets called isn't really relevant to the point being made.

Sure, that makes sense. I made another commit to remove the snippet!

skirtles-code (approved):

PR #2053, opened 2022-10-28
Update web-components.md

MuhammadZainAkram:


skirtles-code:
Thanks for the suggestion.

You didn't provide a description on the PR, so I'm a little unsure about what the objective of this change is.

I think we should avoid making claims about 'forever'. I also feel that the addition changes the emphasis of the sentence, which is currently trying to make a point about only using features that work well today. Mentioning the future muddies that message.

I also feel that the extended sentence doesn't really scan as well, due to the repeated use of 'and'.

yyx990803:
Agreed with @skirtles-code - the forever promise seems unnecessary here.

PR #2054, opened 2022-10-29
Add computed refs to isReadonly

skirtles-code:
The `isReadonly()` function will return `true` for a `computed()` ref that doesn't have a `set` function. That isn't currently mentioned.

I've added a brief description of what it means for an object to be 'readonly', otherwise the list that follows may seem a bit arbitrary. I've then included `computed()` refs in the list of things that are readonly.


NataliaTepluhina (approved):

skirtles-code:
This was merged into `main` a week ago, but GitHub is still showing it as unmerged. I'm going to merge it again to ensure that the change is detected by the translation workflow.

PR #2055, opened 2022-11-02
Update slots.md

cedricai:
## Description of Problem
When I learn to Scoped Slots of the slots charater in the playground, I find  it doesn't work as expected to `<MyComponent v-slot="slotProps">`. But when I try using `<MyComponent v-slot="$slotProps">` or `<MyComponent slot="slotProps">`, it works as expected. So I think this is a bug and it needs to be fixed.
## Proposed Solution
Change `<MyComponent v-slot="slotProps">` to `<MyComponent v-slot="$slotProps">`.
## Additional Information
related url: https://vuejs.org/guide/components/slots.html#scoped-slots
related code: 
```vue
<script>
import MyComponent from './MyComponent.vue'
  
export default {
  components: {
    MyComponent
  },
  data() {
    return {
      slotProps: {
        text: 'xxx',
        count: 10
      }
    }
  }
}
</script>

<template>
	<MyComponent v-slot="$slotProps">
  	{{ slotProps.text }} {{ slotProps.count }}
  </MyComponent>
</template>
```

### DIFF
--- a/src/guide/components/slots.md
+++ b/src/guide/components/slots.md
@@ -333,7 +333,7 @@ In fact, we can do exactly that - we can pass attributes to a slot outlet just l
 Receiving the slot props is a bit different when using a single default slot vs. using named slots. We are going to show how to receive props using a single default slot first, by using `v-slot` directly on the child component tag:
 
 ```vue-html
-<MyComponent v-slot="slotProps">
+<MyComponent v-slot="$slotProps">
   {{ slotProps.text }} {{ slotProps.count }}
 </MyComponent>
 ```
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

The current examples are correct. The name of the object received via the slot can be whatever you want, but the name needs to match the name used by the nested elements.

The example you gave above is using `slotProps` as a `data` property. That should be removed. `slotProps` need to come from `<MyComponent>` instead: [Example](https://sfc.vuejs.org/#eNpdUctOwzAQ/JXVXnohNlyjgIS4cEHiA/YSwgZSxQ/Zm1IU5d9ZtzRKK0WKZ2zP7IxnfI7RHCbGGpvcpSHKE/nBxZAE3n5fgq48e4E+BQc7YzdcubYjD0Cej6cLn9y30ygwF7a7nMv1mYCtYCEW8vo1dvVVIOzi2AoXJM12gkOVxyCPhOX3nkLMhHpK7WWeYSWN8FFgWeCK7MKkEovaATTbEKrQ2NUU7/CcvXJtNPscvPZyGp7+N9R0jUOoDRRM+C0Sc21t7rtSyz6bkL6srkxS48Gx4eyqjxR+MicVJrwUoJY3nZan2NSgA5ccUHJp+lcex0Co9aqw4od7BfYmxvIHoDqm7Q==).

PR #2057, opened 2022-11-03
options api - immediate watcher handler runs before "created"

ArtemTropanets:
## Description of Problem
Although it's mentioned that in Options API immediate watcher handler will run on component creation, it might be not instantly obvious that "created" hook will be executed after. Introduced few bugs before I realized it.

## Proposed Solution
Maybe could be useful to explicitly mention in the guide that immediate watcher handler runs before "created" hook.

### DIFF
--- a/src/guide/essentials/watchers.md
+++ b/src/guide/essentials/watchers.md
@@ -244,7 +244,7 @@ export default {
 }
 ```
 
-In above scenario the initial execution of handler will happen after `beforeCreate` hook, but before `created`.
+The initial execution of the handler function will happen just before the `created` hook. Vue will have already processed the `data`, `computed`, and `methods` options, so those properties will be available on the first invocation.
 </div>
 
 <div class="composition-api">
--- a/src/guide/essentials/watchers.md
+++ b/src/guide/essentials/watchers.md
@@ -244,6 +244,7 @@ export default {
 }
 ```
 
+In above scenario the initial execution of handler will happen after `beforeCreate` hook, but before `created`.
 </div>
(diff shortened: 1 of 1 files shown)


src/guide/essentials/watchers.md
```diff
@@ -244,6 +244,7 @@ export default {
 }
 ```
 
+In above scenario the initial execution of handler will happen after `beforeCreate` hook, but before `created`.
```
skirtles-code, line 247: Keeping in mind the outline presented in the [lifecycle diagram](https://vuejs.org/guide/essentials/lifecycle.html#lifecycle-diagram), I think it might be worth mentioning that the watcher is called after the various properties have been initialised, rather than mentioning `beforeCreate` directly. Maybe something like...

```suggestion
The initial execution of the handler function will happen just before the `created` hook. Vue will have already processed the `data`, `computed`, and `methods` options, so those properties will be available on the first invocation.
```

skirtles-code (commented):
Thanks for the PR, good idea.

I've made a suggestion about the exact wording below.

src/guide/essentials/watchers.md
ArtemTropanets: Thanks, I like the suggested wording.

skirtles-code (approved):
I'll give others a few days to comment on the proposed change, but I'm happy with it.

NataliaTepluhina:
I'm fine with the change too 👍🏻

NataliaTepluhina (approved):

PR #2037, opened 2022-10-18
Fix typos and grammar

devisasari:

PR #2060, opened 2022-11-04
deleting unnecessary comma and improve tone

leythadan:
## Description of Problem
There's an unnecessary comma and a minor fix with the tone of one sentence.
## Proposed Solution
remove the comma and use a positive tone


skirtles-code:
Thanks for the PR!

We have an existing PR that tries to remove that same comma. See <https://github.com/vuejs/docs/pull/2037/files#r997710710>. I personally think it makes the sentence more difficult to read without it, as explained in the linked comments.

The changes you've proposed to that second paragraph seem to significantly change the meaning. It doesn't really convey the intended sentiment in its new form.

skirtles-code:
For the reasons I mentioned above I'm going to close this. Thanks again for the suggestions.

PR #2061, opened 2022-11-04
missing pronoun

leythadan:
## Description of Problem
There appears to be a missing pronoun, `it.`
## Proposed Solution
Add the missing pronoun.

### DIFF
--- a/src/guide/built-ins/keep-alive.md
+++ b/src/guide/built-ins/keep-alive.md
@@ -14,7 +14,7 @@ In the Component Basics chapter, we introduced the syntax for [Dynamic Component
 <component :is="activeComponent" />
 ```
 
-By default, an active component instance will be unmounted when switched away from. This will cause any changed state it holds to be lost. When this component is displayed again, a new instance will be created with only the initial state.
+By default, an active component instance will be unmounted when switched away from it. This will cause any changed state it holds to be lost. When this component is displayed again, a new instance will be created with only the initial state.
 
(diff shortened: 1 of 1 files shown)


src/guide/built-ins/keep-alive.md
```diff
@@ -14,7 +14,7 @@ In the Component Basics chapter, we introduced the syntax for [Dynamic Component
 <component :is="activeComponent" />
 ```
 
-By default, an active component instance will be unmounted when switched away from. This will cause any changed state it holds to be lost. When this component is displayed again, a new instance will be created with only the initial state.
+By default, an active component instance will be unmounted when switched away from it. This will cause any changed state it holds to be lost. When this component is displayed again, a new instance will be created with only the initial state.
```
skirtles-code, line 17: I think we need to change the verb slightly too:

```suggestion
By default, an active component instance will be unmounted when switching away from it. This will cause any changed state it holds to be lost. When this component is displayed again, a new instance will be created with only the initial state.
```

skirtles-code (changes requested):
I agree, ending the sentence with a preposition is not ideal.

I don't think we can just put 'it' on the end, but with a small tweak earlier in the sentence it should work. See my suggestion below.

src/guide/built-ins/keep-alive.md
leythadan: @skirtles-code good point.

skirtles-code: If you'd like to accept my proposed change you can merge it into your branch by clicking the 'Commit suggestion' button beneath the change.

skirtles-code (approved):

PR #2063, opened 2022-11-05
minor grammatical errors

leythadan:
## Description of Problem
Some paragraphs have minor grammar errors
## Proposed Solution
improve grammar

### DIFF
--- a/src/guide/extras/web-components.md
+++ b/src/guide/extras/web-components.md
@@ -64,7 +64,7 @@ module.exports = {
 
 ### Passing DOM Properties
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR, but I don't believe any of these are errors. The grammar of those 3 sentences appears to be correct to me.

leythadan:
You've got to be kidding me. Do you think the following sentence is correct? 

`If the custom elements will be used in an application that is also using Vue, you can choose to externalize Vue from the built bundle so that the elements will be using the same copy of Vue from the host application.`

skirtles-code:
I think the sentence is maybe a bit lengthy. I also think it would be better if it ended '*as* the host application', rather than 'from'. But neither of those problems were addressed by your change.

The part you changed appears to be fine to me. From a grammatical perspective I think either would wording would work, but I prefer the use of 'will be' in this context. We're talking about custom elements like they're a library, so I think it makes more sense to refer to how they're going to be used (future) rather than how they are being used (present).

PR #2065, opened 2022-11-07
fix some anchors

wxsms:
## Description of Problem

fix a problem introduce by #2049

### DIFF
--- a/src/guide/best-practices/accessibility.md
+++ b/src/guide/best-practices/accessibility.md
@@ -478,7 +478,7 @@ The World Wide Web Consortium (W3C) Web Accessibility Initiative (WAI) develops 
 - [Robust](https://www.w3.org/TR/WCAG21/#robust)
   - Users must be able to access the content as technologies advance
 
-#### Web Accessibility Initiative – Accessible Rich Internet Applications (WAI-ARIA) {#web-accessibility-initiative-–-accessible-rich-internet-applications-wai-aria}
+#### Web Accessibility Initiative – Accessible Rich Internet Applications (WAI-ARIA) {#web-accessibility-initiative-accessible-rich-internet-applications-wai-aria}
 
 W3C's WAI-ARIA provides guidance on how to build dynamic content and advanced user interface controls.
 
--- a/src/guide/scaling-up/testing.md
+++ b/src/guide/scaling-up/testing.md
@@ -97,13 +97,13 @@ A component can be tested in two ways:
 
    Tests that are "Blackbox tests" are unaware of the implementation details of a component. These tests mock as little as possible to test the integration of your component and the entire system. They usually render all child components and are considered more of an "integration test". See the [Component Testing recommendations](#component-testing) below.
 
-### Recommendation {#recommendation} {#recommendation} {#recommendation}
+### Recommendation {#recommendation}
 
 - [Vitest](https://vitest.dev/)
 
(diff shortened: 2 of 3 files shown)


src/guide/best-practices/accessibility.md
```diff
@@ -478,7 +478,7 @@ The World Wide Web Consortium (W3C) Web Accessibility Initiative (WAI) develops
 - [Robust](https://www.w3.org/TR/WCAG21/#robust)
   - Users must be able to access the content as technologies advance
 
-#### Web Accessibility Initiative – Accessible Rich Internet Applications (WAI-ARIA) {#web-accessibility-initiative-–-accessible-rich-internet-applications-wai-aria}
+#### Web Accessibility Initiative – Accessible Rich Internet Applications (WAI-ARIA) {#web-accessibility-initiative-accessible-rich-internet-applications-wai-aria}
```
skirtles-code, line 481: I believe this would be a change to the original anchor that was generated by VitePress. What's the motivation behind changing it?

wxsms: It looks strange (3 dashes), I'm not sure if it's the original anchor that was generated by VitePress or a mistake of my script from #2049. If you think this is okay, feel free to close this one.

wxsms: Btw I found this when sycn these anchor to zh-cn doc. And I found that zh-cn doc is using the one-dash version anchor. Thats why I think it might be a mistake of my script and created this PR.

edimitchel (changes requested):
It seems that forms.md have the same issue:
search for {#checkbox} and {#radio}

skirtles-code:
Those other duplicates have already been fixed in #2066.

src/guide/best-practices/accessibility.md
skirtles-code: The header with 3 dashes is the same as we had previously. You can see it in the preview deployments for some of the older PRs. I've also checked it locally with an older branch.

Perhaps this dash should be stripped out like other punctuation, but in that case I think it should be changed in VitePress too.

wxsms: okay, feel free to close this if no need to change. thanks.

src/guide/scaling-up/testing.md
```diff
@@ -97,13 +97,13 @@ A component can be tested in two ways:
...
+### Recommendation {#recommendation}
 
 - [Vitest](https://vitest.dev/)
 
   Since the official setup created by `create-vue` is based on [Vite](https://vitejs.dev/), we recommend using a unit testing framework that can leverage the same configuration and transform pipeline directly from Vite. [Vitest](https://vitest.dev/) is a unit testing framework designed specifically for this purpose, created and maintained by Vue / Vite team members. It integrates with Vite-based projects with minimal effort, and is blazing fast.
 
-### Other Options {#other-options} {#other-options} {#other-options}
+### Other Options {#other-options}
 
 - [Peeky](https://peeky.dev/) is another fast unit test runner with first-class Vite integration. It is also created by a Vue core team member and offers a GUI-based testing interface.
```
Jinjiang, line 109: It would be great if we can add custom anchors for line 208, 226, 266, and 272 by this chance, too. Thanks.

wxsms: it was resolved in #2067 already.

skirtles-code:
I believe everything is fine here, relative to what we had previously, so I'm going to close this.

PR #2069, opened 2022-11-08
Update: watchers.md

JohnnyEvo:
Add `Eager Watchers` section to composition side

## Description of Problem

When we are in `composition` option, `Eager watchers` is not mentioned.

## Proposed Solution

Add `Eager watchers` in `composition` option.

### DIFF
--- a/src/guide/essentials/watchers.md
+++ b/src/guide/essentials/watchers.md
@@ -264,7 +264,9 @@ watch(obj, (newValue, oldValue) => {
 
 ## `watchEffect()` \*\* {#watcheffect}
 
-`watch()` is lazy: the callback won't be called until the watched source has changed. But in some cases we may want the same callback logic to be run eagerly - for example, we may want to fetch some initial data, and then re-fetch the data whenever relevant state changes. We may find ourselves doing this:
+`watch()` is lazy: the callback won't be called until the watched source has changed. But in some cases we may want the same callback logic to be run eagerly, an alternative to [`watch()`](/api/reactivity-core.html#watch) exists.
+
+We may find ourselves doing this:
 
 ```js
 const url = ref('https://...')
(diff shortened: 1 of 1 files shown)


src/guide/essentials/watchers.md
```diff
@@ -245,6 +245,21 @@ export default {
...
+
+<div class="composition-api">
+  
+`watch` is lazy by default: the callback won't be called until the watched source has changed. But in some cases we may want the same callback logic to be run eagerly - for example, we may want to fetch some initial data, and then re-fetch the data whenever relevant state changes.
+
+We can force a watcher's callback to be executed immediately by declaring it using an object with a `handler` function and the `immediate: true` option:
+
+```js
+watch(obj, (newValue, oldValue) => {
+  // logic
```
skirtles-code, line 259: Maybe use `...`, like the other examples?

```suggestion
  // ...
```

skirtles-code (changes requested):
Good point, this does need to be covered.

I've provided some feedback on your proposed changes below.

JohnnyEvo:
Thank you, I considered your review

skirtles-code:
@JohnnyEvo Thanks for making those changes.

I've just had another read through and I think we now have a problem with the section about `watchEffect()`. The wording of that section no longer makes sense now that it's preceded by this new section about `immediate: true`.

Would you like to attempt that rewording yourself? If not then I can give it a go.

skirtles-code:
The problem is that the `watchEffect()` section presents `watchEffect()` as a way to implement eager watchers. It starts with a sentence saying that `watch()` is lazy and then shows how `watchEffect()` is the solution to that problem. It doesn't make sense if it comes immediately after a section all about eager watchers. We're presenting two solutions to the same problem without acknowledging the overlap.

With hindsight I don't think the omission of the section about eager watcher is an oversight. I think it was intentional, with the section about `watchEffect()` being seen as a way to solve the same problem. I still think having a section about eager watchers is a good idea, but we can't have the `watchEffect()` section framed as a solution to a problem we've just solved in the preceding section.

JohnnyEvo:
I understand the inconsistency. 
We can say that there is an alternative for the lazy watch... but what is the point of having two different ways of doing the same thing? Why choose one or the other?

skirtles-code:
I've pulled these commits into #2144 and made some changes to the `watchEffect()` section, so that it doesn't seem quite so out of place.

PR #2075, opened 2022-11-10
Composition API version

pbkompasz:
## Description of Problem

Some code examples were missing the composition API/options API variant, toggle button was not switching between the two variants.

## Proposed Solution

Added missing Composition API examples.

### DIFF
--- a/src/guide/scaling-up/sfc.md
+++ b/src/guide/scaling-up/sfc.md
@@ -4,6 +4,8 @@
 
 Vue Single-File Components (a.k.a. `*.vue` files, abbreviated as **SFC**) is a special file format that allows us to encapsulate the template, logic, **and** styling of a Vue component in a single file. Here's an example SFC:
 
+<div class="options-api">
+
 ```vue
 <script>
 export default {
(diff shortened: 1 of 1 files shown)


src/guide/scaling-up/sfc.md
```diff
@@ -27,6 +29,31 @@ export default {
...
 
+</div>
+
+<div class="composition-api">
+
+```vue
+<script setup>
+  import { ref } from 'vue';
+  const greeting = ref('Hello World!');
+}
```
skirtles-code, line 40: I suggest removing the indent, the semicolons, and the trailing `}`:
```suggestion
import { ref } from 'vue'
const greeting = ref('Hello World!')
```

src/guide/scaling-up/sfc.md
```diff
@@ -54,16 +81,36 @@ That said, we do realize there are scenarios where SFCs can feel like overkill.
 
 Vue SFC is a framework-specific file format and must be pre-compiled by [@vue/compiler-sfc](https://github.com/vuejs/core/tree/main/packages/compiler-sfc) into standard JavaScript and CSS. A compiled SFC is a standard JavaScript (ES) module - which means with proper build setup you can import an SFC like a module:
 
-```js
-import MyComponent from './MyComponent.vue'
+<div class="options-api">
```
skirtles-code, line 84: I don't think we should change this second example. The current example is intended to show importing a `.vue` file into a plain `.js` file. It's supposed to illustrate how `.vue` files are just ES modules. Showing a `.vue` importing a `.vue` doesn't quite make the same point about compatibility with normal JavaScript.

The existing example is valid for both the Options API and the Composition API. The `components` option is common to both.

skirtles-code (changes requested):
Thanks for the PR!

I've provided some feedback below. In my opinion, introducing a Composition API version of the first example should be fine, but I don't think we can change the second example. More details below.

skirtles-code:
This PR is superseded by #2145. The commits from this PR are included in that PR.

Issue #2030, opened 2022-10-17
Where is Style Guide in menu?


ozum:
Hi,

I couldn’t find the link to “Style Guide” in the menu. I try to use search “Style Guide” using the search box, but no result is returned. As a Vue 2 user, I know it exists at https://vuejs.org/style-guide/ 

It seems like a mysterious page or I missed something obvious :)

Great docs by the way, thanks who wrote.

Regards,

awxiaoxian2020:
Any plan about this?

awxiaoxian2020:
Any update for the guide? It has passed almost three years when the Vue 3 released. I think we should take action for the guide.

awxiaoxian2020:
👀

ethanherbertson:
Went searching for an updated style guide, and was surprised that it is in the state it's in. FWIW this issue gets a +1 from me.

AloisSeckar:
Just watched [this older video](https://www.youtube.com/watch?v=RFlQ8HP8Tr4&t=473s) where the author compares React and Vue and praises Style Guide as one of the great things...

How does it look with Style Guide v3 update and can we help in some way to make it back online sooner?

PR #2076, opened 2022-11-10
Style Guide readded #2030

pbkompasz:
## Description of Problem

Fix for #2030.
[Style Guide](https://vuejs.org/style-guide/) could not be reached.
Search still does not return any results for anything related to this guide.


skirtles-code:
Thanks for the PR, but the Style Guide is intentionally hidden and requires a significant amount of work before it can be re-enabled.

PR #2077, opened 2022-11-11
Fix API index heading text

skirtles-code:
<https://vuejs.org/api/> is currently showing the wrong text for API entries. It should be showing the heading text but instead it's showing the anchor string. e.g. `app-mount` should say `app.mount()`.

### DIFF
--- a/src/api/ApiIndex.vue
+++ b/src/api/ApiIndex.vue
@@ -46,23 +46,6 @@ const filtered = computed(() => {
     })
     .filter((i) => i) as APIGroup[]
 })
-
-// same as vitepress' slugify logic
-function slugify(text: string): string {
-  return (
-    text
-      // Replace special characters
-      .replace(/[\s~`!@#$%^&*()\-_+=[\]{}|\\;:"'<>,.?/]+/g, '-')
-      // Remove continuous separators
-      .replace(/\-{2,}/g, '-')
-      // Remove prefixing and trailing separators
-      .replace(/^\-+|\-+$/g, '')
-      // ensure it doesn't start with a number (#121)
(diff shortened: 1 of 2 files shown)


src/api/ApiIndex.vue
```diff
@@ -95,7 +78,7 @@ function slugify(text: string): string {
           <h3>{{ item.text }}</h3>
           <ul>
             <li v-for="h of item.headers" :key="h.anchor">
-              <a :href="item.link + '.html#' + slugify(h.anchor)">{{ h.anchor }}</a>
+              <a :href="item.link + '.html#' + h.anchor">{{ h.text }}</a>
```
skirtles-code, line 81: Changing `h.anchor` to `h.text` is the only part of this PR that is strictly necessary to fix the heading text. However...

This code has been through a few previous iterations and, in my opinion, the call to `slugify()` no longer makes sense. The anchor should already be suitably slugified. The other changes are primarily aimed at ensuring that `anchor` is already in the correct form to be used without needing further manipulation.

src/api/api.data.ts
```diff
@@ -57,22 +60,39 @@ function parsePageHeaders(link: string) {
   const h2s = src.match(/^## [^\n]+/gm)
   let headers: APIHeader[] = []
   if (h2s) {
+    const anchorRE = /\{#([^}]+)\}/
```
skirtles-code, line 63: I changed this RegExp. In practice, the previous RegExp worked fine for the current headers in the API, but it assumed that the anchors can only contain letters, numbers and hyphens. We already have headers elsewhere in the docs that use other characters and future additions to the API may do the same.

PR #2080, opened 2022-11-13
Update forms.md

23olddog:
## Description of Problem
When I was reading the chapter "Form Input Bindings" about single-line text and multi-line text input, if the content I typed was consecutive characters without spaces, the content would overflow outside the demo module, affecting the reading, I wondered if this was a potential bug, so I added the word-break CSS property to them'word-break: break-word; '
image

## Proposed Solution
I added'word-break: break-word; 'to the < p > tag to make it wrap automatically to solve this problem, and the out of memory problem will be solved after modification
image

### DIFF
--- a/src/guide/essentials/forms.md
+++ b/src/guide/essentials/forms.md
@@ -57,7 +57,7 @@ In addition, `v-model` can be used on inputs of different types, `<textarea>`, a
 ```
 
 <div class="demo">
-  <p>Message is: {{ message }}</p>
+  <p style="word-break:break-word;">Message is: {{ message }}</p>
   <input v-model="message" placeholder="edit me" />
 </div>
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

> and the out of memory problem will be solved after modification

I'm not sure what you mean by this. Could you explain this part?

23olddog:
> Thanks for the PR!
> 
> > and the out of memory problem will be solved after modification
> 
> I'm not sure what you mean by this. Could you explain this part?



> Thanks for the PR!
> 
> > and the out of memory problem will be solved after modification
> 
> I'm not sure what you mean by this. Could you explain this part?

Sorry for the misunderstanding of my expression. This sentence means that when I add the attribute of'word-break: break-word 'to the < p > tag, the text in the < p > tag will not exceed the scope of the demo module

skirtles-code:
I believe this has been fixed by an upstream change to the theme, https://github.com/vuejs/theme/pull/79, so we no longer need to fix it directly in the docs.

Thanks again for bringing the problem to our attention.

PR #2086, opened 2022-11-15
Update application.md example

helloann:
maybe tag.startsWith('icon-')  is better

### DIFF
--- a/src/api/application.md
+++ b/src/api/application.md
@@ -422,7 +422,7 @@ Specifies a check method to recognize native custom elements.
   ```js
   // treat all tags starting with 'ion-' as custom elements
   app.config.compilerOptions.isCustomElement = (tag) => {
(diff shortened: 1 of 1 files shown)


src/api/application.md
```diff
@@ -422,7 +422,7 @@ Specifies a check method to recognize native custom elements.
   ```js
   // treat all tags starting with 'ion-' as custom elements
   app.config.compilerOptions.isCustomElement = (tag) => {
-    return tag.startsWith('ion-')
+    return tag.startsWith('icon-')
```
NataliaTepluhina, line 425: Could you please provide more context on why do you think `icon` is better? `ion-` is simply a prefix here, I doubt that `icon-table` would make more sense 🤔

helloann: I think you are right. Thank you. @skirtles-code

PR #2087, opened 2022-11-15
Update web-components.md example

helloann:
Maybe change tag.startsWith('ion-')  to tag.startsWith('icon-')  is more reasonable

## Description of Problem
tag.startsWith('ion-')  seems to be a syntax error 

## Proposed Solution
tag.startsWith('icon-')   is more reasonable


skirtles-code:
See discussion at <https://github.com/vuejs/docs/pull/2086#discussion_r1022404477>.

PR #2088, opened 2022-11-15
Update reactivity-core.md

helloann:
watch() shares the same flush timing、 debugging options、cleanup callback and stopping the watcher with watchEffect()

## Description of Problem
Missing `cleanup callback` and `stopping  the watcher` example

## Proposed Solution
add `cleanup callback` and  `stopping  the watcher` example


helloann:
Looking forward your opinion @skirtles-code

PR #2091, opened 2022-11-17
Update reactivity-utilities.md

myquilt1101:
In fact, the source object can be a reactive object or a plain object. Also, the definition of toRef illustrates this point.

### DIFF
--- a/src/api/reactivity-utilities.md
+++ b/src/api/reactivity-utilities.md
@@ -41,7 +41,7 @@ Returns the inner value if the argument is a ref, otherwise return the argument 
 
 ## toRef() {#toref}
 
-Can be used to create a ref for a property on a source reactive object. The created ref is synced with its source property: mutating the source property will update the ref, and vice-versa.
+Can be used to create a ref for a property on a source object. The created ref is synced with its source property: mutating the source property will update the ref, and vice-versa.
 
 - **Type**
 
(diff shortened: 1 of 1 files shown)


src/api/reactivity-utilities.md
```diff
@@ -102,7 +102,7 @@ Can be used to create a ref for a property on a source reactive object. The crea
 
 ## toRefs() {#torefs}
 
-Converts a reactive object to a plain object where each property of the resulting object is a ref pointing to the corresponding property of the original object. Each individual ref is created using [`toRef()`](#toref).
+Converts a object to a plain object where each property of the resulting object is a ref pointing to the corresponding property of the original object. Each individual ref is created using [`toRef()`](#toref).
```
skirtles-code, line 105: I believe the object passed to `toRefs` does need to be reactive. There's a warning logged if it isn't:

<https://github.com/vuejs/core/blob/f3e4f038bf299794b2ca4750b4678074660113c8/packages/reactivity/src/ref.ts#L205>

myquilt1101: Thanks for your reply, I just read the ref.ts and got it.
Now I wonder if "a reactive source object" is exactly correct in the description of "toRef".

myquilt1101: > I believe the object passed to `toRefs` does need to be reactive. There's a warning logged if it isn't:
> 
> https://github.com/vuejs/core/blob/f3e4f038bf299794b2ca4750b4678074660113c8/packages/reactivity/src/ref.ts#L205

Thanks for your reply, I just read the ref.ts and got it.
Now I wonder if "a reactive source object" is exactly correct in the description of "toRef".

PR #2092, opened 2022-11-17
Simplified example

helloann:
Simplified example

## Description of Problem
It use array.reduce before, but maybe we can change the example to make it easier to understand

### DIFF
--- a/src/guide/reusability/plugins.md
+++ b/src/guide/reusability/plugins.md
@@ -65,9 +65,9 @@ export default {
     app.config.globalProperties.$translate = (key) => {
       // retrieve a nested property in `options`
       // using `key` as the path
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

I do generally agree that examples should avoid using `reduce()`, as it tends to make an example difficult to follow. However, the proposed simplification here would change the behaviour of the example. The implementation using `reduce()` allows for arbitrary depth in the path, whereas the proposed change would only allow for a single dot.

For example, the existing implementation would allow for something like `$translate('admin.config.settings.label')`.

I don't think the simplification brings enough benefit to justify losing more realistic functionality.

helloann:
Thanks for your reply, I learned a lot from you .

PR #2097, opened 2022-11-19

jaireina:
Simplifying sentence for more clarity

## Description of Problem
Original phrase was too verbose and a bit confusing.

## Proposed Solution
Changing the "that each component needs to update" to "that each component updates"


skirtles-code (approved):
Thanks!

PR #2099, opened 2022-11-20
Add space in TodosMVC example

dmathijs:
## Description of the Problem
There is a missing space, which wrongly concats the {{ remaining }} value and the 'item'/'items' word.

## Proposed Solution
Added spaces

### DIFF
--- a/src/examples/src/todomvc/App/template.html
+++ b/src/examples/src/todomvc/App/template.html
@@ -45,7 +45,7 @@
   <footer class="footer" v-show="todos.length">
     <span class="todo-count">
       <strong>{{ remaining }}</strong>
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Thanks for the PR!

An alternative fix would be:

```html
<span> item{{ remaining === 1 ? '' : 's' }} left</span>
```

I'm not sure that's really any better though.

I think ideally the space would be between the tags, but it seems a bit fiddly to get that to happen.

I'll leave this open a bit longer, see whether anyone else has any ideas, but I'd be happy to merge this PR as is.

NataliaTepluhina (approved):

PR #2102, opened 2022-11-22
The watch lack of some example

helloann:
watch() will share the same flush timing、 debugging options、cleanup callback and stopping the watcher with watchEffect():

## Description of Problem
It's only  flush timing、 debugging options. Lack of cleanup callback and stopping handler.

## Proposed Solution
Add  cleanup callback and stopping handler.

### DIFF
--- a/src/api/reactivity-core.md
+++ b/src/api/reactivity-core.md
@@ -446,8 +446,12 @@ Watches one or more reactive data sources and invokes a callback function when t
   })
   ```
 
-  `watch()` shares the same flush timing and debugging options with [`watchEffect()`](#watcheffect):
+  `watch()` shares the same flush timing、 debugging options、cleanup callback and stopping the watcher with [`watchEffect()`](#watcheffect):
 
+  **Example**
+   
+  1. flush timing、 debugging options
+  
   ```js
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR! Sorry for the delay in getting back to you.

I've just opened #2155, which attempts to address the same problem. I appreciate the effort you put into creating those examples, but I felt there were some problems with them that would have taken time to fix and it was much quicker for me just to open a new PR.

If you're interested, here are some of the concerns I had:

* We're already in the **Example** section, so there's no need to repeat that heading.
* The text introducing these examples was using a `、`character in place of a comma.
* The examples didn't need to be numbered, they are already in the examples section.
* The `Side effect cleanup` example was inconsistent in its use of `count` and `id`. The comment mentions `id`, when I believe it should mention `count`. The final line also updates `count.value`, which seems unlikely to be correct in this context.
* The example for `Stopping the watcher` seemed unnecessarily complex. Compare that to the equivalent example for `watchEffect()`.

Thanks again for highlighting this gap in the docs and please feel free to give any feedback you have on the changes I've proposed in #2155.

PR #2104, opened 2022-11-22
Update list.md

vivekbsolanki:


skirtles-code:
Thanks for the PR, but I believe the current wording is correct. The word *methods* should be plural as it refers to both `sort()` and `reverse()`.

PR #2105, opened 2022-11-23
Added a warning regarding async setup

### LINKED ISSUE #1032: Guide for `setup` should make it clear that setup has to be synchronous.
I've encountered multiple people trying to use `await` in `setup()`, writing code like:

```js
async setup() {
  const data = await myAPICall()

  return {
    data
  }
}
```

Which fails because `async setup()` can only be used when the component is a descendant of a `Suspense` component - which is an experimental feature currently only documented in the migration guide.

We could therefore add a warning to the [normal Guide about `setup`](https://v3.vuejs.org/guide/composition-api-setup.html) informing users that `setup()` has to be synchronous and `async setup()` is not permitted for the normal, non-Suspense usage.

We might also document workarounds like an iife:

```js
setup() {
  const data = ref()
  
  (
    async () => data.value = await myAPICall()
  )()

  return {
    data
  }
}
```

NataliaTepluhina:
## Description of Problem

Currently we don't have any information in the docs regarding the fact that normally, `setup()` should always be called synchronously. This raises questions from users who try to call `async setup()` and fail.

## Proposed Solution

Add a warning to `setup()` docs

## Additional Information

Close #1032

### DIFF
--- a/src/api/composition-api-setup.md
+++ b/src/api/composition-api-setup.md
@@ -45,7 +45,7 @@ Note that [refs](/api/reactivity-core.html#ref) returned from `setup` are [autom
 :::
 
 :::warning
-`setup()` should always be called _synchronously_. The only case when `async setup()` can be used is when the component is a descentand of a [Suspense](../guide/built-ins/suspense.html) component.
+`setup()` should return an object _synchronously_. The only case when `async setup()` can be used is when the component is a descendant of a [Suspense](../guide/built-ins/suspense.html) component.
 :::
 
 ## Accessing Props {#accessing-props}
--- a/src/api/composition-api-setup.md
+++ b/src/api/composition-api-setup.md
@@ -44,6 +44,10 @@ Note that [refs](/api/reactivity-core.html#ref) returned from `setup` are [autom
 `setup()` itself does not have access to the component instance - `this` will have a value of `undefined` inside `setup()`. You can access Composition-API-exposed values from Options API, but not the other way around.
 :::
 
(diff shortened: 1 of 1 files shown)


src/api/composition-api-setup.md
```diff
@@ -44,6 +44,10 @@ Note that [refs](/api/reactivity-core.html#ref) returned from `setup` are [autom
 `setup()` itself does not have access to the component instance - `this` will have a value of `undefined` inside `setup()`. You can access Composition-API-exposed values from Options API, but not the other way around.
 :::
 
+:::warning
+`setup()` should always be called _synchronously_. The only case when `async setup()` can be used is when the component is a descentand of a [Suspense](../guide/built-ins/suspense.html) component.
```
skirtles-code, line 48:
```suggestion
`setup()` should return an object _synchronously_. The only case when `async setup()` can be used is when the component is a descendant of a [Suspense](../guide/built-ins/suspense.html) component.
```

skirtles-code (changes requested):
Sorry for the delay.

We already have a 'tip' just before this 'warning'. Do you think we could remove the two containers and have them as normal paragraphs instead? I think the words 'Note that' from the earlier paragraph can probably be removed too.

I've made a suggestion about the proposed wording below.

NataliaTepluhina:
@skirtles-code thank you for the review! I've applied the requested changes, could you please take a second look?

skirtles-code (approved):

PR #2114, opened 2022-12-02
Add WebdriverIO as alternative

christian-bromann:
## Description of Problem

Current testing docs only offer PW and NW as alternative solutions. WebdriverIO has released v8 which includes component testing and has dedicated docs how to get up and running when using Vue as framework.

Using JSDOM and JavaScript click events do not seem like a viable approach for testing component these days as they don't mimic user behavior at all. For example, clicks through `fireEvent` of `@testing-library/vue` do not recognise whether another element is blocking the access to an element. I've done a [whole set of experiments](https://github.com/webdriverio/component-testing-examples/tree/main/framework-comparisons) showing that using WebDriver which is a W3C web standard as automation protocol to mimic user behavior prevents stupid errors to pass through and also offers much more capabilities, e.g. user gestures.

 For pure component tests I personally would not advice to use JSDOM at all given that:

- it lacks in certain features (e.g. triggering pseudo states) and has some [known caveats](https://github.com/jsdom/jsdom#caveats)
- it completely ignores the interactivity state of elements, e.g. if my component hides a button or an input is disabled, many user interactions are still possible within JSDOM and non-organic interactions via JS clicks
- mounting and rendering components in an actual browser allows the use of common developer debugging tools like DevTools rather than using e.g. `screen.debug()` to print the DOM tree providing a much better DX
- speed advancements are just minimal, starting a headless browser is almost as fast

## Proposed Solution

As WebdriverIO as viable alternative.

## Additional Information

n/a

### DIFF
--- a/src/guide/scaling-up/testing.md
+++ b/src/guide/scaling-up/testing.md
@@ -225,9 +225,9 @@ We recommend using `@testing-library/vue` for testing components in applications
 
 ### Other Options {#other-options-1}
 
-- [WebdriverIO](https://webdriver.io/docs/component-testing/vue) for cross browser component testing that relies on native user interaction based on standardised automation. Can also be used with Testing Library. 
+- [Nightwatch](https://nightwatchjs.org/) is an E2E test runner with Vue Component Testing support. ([Example Project](https://github.com/nightwatchjs-community/todo-vue))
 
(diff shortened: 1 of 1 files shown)


christian-bromann:
@yyx990803 sorry for the ping but any chance we can get a review for this?

PR #2123, opened 2022-12-11
Add class to the Skip Link in the accessibility.md

orbanszlrd:
- CSS syntax is case-insensitive so it is better to use kebab case (skip-link) instead of camel case (skipLink)
- The anchor tag must have a class of skip-link

### DIFF
--- a/src/guide/best-practices/accessibility.md
+++ b/src/guide/best-practices/accessibility.md
@@ -15,7 +15,7 @@ Typically this is done on the top of `App.vue` as it will be the first focusable
 ```vue-html
 <ul class="skip-links">
   <li>
-    <a href="#main" ref="skipLink">Skip to main content</a>
+    <a href="#main" ref="skipLink" class="skip-link">Skip to main content</a>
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR! Something definitely seems off here with the current example, with a `class` of `skip-links` and CSS for `.skipLink`.

I'm a bit confused by the original code. I don't understand why the `<a>` is wrapped in a `<ul>`/`<li>`. The `:focus` styling should surely be applied to the `<a>`, but I'm not so sure about the rest of the CSS.

@NataliaTepluhina Any idea why this example needs the `<ul>`/`<li>` wrapper? That seems just to be complicating things. The docs themselves have a skip link and don't have a wrapper like that.

NataliaTepluhina:
@skirtles-code I am equally confused (I am not the author of the original code 😅 ). Let's merge this PR and handle removing `ul`/`li` tags in the follow-up

NataliaTepluhina (approved):

PR #2125, opened 2022-12-15
Fix broken portal-vue link

AnaaMarijaa:
## Description of Problem

The issue found is the portal-vue URL under Linusborg leading to a 404 page.

## Proposed Solution

The proposed solution - remove the * at the end of the URL.

Thanks!


NataliaTepluhina (approved):
Thank you for the fix!

PR #2126, opened 2022-12-16
Add missing defineExpose import

raulrpearson:
## Description of Problem

An example on `guide/essentials/template-refs.md` is missing the import of `defineExpose`.

## Proposed Solution

Add it.

### DIFF
--- a/src/guide/essentials/template-refs.md
+++ b/src/guide/essentials/template-refs.md
@@ -222,7 +222,7 @@ An exception here is that components using `<script setup>` are **private by def
 
 ```vue
 <script setup>
-import { ref } from 'vue'
+import { defineExpose, ref } from 'vue'
 
 const a = 1
 const b = ref(2)


skirtles-code:
Thanks for the PR, but `defineExpose` is a compiler macro and doesn't need to be imported.

raulrpearson:
Interesting, I didn't know that. Do you think it would be worth mentioning that instead?

As a beginner I was a bit confused. I tested it out on a project and got a red squiggle under my use of `defineExpose`, so I assumed it required importing. I guess there's something wrong with my setup, then.

If you'd like me to suggest some phrasing for that "compiler macro" piece of information, let me know and I'll add a commit. Or feel free to add it yourself or close the PR if you think that's the best course of action. And thanks for the quick feedback!

skirtles-code:
I've opened #2140 to try to address this. More work may be needed beyond that, as the concept of 'compiler macros' hasn't been introduced properly and we're largely glossing over it in that guide.

PR #2129, opened 2022-12-19
Update coc.md

phedhong:


src/about/coc.md
```diff
@@ -1,4 +1,4 @@
-# Code Of Conduct {#code-of-conduct}
+# Code Of Conduct {#code-of-conduct}
```
NataliaTepluhina, line 1: **question**: Could you please provide more context on why we need whitespace here?

PR #2131, opened 2022-12-20

gass-git:
grammar fix: article "a" missing


skirtles-code:
Thank you for the PR, but I think the current wording is correct.

PR #2136, opened 2022-12-23
Update reactivity-in-depth.md

ritechoice23:
correct a typo


skirtles-code:
Thanks for the PR, but I don't believe this is actually a typo. *Automagical* is an informal computing term referring to something that happens automatically, as if by magic.

PR #2138, opened 2022-12-26
Update application.md

Heniker:
## Description of Problem
Guide  [Augmenting Global Properties](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties) is hard to find, because it is located under Options API section, while it actually has uses in Composition API as well.

## Proposed Solution
Mention [Augmenting Global Properties](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties) under `app.config.globalProperties`.

### DIFF
--- a/src/api/application.md
+++ b/src/api/application.md
@@ -523,7 +523,7 @@ An object that can be used to register global properties that can be accessed on
   }
   ```
   
-- **See also:** [Guide - Augmenting Global Properties](/guide/typescript/options-api.html#augmenting-global-properties)
+- **See also:** [Guide - Augmenting Global Properties](/guide/typescript/options-api.html#augmenting-global-properties) <sup class="vt-badge ts" />
 
 ## app.config.optionMergeStrategies {#app-config-optionmergestrategies}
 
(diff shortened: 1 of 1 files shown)


src/api/application.md
```diff
@@ -522,6 +522,8 @@ An object that can be used to register global properties that can be accessed on
     }
   }
   ```
+  
+- **See also:** [Guide - Augmenting Global Properties](/guide/typescript/options-api.html#augmenting-global-properties)
```
skirtles-code, line 526:
```suggestion
- **See also:** [Guide - Augmenting Global Properties](/guide/typescript/options-api.html#augmenting-global-properties) <sup class="vt-badge ts" />
```

skirtles-code (changes requested):
Good idea.

I suggest adding the TS badge at the end (see below). I'm not sure how consistent we are with this, but links to TS-specific topics are supposed to have that badge.

Heniker:
Yea, that makes sense. Accepted change.

skirtles-code (approved):

PR #2144, opened 2023-01-03
Immediate watchers with the Composition API

skirtles-code:
PR #2069 attempted to introduce a section in the watcher docs about using `immediate: true` with the Composition API. The new additions were fine, but they posed a problem. The existing section about `watchEffect()` is built on the premise that it is *the* way to create an eager watcher. This no longer made sense when it followed a section about `immediate: true`.

This PR includes the original 2 commits from #2069, adding docs for `immediate: true`. These are closely based on the existing section about `immediate: true` with the Options API.

In addition, I've reworked the section about `watchEffect()` to avoid presenting it as just a form of eager watcher. I've changed the main example in that section to something a bit more realistic, though it is still similar to the old example. Most importantly, the previous example wasn't using `immediate: true`, which doesn't really make sense if we've just introduced `immediate: true` in the previous section. I've also tried to explain the benefits of `watchEffect()` a bit more explicitly, as I don't think it's obvious just from looking at the example.


src/guide/essentials/watchers.md
```diff
@@ -245,41 +245,57 @@ export default {
...
+  
+</div>
+
+<div class="composition-api">
+
+We can force a watcher's callback to be executed immediately by passing the `immediate: true` option:
+
+```js
+watch(source, (newValue, oldValue) => {
+  // ...
```
yyx990803, line 257: We should add a bit of descriptive comment here, e.g. `executed immediately, then again when source changes...`

yyx990803 (approved):

PR #2147, opened 2023-01-04
Disable VueJobs widget

IsraelOrtuno:
As discussed with @yyx990803 this PR will disable the Jobs widget in the header of the documentation pages.


yyx990803:
This is an experiment we've running with VueJobs and we are sunsetting it.

PR #2149, opened 2023-01-05
Add caveats around mixing `<script setup>` and `<script>`

skirtles-code:
This PR is based on the conversation on <https://github.com/vuejs/core/issues/6242>.

Using `<script setup>` alongside a separate `<script>` tag was intended as an escape hatch for some specific use cases, as described in the docs. But in practice it has been used in various other ways that were not really intended. However, the tone of the docs does not currently discourage that kind of usage.

This PR makes some of the caveats more explicit.

### DIFF
--- a/src/api/sfc-script-setup.md
+++ b/src/api/sfc-script-setup.md
@@ -239,6 +239,13 @@ export default {
 </script>
 ```
 
+Support for combining `<script setup>` and `<script>` in the same component is limited to the scenarios described above. Specifically:
+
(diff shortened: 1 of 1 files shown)


skirtles-code:
@LinusBorg Would you mind checking that this change is an accurate description of the caveats around combining `<script setup>` and `<setup>`?

yyx990803 (approved):

LinusBorg (approved):

PR #2153, opened 2023-01-07
Move component v-model to its own page

skirtles-code:
## Motivation

The docs for component `v-model` are currently located at the end of the page about component events. I think that made sense historically, but there's now enough material on both topics to justify splitting them into two pages.

While it's clearly necessary to document component events before component `v-model`, I don't think they're really the same topic. I frequently find myself linking to the docs for component `v-model`, and it took me a while to 'learn' that they lived at the bottom of the page about events. Time and again I would look for a separate page, like the page about 'Form Input Bindings' for using `v-model` on native elements.

## Redirects

I've updated internal links, but I've also tried to provide redirects for external links to the `v-model` docs. See the code at the top of `events.md` for details.

## Notes for translators

I've tried to make minimal changes to the text itself. The new `v-model.md` should contain almost the same content that was in `events.md`. The only intentional changes I've made to the text are:

* Updating the headings. In particular, headings are now `##` instead of `###`.
* The top of the page, up to the first example. The previous wording assumed it was part of the page about events, so I've written a brief new introduction.


yyx990803 (approved):
I like this idea.

NataliaTepluhina (approved):
@skirtles-code great work on moving `v-model` to a separate doc!

PR #2158, opened 2023-01-08
Added guidance regarding knowledge of components

sanathkandikanti:
## Description of Problem
This section starts off talking about components and most of the explanation revolves around components.

## Proposed Solution
Added guidance that readers can skip this section and come back later after learning about components.

### DIFF
--- a/src/guide/essentials/lifecycle.md
+++ b/src/guide/essentials/lifecycle.md
@@ -1,5 +1,7 @@
 # Lifecycle Hooks {#lifecycle-hooks}
 
+> This section assumes knowledge of [Components](/guide/essentials/component-basics). Feel free to skip it and come back later.
+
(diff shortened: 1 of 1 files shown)


skirtles-code:
Here are my thoughts on this...

Pretty much every page of the docs is about components in some way. Roughly speaking, the **Essentials** section covers topics that apply when working just within a single component. The **Components In-Depth** section covers topics that apply to working with parent-child relationships.

We include this same warning in 3 other places in the **Essentials** section. In each case it's for topics that are related to working with multiple components in a parent-child relationship.

I don't think that applies here. The page about **Lifecycle Hooks** discusses them from the perspective of a single component, just like other pages in the **Essentials** guide. I can't see anything on that page that assumes knowledge of topics that are covered later in the docs.

sanathkandikanti:
Got it, cognitive load for Lifecycle Hooks sections felt high compared to other sections at the time or reading. Personally i would place it after the "Components Basics" in order. Closing this PR.

Issue #1841, opened 2022-07-20
"Usage Restrictions" confuses me


jasonjiwork:
The following is the original text:

> Composables should only be called synchronously in <script setup> or the setup() hook. In some cases, you can also call them in lifecycle hooks like onMounted().
> 
> These are the contexts where Vue is able to determine the current active component instance. Access to an active component instance is necessary so that:
> 
> Lifecycle hooks can be registered to it.
> 
> Computed properties and watchers can be linked to it for disposal on component unmount.
> 
> TIP
> 
> <script setup> is the only place where you can call composables after usage of await. The compiler automatically restores the active instance context after the async operation for you.

I think this information is important, but it's too abbreviated.
It's easy misunderstanding it. Can we add some code examples?
It seems the first sentence and sentence in Tip are contradictory.

jasonjiwork:
Very disappointed. My question has been filed for five days. No response.
It seems Vue core team is not that high efficiency.

skirtles-code:
Thank you for your suggestion.

Unless I'm missing something it doesn't appear to be urgent. I'm unclear why you expect it to be dealt with within 5 days. This is an open source project and the people working on it are volunteers.

jasonjiwork:
Sorry for my reckless comment!
Vue helps me a lot. Thank you for your greate job.
I just want it to be better.

Issue #1293, opened 2021-10-22
Add a doc section on composables with guidelines about how to write them in the correct idiomatic way


julie777:
There are some significant composable libraries out there. And, since using the composition api and composables is such a great reuse tool it would be nice to know how to avoid problems.

Document the "useFeature" paradigm and how it interacts with the component scope and how reactivity works.
Examples, some of which I still don't understand fully:
- why the useFeature naming is a good idea
- does useFeature need to be exported by name like ```import {useRoute} from 'vue-router``` or should I export it as the default
- global vs local/component scope in the composable
- if I create a composable ```useFeature``` and it needs to use vue router then ```useRouter()``` must be inside the ```useFeature``` function so it ends up in the proper scope in the component.
- how does reactivity work
  - computed function inside useFeature function
  - computed function outside useFeature function
  - passing a component prop to ```useFeature(props.something)```
  - etc

AlexVipond:
Would be great to have this kind of guidance in the style guide. This could be a good issue to start collecting people's opinions.

Here are my opinions on the points you raised (informed by my work on [Baleada Vue Features](https://github.com/baleada/vue-features)):

## why the useFeature naming is a good idea
- Consistent naming is a good idea in general
- useFeature has emerged from React as a standard in the front end community. No good reason to break that rule since it seems to work well in the Vue ecosystem too

I have a personal style guide that requires `useSomething` functions to meet two conditions:
- `useSomething` must internally use reactivity APIs, and possibly perform side effects during a component lifecycle (i.e. I must set up at least one watcher or lifecycle hook)
- `useSomething` must have a return value that includes reactive state


## does useFeature need to be exported by name
Exporting by name and avoiding `export default`, even when we're only exporting one thing from a file, is generally considered the best way to avoid edge cases and difficult bugs with build tools.


## global vs local/component scope in the composable
We shouldn't use global stores until we have a good reason to do so. I'm not clear yet on exactly what those reasons might be, but I believe the VueUse project does some global store stuff internally, so that code is worth studying.


## Using third party composition functions like `useRouter` inside your own composition functions
Always pretend we're writing code inside a `setup` function, even if we've abstracted several steps away from `setup`. Reactivity APIs should never be used outside `setup`, unless we fully understand why and how to use [`effectScope`](https://v3.vuejs.org/api/effect-scope.html#effectscope) to collect and clean up side effects.

[...]

PR #2170, opened 2023-01-14
clarify restrictions on where composables can be invoked

scvnc:
## Description of Problem

I have seen some vue developers interpret the following from the ["Composables -> Usage Restrictions"](https://vuejs.org/guide/reusability/composables.html#usage-restrictions) section: when calling composables in a `<script setup>` or `setup()` context, you must call them synchronously. This is good, however feel like we may need to be clear that composables also should only be called in `<script setup>` or `setup()` regardless of whether they are a/synchronously called, and for the reasons described later in the section.

## Proposed Solution

Change how the section is phrased to emphasise that composables, in all cases, only be called:

1. in `<script setup>` or `setup()` hook
2. synchronously

And/or start discussion about best practices about where composables can be called from.

## Additional Information

We should discuss whether this is in the realm of a hard restriction, simply a "best practice," or clarity on "best practices" for using composables outside of a vue context.  

### Related

- [Vue Chat discussion](https://discord.com/channels/325477692906536972/546534697166045204/1063524833025077278)
- [#7488 Composables usage restrictions][7488]
- [#6114 using composables outside script setup (eg. Pinia action)][6114]  
- [A suggestion that composables are to be only used in a vue context](https://github.com/vuejs/core/discussions/6632#discussioncomment-3695121)
- #1841 
- #1293 
  -  "Reactivity APIs should never be used outside setup, unless we fully understand why and how to use effectScope to collect and clean up side effects."

### Counter examples

Libraries like VueUse have techniques like [`tryOnUnmounted`](https://vueuse.org/shared/tryonunmounted/#tryonunmounted) which seem to imply that their composables could work outside of a vue context. 

Pinia also suggests usage of their `useStore()` composable [outside of the context of a component](https://pinia.vuejs.org/core-concepts/outside-component-usage.html) and other cases where you could use a composable within its framework.

[6114]: https://github.com/vuejs/core/discussions/6114
[7488]: https://github.com/vuejs/core/discussions/7488

### DIFF
--- a/src/guide/reusability/composables.md
+++ b/src/guide/reusability/composables.md
@@ -277,9 +277,9 @@ It is OK to perform side effects (e.g. adding DOM event listeners or fetching da
 
 ### Usage Restrictions {#usage-restrictions}
 
-Composables should only be called **synchronously** in `<script setup>` or the `setup()` hook. In some cases, you can also call them in lifecycle hooks like `onMounted()`.
+Composables should only be called in `<script setup>` or the `setup()` hook. They should also be called **synchronously** in these contexts. In some cases, you can also call them in lifecycle hooks like `onMounted()`.
 
-These are the contexts where Vue is able to determine the current active component instance. Access to an active component instance is necessary so that:
+These restrictions are important because these are the contexts where Vue is able to determine the current active component instance. Access to an active component instance is necessary so that:
 
 1. Lifecycle hooks can be registered to it.
 


scvnc:
While digesting some of the related materials, I'm thinking something like this would be worthwhile to add:

> For routine Vue app development, it is a good rule of thumb to only call composables in a ``<script setup>`` or `setup()` hook. They should also be called **synchronously** in these contexts.
>
> That said, there are some other places where it is safe to call a composable. Those places are where `effectScope` has been properly implemented. A notable example of this is in Pinia, where [composables work when defining a store](https://pinia.vuejs.org/cookbook/composables.html), and with their own rules.

PR #2174, opened 2023-01-18
Update compositon-api.md

csnktms:
## Description of Problem
There is no such thing as casting in typescript, but asserting. There are no actual types during runtime. You are able to trick the compiler into thinking that the left-hand side of the `as` keyword has the shape of the specified type on the right-hand side.

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions

## Proposed Solution
Use `as` instead of `cast` in the docs.

### DIFF
--- a/src/guide/typescript/composition-api.md
+++ b/src/guide/typescript/composition-api.md
@@ -316,7 +316,7 @@ function handleChange(event) {
 </template>
 ```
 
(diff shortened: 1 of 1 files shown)


zhenzhenChange:
But cast has a clearly marked, constructed meaning.
And as is an assertion, which is not so appropriate in this context.

skirtles-code:
Thanks for the PR. I've reworded this sentence in #2330 to dodge the problem.

PR #2186, opened 2023-01-24
Typo in tutorial

piermarcobarbe:
The tutorial refers to `todo`, while (essentially everywhere) it's referred as `todos`.

### DIFF
--- a/src/tutorial/src/step-7/description.md
+++ b/src/tutorial/src/step-7/description.md
@@ -10,7 +10,7 @@ We can use the `v-for` directive to render a list of elements based on a source 
 </ul>
 ```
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

Both `todos` and `todo` are used in the code, but they're different things. `v-for="todo in todos"`. I believe the current wording of the sentence, with `todo`, is correct. I don't think it's intending to describe `todos`.

bencodezen:
> Thanks for the PR!
> 
> Both `todos` and `todo` are used in the code, but they're different things. `v-for="todo in todos"`. I believe the current wording of the sentence, with `todo`, is correct. I don't think it's intending to describe `todos`.

Seconded on this interpretation. I'm inclined to suggest we improve on the naming to reduce confusion in the future. Closing this PR for now in the meantime.

PR #2188, opened 2023-01-26
Update list.md

mcfrego:
## Description of Problem
v-for directive for objects iterate through values, not keys. Text description is wrong.

## Proposed Solution
Change text description to Object.values()

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -134,7 +134,7 @@ You can also use `of` as the delimiter instead of `in`, so that it is closer to 
 
 ## `v-for` with an Object {#v-for-with-an-object}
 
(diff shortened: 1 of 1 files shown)


skirtles-code:
Thanks for the PR!

That sentence is specifically discussing the iteration *order*. I believe the current wording is correct, as Vue uses `Object.keys()` internally to iterate over objects:

https://github.com/vuejs/core/blob/0fbc19f2d034a3a50f5c9114ae58bf8f06cf241d/packages/runtime-core/src/helpers/renderList.ts#L81

I think `Object.values()` would yield the same order, but the way that sentence is currently worded seems to be suggesting that `Object.keys()` is used internally, which is correct.

Also note that the key is provided as the second 'argument' of `v-for`, so `v-for` is effectively iterating over both the keys and the values. In that regard, it isn't misleading to suggest that `v-for` allows you to iterate over the keys.

PR #2194, opened 2023-01-28
docs: hyphenate 'real world' when used as an adjective

de-oz:
## Description of Problem

The expression 'real world' should have a hyphen when it is used as an adjective.

## Proposed Solution

real world -> real-world


skirtles-code (approved):
Thanks!

PR #2195, opened 2023-01-28
docs: fix aria-labelledby example.

ByMykel:
## Description of Problem

The `id` used in the aria-labelledby attribute is duplicated.

## Proposed Solution

Change the `id`'s to make them unique.


skirtles-code (approved):
Thanks, looks good!

PR #2196, opened 2023-01-30
Fix Vue Mastery link

tutorialcode:

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -13,7 +13,7 @@ footer: false
 
 <style src="@theme/styles/vue-mastery.css"></style>
 <div class="vue-mastery-link">
-  <a href="https://www.vuemastery.com/courses-path/beginner" target="_blank">
+  <a href="https://www.vuemastery.com/courses/" target="_blank">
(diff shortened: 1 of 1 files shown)


src/guide/introduction.md
```diff
@@ -13,7 +13,7 @@ footer: false
 
 <style src="@theme/styles/vue-mastery.css"></style>
 <div class="vue-mastery-link">
-  <a href="https://www.vuemastery.com/courses-path/beginner" target="_blank">
+  <a href="https://www.vuemastery.com/courses/" target="_blank">
```
NataliaTepluhina, line 16: @tutorialcode could you please provide more context on why we're changing the URL? Beginner path URL is correct 🤔

NataliaTepluhina (changes requested):

src/guide/introduction.md
tutorialcode: It was a dead link, so we re-routed it to the right URL internally as a quick fix. It should be the new URL.

skirtles-code (approved):

PR #2199, opened 2023-02-01
Add a glossary

skirtles-code:
Deployed: <https://deploy-preview-2199--vuejs.netlify.app/glossary/>

The motivation is to help developers to understand what key technical terms mean in a Vue context. e.g. What is a 'compiler macro' or a 'fragment'? These are intended to be relatively brief introductions to the terms, with links to read more.

The main guide could link to these entries when a term is potentially unclear to the reader, though I haven't currently added any such links.

In some cases, the need for a glossary entry could be an indication that we need to do a better job of explaining a concept in the guide. However, the guide has its own specific role, and trying to clarify the exact meaning of a term can potentially derail the flow or require discussion of topics that haven't been covered yet. The glossary doesn't need to bound by those concerns.

---

Some open discussion points:

- How should the page be styled/themed? The **ON THIS PAGE** list on the right-hand side is far too long.
- The terms are currently in alphabetical order in English. How will this impact translations?
- I'm still not really happy with the opening paragraph at the top that explains the page.

---

Some initial questions I asked when I first opened this PR:

- Am I defining the right terms?
- Are they defined in the right level of detail? It is currently quite inconsistent in the level of detail for different entries.
- What tone should we be aiming for? More formal definitions or chattier descriptions?
- How advanced or basic should it go? Some of the terms describe advanced concepts, whereas others touch on beginner-level Vue ideas. It's difficult to imagine both extremes having the same target audience.

---

Some glossaries we might want to use as reference points:

- Angular: <https://angular.io/guide/glossary>
- React: <https://reactjs.org/docs/glossary.html>

### DIFF
--- a/src/glossary/index.md
+++ b/src/glossary/index.md
@@ -1,10 +1,10 @@
-# Glossary
+# Glossary {#glossary}
 
 This glossary is intended to provide some guidance about the meanings of technical terms that are in common usage when talking about Vue. It is intended to be *descriptive* of how terms are commonly used, not a *prescriptive* specification of how they must be used. Some terms may have slightly different meanings or nuances depending on the surrounding context.
 
 [[TOC]]
 
-## async component
+## async component {#async-component}
 
 An *async component* is a wrapper around another component that allows for the wrapped component to be lazy loaded. This is typically used as a way to reduce the size of the built `.js` files, allowing them to be split into smaller chunks that are loaded only when required.
 
@@ -13,7 +13,7 @@ Vue Router has a similar feature for the [lazy loading of route components](http
 For more details see:
 - [Guide - Async Components](/guide/components/async.html)
 
-## compiler macro
+## compiler macro {#compiler-macro}
 
 A *compiler macro* is special code that is processed by a compiler and converted into something else. They are effectively just a clever form of string replacement.
 
@@ -23,7 +23,7 @@ Macros have limitations on their use that don't apply to normal JavaScript code.
 
 These macros are conceptually similar to dynamic imports using `import()`. While `import()` looks a lot like a function call, it's actually special syntax that gets processed by the bundler as part of the build. Like with `defineProps()`, we can't create an alias using `const myImport = import`. There are also restrictions on exactly what syntax can be used to pass values to the 'function', so that the compiler can understand the values.
 
-## component
+## component {#component}
 
 The term *component* is not unique to Vue. It is common to many UI frameworks. It describes a chunk of the UI, such as a button or checkbox. Components can also be combined to form larger components.
 
@@ -56,7 +56,7 @@ The word 'component' also features in several other terms:
 - [functional component](#functional-component)
 - [Web Component](#web-component)
 
-## composable
+## composable {#composable}
 
 The term *composable* describes a common usage pattern in Vue. It isn't a separate feature of Vue, it's just a way of using the framework's Composition API. As with many patterns, there can be some disagreement about whether specific code qualifies for the label.
 
@@ -72,7 +72,7 @@ Not all JavaScript utility functions are composables. If a function doesn't use 
 
 See [Guide - Composables](/guide/reusability/composables.html) for more details about writing composables.
 
-## Composition API
+## Composition API {#composition-api}
 
 The *Composition API* is a collection of functions used to write components and composables in Vue. It includes the [Reactivity API](#reactivity-api), as well as various other functions, such as `provide()` and `inject()` and lifecycle hooks such as `onMounted()`.
 
@@ -118,7 +118,7 @@ export default {
 
 Even though `computed()` is part of the Composition API, the component above would usually be described as being an Options API component, not a Composition API component.
 
-## custom element
+## custom element {#custom-element}
 
 A *custom element* is a feature of the [Web Components](#web-components) standard, which is implemented in modern web browsers. It refers to the ability to use a custom HTML element in your HTML markup to include a Web Component at that point in the page.
 
@@ -129,7 +129,7 @@ Custom elements should not be confused with the ability to include Vue component
 For more details see:
 - [Guide - Vue and Web Components](/guide/extras/web-components.html)
 
-## directive
+## directive {#directive}
 
 The term *directive* refers to template attributes beginning with the `v-` prefix, or their equivalent shorthands.
 
@@ -141,7 +141,7 @@ For more details see:
 - [Guide - Template Syntax - Directives](/guide/essentials/template-syntax.html#directives)
 - [Guide - Custom Directives](/guide/reusability/custom-directives.html)
 
-## dynamic component
+## dynamic component {#dynamic-component}
 
 The term *dynamic component* is used to describe cases where the choice of which child component to render needs to be made dynamically. Typically, this is achieved using `<component :is="type">`.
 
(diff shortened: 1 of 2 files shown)


NataliaTepluhina:
@skirtles-code thank you for working on this! Even in this unpolished state the draft looks very promising.

> Am I defining the right terms?

I would say "yes". "What is composable" is probably the most common question I hear about Vue and you started right from it 😅 

> Are they defined in the right level of detail? It is currently quite inconsistent in the level of detail for different entries.

IMO, they are _good enough_. Perfect level of detail will come with iteration and community contributions. Different terms will naturally have different level of detail because the scope is very different there.

> What tone should we be aiming for? More formal definitions or chattier descriptions?

IMO, glossary should aim for formal definitions

> How advanced or basic should it go? Some of the terms describe advanced concepts, whereas others touch on beginner-level Vue ideas. It's difficult to imagine both extremes having the same target audience.

Both! Because there will be both kinds of the target audience.

src/glossary/index.md
```diff
@@ -0,0 +1,454 @@
...
+An *async component* is a wrapper around another component that allows for the wrapped component to be lazy loaded. This is typically used as a way to reduce the size of the built `.js` files, allowing them to be split into smaller chunks that are loaded only when required.
+
+Vue Router has a similar feature for the [lazy loading of route components](https://router.vuejs.org/guide/advanced/lazy-loading.html), though this does not use Vue's async components feature.
+
+For more details see:
+- [Guide - Async Components](/guide/components/async.html)
+
+## compiler macro {#compiler-macro}
+
+A *compiler macro* is special code that is processed by a compiler and converted into something else. They are effectively just a clever form of string replacement.
```
yyx990803, line 18:
```suggestion
A *compiler macro* is special code that is processed by a compiler and converted into something else. They are effectively a clever form of string replacement.
```

src/glossary/index.md
```diff
@@ -0,0 +1,454 @@
...
+
+Vue's [SFC](#single-file-component) compiler supports various macros, such as `defineProps()`, `defineEmits()` and `defineExpose()`. These macros are intentionally designed to look like normal JavaScript functions, but they aren't functions at all. These are special strings that the compiler detects and replaces with the real JavaScript code that will run in the browser.
```
yyx990803, line 20:
```suggestion
Vue's [SFC](#single-file-component) compiler supports various macros, such as `defineProps()`, `defineEmits()` and `defineExpose()`. These macros are intentionally designed to look like normal JavaScript functions so that they can leverage the same parser and type inference tooling around JavaScript / TypeScript. However they are not actual functions that are run in the browser. These are special strings that the compiler detects and replaces with the real JavaScript code that will actually be run.
```

src/glossary/index.md
```diff
@@ -0,0 +1,454 @@
...
+
+Macros have limitations on their use that don't apply to normal JavaScript code. For example, you might think that `const dp = defineProps` would allow you to create an alias for `defineProps`, but it'll actually result in an error. There are also limitations on what values can be passed to `defineProps()`, as the 'arguments' have to be processed by the compiler and not at runtime.
+
+These macros are conceptually similar to dynamic imports using `import()`. While `import()` looks a lot like a function call, it's actually special syntax that gets processed by the bundler as part of the build. Like with `defineProps()`, we can't create an alias using `const myImport = import`. There are also restrictions on exactly what syntax can be used to pass values to the `import()` 'function', so that the bundler can understand the values.
```
yyx990803, line 24: The analogy here isn't 100% correct and might be misleading, since dynamic `import()` **is** in fact a native JavaScript feature. It's just bundlers happen to use it (with syntax limitations) as a form of compilation hint.

src/glossary/index.md
```diff
@@ -0,0 +1,454 @@
...
+- [dynamic component](#dynamic-component)
+- [functional component](#functional-component)
+- [Web Component](#web-component)
+
+## composable {#composable}
+
+The term *composable* describes a common usage pattern in Vue. It isn't a separate feature of Vue, it's just a way of using the framework's [Composition API](#composition-api).
+
+* Composables are used to encapsulate and reuse stateful logic.
+* A composable is a function.
```
yyx990803, line 68: I think this should be moved to be the first in the list.

src/glossary/index.md
```diff
@@ -0,0 +1,454 @@
...
+* The function is typically expected to be called during the synchronous execution of a component's `setup()` function (or, equivalently, during the execution of a `<script setup>` block). This ties the invocation of the composable to the current component context, e.g. via calls to `provide()`, `inject()` or `onMounted()`.
+* Composables typically return a plain object, not a reactive object. This object usually contains refs and functions and is expected to be destructured within the calling code.
+
+As with many patterns, there can be some disagreement about whether specific code qualifies for the label. Not all JavaScript utility functions are composables. If a function doesn't use the Composition API then it probably isn't a composable. If it doesn't expect to be called during the synchronous execution of `setup()` then it probably isn't a composable. Composables are specifically used to encapsulate stateful logic, they are not just a naming convention for functions.
+
+See [Guide - Composables](/guide/reusability/composables.html) for more details about writing composables.
+
+## Composition API {#composition-api}
+
```
yyx990803, line 78: I think this section can me made simpler and link to [Composition FAQ](https://vuejs.org/guide/extras/composition-api-faq.html#composition-api-faq) instead.

src/glossary/index.md
```diff
@@ -0,0 +1,454 @@
...
+For more details see:
+- [Guide - Event Handling](/guide/essentials/event-handling.html)
+- [Guide - Component Events](/guide/components/events.html)
+
+## fragment {#fragment}
+
+The term *fragment* refers to a special type of [VNode](#vnode) that is used as a parent for other VNodes, but which doesn't render any elements itself.
+
+The name comes from the similar concept of a [`DocumentFragment`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) in the standard DOM.
```
yyx990803, line 171:
```suggestion
The name comes from the similar concept of a [`DocumentFragment`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) in the native DOM API.
```

src/glossary/index.md
```diff
@@ -0,0 +1,454 @@
...
+The term *virtual DOM* (VDOM) is not unique to Vue. It is a common approach used by several web frameworks for managing updates to the UI.
+
+Browsers use a tree of nodes to represent the current state of the page. That tree, and the JavaScript APIs used to interact with it, are referred to as the *document object model*, or *DOM*.
+
+Manipulating the DOM is a major performance bottleneck. The virtual DOM provides one strategy for managing that.
+
+Rather than creating DOM nodes directly, Vue components generate a description of what DOM nodes they would like. These descriptors are plain JavaScript objects, known as VNodes (virtual DOM nodes). Creating VNodes is relatively cheap.
+
+Every time a component re-renders, the new tree of VNodes is compared to the previous tree of VNodes and any differences are then applied to the real DOM. If nothing has changed then the DOM doesn't need to be touched.
+
```
yyx990803, line 436: Worth adding a note that Vue improves virtual DOM runtime performance with compilers, since many new users nowadays assume virtual DOM to be associated with bad performance due to Svelte / Solid's marketing.

yyx990803:
Any particular reason why all terms start with lowercase? It feels a bit weird to be honest.

skirtles-code:
> Any particular reason why all terms start with lowercase? It feels a bit weird to be honest.

Yeah. I was aiming for the kind of casing you'd see in a dictionary, with words written in the case they would usually have if they were used in the middle of a sentence. So *Composition API* and *Single-File Component* get capital letters, whereas terms like *props* are just lowercase.

For reference, the [Angular glossary](https://angular.io/guide/glossary) uses this same casing convention, though the [React glossary](https://legacy.reactjs.org/docs/glossary.html) does not.

PR #2204, opened 2023-02-03
missed element in TypeScript interface

AloisSeckar:
## Description of Problem
`src/about/team/Member.ts` is missing `linkedin` property in `Social` interface definition. While the page still works as intended since the data are presented in `.json` files, TypeScript check in Visual Studio points out this error on `TeamMember.vue`, where the property is called.

## Proposed Solution
Add missing element into interface definition.


skirtles-code (approved):
Thanks for the fix!

PR #2205, opened 2023-02-04
fix(reactivity-fundamentals): Ref Unwrapping

gaxrai:


NataliaTepluhina (approved):
Thank you for the fix!

PR #2209, opened 2023-02-06
Fix:Error when there is no ecosystem field in the theme configuration…

zhoujiahua:
… navigation

## Description of Problem

In the local launch of the Chinese version of the document when the white screen problem can not be loaded, through the error message found in the document `.vitepress/config.ts` file in the `ThemeConfig` object missing `Ecosystem ` This object will throw an error.

By debug code in the `.vitepress/theme/components/SiteMap.vue` file we when the object `Ecosystem` is missing we use find method return value is `undefined` type; when we finally merge the data in the 'undefined' type to get the property value when the error.

## Proposed Solution

Add null detection when using the return value `ecosystem`.

## Additional Information

Chinese document error.

Japanese documentation reports an error.

### DIFF
--- a/.vitepress/theme/components/SiteMap.vue
+++ b/.vitepress/theme/components/SiteMap.vue
@@ -5,9 +5,8 @@ import { useData } from 'vitepress'
 const data = useData()
 const nav = data.site.value.themeConfig.nav
 const ecosystem = nav.find((i: any) => i.text === 'Ecosystem')
-const items = nav
-  .filter((i: any) => i !== ecosystem && i.items)
-  .concat(ecosystem.items)
+let items = nav.filter((i: any) => i !== ecosystem && i.items)
+items = ecosystem?.items ?  items.concat(ecosystem.items) : items
 </script>
 
 <template>


.vitepress/theme/components/SiteMap.vue
```diff
@@ -5,9 +5,8 @@ import { useData } from 'vitepress'
 const data = useData()
 const nav = data.site.value.themeConfig.nav
 const ecosystem = nav.find((i: any) => i.text === 'Ecosystem')
-const items = nav
-  .filter((i: any) => i !== ecosystem && i.items)
-  .concat(ecosystem.items)
+let items = nav.filter((i: any) => i !== ecosystem && i.items)
+items = ecosystem?.items ?  items.concat(ecosystem.items) : items
```
skirtles-code, line 9: I think this change just suppresses the error rather than fixing it.

I believe it's trying to match this entry in the main navigation:

https://github.com/vuejs/docs/blob/e2c54b9b9d66a6e267e42ac55e3098bd46c35eee/.vitepress/config.ts#L38

This fails because it isn't called 'Ecosystem' in translations.

Perhaps it would be better to base the match on `activeMatch` instead, as that should be the same in all languages?

```js
const ecosystem = nav.find((i: any) => i.activeMatch?.includes('ecosystem'))
```

skirtles-code (changes requested):
Thanks for the PR! Feedback below.

.vitepress/theme/components/SiteMap.vue
zhoujiahua: Nice, your solution will be better.

PR #2211, opened 2023-02-06
Configuring Ecosystem Internationalization Fields

zhoujiahua:
## Description of Problem

The value of the ecosystem field will have different values in different language versions, currently in the `.vitepress/theme/components/SiteMap.vue` file is a fixed value, in different language versions if this fixed value is not modified will throw an error.

## Proposed Solution

Add the ecosystem field to the internationalization to get different values depending on the language version; and you need to add the.

## Additional Information

Currently, the error is thrown in Chinese and Japanese documents because the value of the `ecosystem` field is not available.

Chinese document error.

Japanese documentation reports an error.

### DIFF
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -551,9 +551,9 @@ export const sidebar: ThemeConfig['sidebar'] = {
   ]
 }
 
-// Placeholder of the i18n config for @vuejs-translations.
-// const i18n: ThemeConfig['i18n'] = {
-// }
+const i18n: ThemeConfig['i18n'] = {
+  ecosystem: 'Ecosystem',
+}
 
 export default defineConfigWithTheme<ThemeConfig>({
   extends: baseConfig,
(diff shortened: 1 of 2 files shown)


skirtles-code:
You seem to have opened two PRs to fix the same problem.

I've just commented on #2209 with my suggestion for how we should fix this. I'll leave this open for now, but I think what I suggested there would be preferable.

zhoujiahua:
> You seem to have opened two PRs to fix the same problem.
> 
> I've just commented on #2209 with my suggestion for how we should fix this. I'll leave this open for now, but I think what I suggested there would be preferable.

Nice, thank you for the advice. But for your solution there is still no way to avoid returning `undefined ` values, so for the detection of null values for the field `ecosystem`, I think it still needs to be a friendly catch and throw instead of blocking my page load when I start the application locally. Throwing the error to the client's console-side display while in production environment doesn't make sense to me.

skirtles-code:
Hiding the error would allow for subtle problems to manifest in translations. The ecosystem menu options should be present in all translations. If there is a mistake in the logic then that needs to be fixed, not hidden.

PR #2213, opened 2023-02-06
Add mention about usage of `key` attribute in <Transition>

SaviPrograms:
## Description of Problem
There is no information about that it's possible to use component with `key` attribute to transition bettween same component with different data
## Proposed Solution
Added mention

### DIFF
--- a/src/api/built-in-components.md
+++ b/src/api/built-in-components.md
@@ -102,7 +102,7 @@ Provides animated transition effects to a **single** element or component.
   </Transition>
   ```
   
-  Transition on changing key atribute:
+  Transition on changing key attribute:
   
   ```vue-html
   <Transition>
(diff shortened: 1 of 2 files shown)


src/api/built-in-components.md
```diff
@@ -101,6 +101,14 @@ Provides animated transition effects to a **single** element or component.
     <div v-if="ok">toggled content</div>
   </Transition>
   ```
+  
+  Transition on changing key attribute:
+  
+  ```vue-html
+  <Transition>
+    <div :key="data-id">{{changing-text}}</div>
```
skirtles-code, line 109: The values `data-id` and `changing-text` wouldn't work in practice, as the hyphens would be treated as subtraction operators.

Maybe something like this instead?

```suggestion
  Forcing a transition by changing the `key` attribute:
  
  ```vue-html
  <Transition>
    <div :key="text">{{ text }}</div>
```

skirtles-code (changes requested):
Thanks for the PR!

This is a good point, we should mention `key`.

I've made a suggestion for the API entry. See below.

skirtles-code:
I've integrated these commits into #2250.

PR #2216, opened 2023-02-07
Update template-syntax.md

abirhussain:


src/guide/essentials/template-syntax.md
```diff
@@ -225,7 +225,7 @@ Similarly, you can use dynamic arguments to bind a handler to a dynamic event na
 <a v-on:[eventName]="doSomething"> ... </a>
 
 <!-- shorthand -->
-<a @[eventName]="doSomething">
+<a @[eventName]="doSomething"> ...</a>
```
skirtles-code, line 228: Nice catch.

Could you add in the extra space, to keep it consistent with the other examples?
```suggestion
<a @[eventName]="doSomething"> ... </a>
```

PR #2218, opened 2023-02-07
Fix broken link

xfq:
## Description of Problem

The link is broken. See https://vuejs.org/guide/extras/composition-api-faq.html#what-is-composition-api

## Proposed Solution

Fix the link.


skirtles-code:
Thanks for the fix!

PR #2219, opened 2023-02-07
docs: convert vs. to vs

rxliuli:
## Description of Problem

The documentation uses `vs.` instead of `vs` in many places, which seems like a low-level mistake?

image

## Proposed Solution

Replace all occurrences of `vs.` with `vs`

image

### DIFF
--- a/src/examples/ExampleRepl.vue
+++ b/src/examples/ExampleRepl.vue
@@ -18,8 +18,8 @@ onHashChange(updateExample)
 /**
  * We perform some runtime logic to transform source files into different
  * API / format combinations:
- * - Options vs. Composition
- * - plain HTML vs. SFCs
+ * - Options vs Composition
+ * - plain HTML vs SFCs
  */
 function updateExample() {
   let hash = location.hash.slice(1)
--- a/src/guide/best-practices/production-deployment.md
+++ b/src/guide/best-practices/production-deployment.md
@@ -1,6 +1,6 @@
 # Production Deployment {#production-deployment}
 
-## Development vs. Production {#development-vs-production}
+## Development vs Production {#development-vs-production}
 
 During development, Vue provides a number of features to improve the development experience:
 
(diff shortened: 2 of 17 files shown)


AloisSeckar:
Well, this depends whether Vue.js docs are meant to be written in British or American English. https://www.businesswritingblog.com/business_writing/2022/06/vs-versus-vs-versus-v.html

rxliuli:
Well, I don't know that. But originally found this question in the Chinese docs and was indeed a bit confused when I saw this for the first time. and are still using both, which may indeed be a bit strange for non-native English speakers.

image
image

> ps: I also found that `vs.` in react’s Chinese translation document is translated as `vs`, reference:

- en: https://reactjs.org/docs/glossary.html#controlled-vs-uncontrolled-components
- zh-CN: https://zh-hans.reactjs.org/docs/glossary.html#%E5%8F%97%E6%8E%A7%E7%BB%84%E4%BB%B6-vs-%E9%9D%9E%E5%8F%97%E6%8E%A7%E7%BB%84%E4%BB%B6

AloisSeckar:
You're right, regardless the picked variant, it should be consistent. I'd vote for `vs,` everywhere though. But I am a newb here, let's wait for maintainers to decide.

skirtles-code:
The English docs typically use US English, so I believe `vs.` should be used rather than `vs`. I can't find anywhere in the current English docs that use `vs` without a period. I'm unclear why you're seeing results such as `Computed Caching vs Methods` in the search, as that isn't what I'm seeing currently.

The Chinese translation is free to use whatever convention makes most sense to Chinese readers. The repo is at <https://github.com/vuejs-translations/docs-zh-cn> if you think there's a better convention for translating `vs.`.

rxliuli:
> The English docs typically use US English, so I believe `vs.` should be used rather than `vs`. I can't find anywhere in the current English docs that use `vs` without a period. I'm unclear why you're seeing results such as `Computed Caching vs Methods` in the search, as that isn't what I'm seeing currently.
> 
> The Chinese translation is free to use whatever convention makes most sense to Chinese readers. The repo is at https://github.com/vuejs-translations/docs-zh-cn if you think there's a better convention for translating `vs.`.

If it's already consistent then I should close it

PR #2226, opened 2023-02-12
Fix reference to `defineComponent` type tests

fqp:
## Description of Problem
Broken link

## Proposed Solution
Changed link from https://github.com/vuejs/core/blob/main/test-dts/defineComponent.test-d.tsx to https://github.com/vuejs/core/blob/main/packages/dts-test/defineComponent.test-d.tsx


skirtles-code (approved):
Thanks!

PR #2227, opened 2023-02-12
Simplify the cases

abitwhy:
Remove the tip and make it easy to use.


src/guide/scaling-up/state-management.md
```diff
@@ -172,34 +172,30 @@ import { reactive } from 'vue'
...
-    this.count++
+    store.count++
   }
 })
 ```
 
 ```vue-html{2}
 <template>
-  <button @click="store.increment()">
+  <button @click="store.increment">
```
NataliaTepluhina, line 182: **suggestion**: while I agree with this change, I'd recommend reverting the `this` -> `store` replacement. It doesn't simplify an example and I believe we should demo that you can refer to reactive properties via `this`

NataliaTepluhina (changes requested):

src/guide/scaling-up/state-management.md
abitwhy: Hi~ I'm not sure about that. But if we keep use the `this`, then should also reserve the tip below, which does influence the simplification aim of this pull request.

PR #2229, opened 2023-02-14
docs(guide-extra): change terminal to `vue-termui` link

huynl-96:
## Description of Problem

[Vuminal](https://github.com/ycmjason/vuminal) has already been archived

## Proposed Solution

With Vuminal has already been archived, I suggest we replace it with a more active, maintainable Vue renderer for terminal apps, which is [Vue TermUI](https://github.com/vue-terminal/vue-termui)


webfansplz:
First of all, thank you so much for recommending `Temir`, but i think we can replace it with [vue-termui](https://github.com/vue-terminal/vue-termui).

webfansplz (approved):

NataliaTepluhina (approved):

PR #2243, opened 2023-02-23
re-arrange VueSchoolLink like on other pages

AloisSeckar:
## Description of Problem

Other pages in `Guide/Components In-Depth` (props, events and slots) have first the assumption about user knows the basics and then the VueSchoolLink (if it is provided). `Component Registration` is the only one where first comes the link and the assumption after.

## Proposed Solution

Unless it was deliberate for some reason unclear to me, I suggest to re-arrange it to follow the same pattern on all pages.


tlagoda (commented):
Looks good to me!

skirtles-code (approved):
I can't see any obvious reason for the different ordering either.

PR #2245, opened 2023-02-24
Improve wording

markwatney2016:
## Description of Problem
Current wording is misleading.

## Proposed Solution
Improve wording.


AloisSeckar:
I think your change is misleading. The current word `as` is perfectly right - the method `increment()` is placed there to **serve AS event handler**. There is no other separate event handling function you would use it **with**.

markwatney2016:
Bad pull request.

Issue #2248, opened 2023-02-25
Typescript section still suggests Reactivity transform, without warning


kohasummons:
[Typescript with Composition API](https://github.com/vuejs/docs/blob/main/src/guide/typescript/composition-api.md#typing-component-props) page suggests using the deprecated Reactivity transform as an alternative to `withDefaults` compiler macro.

To help devs move faster, showing a deprecated text box here would be great.

AloisSeckar:
I opened a [PR](https://github.com/vuejs/docs/pull/2249) for this. Suggestions and improvements welcome.

NataliaTepluhina:
I believe we can close this issue now as the recommendation is gone

PR #2249, opened 2023-02-25
Deprecated warning on Reactivity Transform usage

AloisSeckar:
## Description of Problem

Adressing #2248 I wrapped remaining examples of Reactivity Transform into ```:::danger``` block. Not sure if this is the best approach, but it is a simple change and quite expressive outcome.

## Additional Information

This is how it would look:

Reactivity Fundamentals:
image

TypeScript with Composition API:
image

### DIFF
--- a/src/guide/essentials/reactivity-fundamentals.md
+++ b/src/guide/essentials/reactivity-fundamentals.md
@@ -578,6 +578,7 @@ export default {
 
 ## Reactivity Transform <sup class="vt-badge experimental" /> \*\* {#reactivity-transform}
 
+:::danger Deprecated Experimental Feature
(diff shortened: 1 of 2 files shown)


tlagoda (commented):
Looks good to me! 👍

skirtles-code (changes requested):
I think wrapping the entire section in a `danger` box is a bit much visually. Any warning should be similar to the one at the top of <https://vuejs.org/guide/extras/reactivity-transform.html>.

When Evan first announced the deprecation of Reactivity Transforms, he seemed to suggest that the `defineProps` part was going to be retained. That isn't currently mentioned in the post he made last week on the RFC, but I don't think we should change the section about `defineProps` without further clarification.

As for the section in `reactivity-fundamentals.md`, I guess there's a question as to whether we should just remove it, rather than adding a warning. At this point, what would be the motivation behind keeping that section?

AloisSeckar:
I see it already got completely removed from main

PR #2254, opened 2023-02-27
docs: add IDE outline see the difference between option API and compo…

gimjin:
guide/extras/composition-api-faq.html

## Description of Problem
Many developers macroscopically feel composition API in their own code

## Proposed Solution
Tip use IDE outline

## Additional Information
none


NataliaTepluhina:
IThank you for your PR! I am not quite sure how IDE outline should help (also it does depend on the IDE_. Also, we try to limit adding new `tip` blocks because they break the reader's flow. Considering this, I am inclined to close the pull request

PR #2256, opened 2023-02-28
docs: fix a piece of unusable code

ssshooter:
## Description of Problem

The `good` example of [this page](https://vuejs.org/style-guide/rules-use-with-caution.html#implicit-parent-child-communication) is not working.

## Proposed Solution

Fixed that piece of code.

### DIFF
--- a/src/style-guide/rules-use-with-caution.md
+++ b/src/style-guide/rules-use-with-caution.md
@@ -107,6 +107,10 @@ app.component('TodoItem', {
 
 ```js
 app.component('TodoItem', {
+  model: {
+    prop: "todo"
+  },
+  
   props: {
     todo: {
       type: Object,
@@ -114,12 +118,12 @@ app.component('TodoItem', {
     }
   },
 
-  emits: ['input'],
+  emits: ["input"],
 
   template: `
     <input
       :value="todo.text"
-      @input="$emit('input', $event.target.value)"
+      @input="$emit('input', { text: $event.target.value })"
     >
   `
 })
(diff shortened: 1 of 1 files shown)


skirtles-code (commented):
Thanks for the suggestions!

I'm afraid I don't understand the proposed changes.

The example is trying to show 'props down, events up'. It doesn't claim to be an example of using `v-model`. The proposed changes seem to be an attempt to convert the example to `v-model`. However, the changes would only work with Vue 2, not Vue 3. In particular, the `model` option no longer exists in Vue 3.

ssshooter:
> Thanks for the suggestions!
> 
> I'm afraid I don't understand the proposed changes.
> 
> The example is trying to show 'props down, events up'. It doesn't claim to be an example of using `v-model`. The proposed changes seem to be an attempt to convert the example to `v-model`. However, the changes would only work with Vue 2, not Vue 3. In particular, the `model` option no longer exists in Vue 3.

But it seems `$emit` a wrong value, `todo` is an object.

AloisSeckar:
`$emit` can emit whatever you want. In this example `$event.target.value` is a string holding the current text in the html `<input>` element. It depends on the event listener in parent how it deals with the incomming value. But we don't see that here as the example only shows the application of emits in the child component.

Your code is technically correct - instead of plain string, it would emit an object with a "text" prop, but it is unnecesary. You can easily listen to emitted strings and construct the object yourself in parent - if you need to do so. Written like this the component would probably be less flexible, because it forces implementing components to deal with its verbose output.

ssshooter:
> 

Understood, thank you for explaining.

PR #2257, opened 2023-02-28
feat: multiple-v-model-modifiers

liuseen-l:
## Description of Problem
在我看文档的时候，发现对于多个v-model修饰符的描述信息并没有，然后需要自己去调试测试代码看结果

## Proposed Solution
然后我在原来文档的基础上添加了一个对于多个v-model修饰符的操作文档，可能还需要进一步的改进，希望大家也也能够给予一点建议

->
## Description of Problem
When I read the documentation, I found that there was no description information for multiple v-model modifiers, and then I needed to debug the test code myself to see the results

## Proposed Solution
Then I added an operation document for multiple v-model modifiers on the basis of the original document, which may need further improvement, and I hope you can also give a little advice

### DIFF
--- a/src/guide/components/events.md
+++ b/src/guide/components/events.md
@@ -9,7 +9,8 @@ if (typeof window !== 'undefined') {
     '#usage-with-v-model',
     '#v-model-arguments',
     '#multiple-v-model-bindings',
-    '#handling-v-model-modifiers'
+    '#handling-v-model-modifiers',
+    '#multiple-v-model-modifiers'
   ].includes(hash)) {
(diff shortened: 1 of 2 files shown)


skirtles-code (commented):
Thanks for the suggestion.

I'm not quite sure what the motivation is here. The main ideas are already covered in the earlier sections. Adding modifiers for multiple `v-model` props isn't really a special case, it works just the same as adding modifiers for each prop individually. I'm not sure whether this scenario is common enough to justify documenting it explicitly.

PR #2260, opened 2023-03-01
Missing import for mounted

RandSohail:
## Description of Problem
In the first coding example, it supposes to import mounted and then use it in the setup

## Proposed Solution
Simply import it from vue

### DIFF
--- a/src/api/composition-api-setup.md
+++ b/src/api/composition-api-setup.md
@@ -15,7 +15,7 @@ We can declare reactive state using [Reactivity APIs](./reactivity-core.html) an
 
 ```vue
 <script>
-import { ref } from 'vue'
+import { ref, mounted } from 'vue'
 
 export default {
   setup() {


AloisSeckar:
No, it doesn't. It is a lifecycle hook, you don't need to import those (as well as you don't import `setup`).

The current code [works as intended](https://sfc.vuejs.org/#eNpVUc1uwjAMfhUrF0BAwhkVNI677QFyKcWlhSaOEnebVPXd57QUaVLV9ov9/dgZ1CUE/d2jOqoiVbENfLa+dYEiwwARaxihjuRgJU0r663H36l4w7rsO2myHiAh92G9mQFART6xvHvPcMoi68MmU3PNGMgKCYEJGF3oSkYo/Q2IG4xAgVuhw+XrExqiZ5ppURyiXwyyhYjPYMyfcTcbuHyOt/9ZqEPd0X3NTZv0xNzkHIeJaL08hXkPL2CJJQiguPbM5OGj6trqebJqEthurToPw2vIcSzM3CaUwrz5aqfmXe5dGfQjkZc9T8Hsq5CsOi5RrZIdZ2xVwxzS0ZhUV/l2HklTvBv501H8Wocak9tfI/0kjCJs1TKLGv8A1RWe8Q==).

skirtles-code:
Thanks for the suggestion, but I think you've misunderstood.

The `onMounted` function, used inside `setup`, would need importing. But that isn't what's being shown here. This example is showing the `mounted` hook provided by the Options API, which isn't something that can be imported.

PR #2261, opened 2023-03-01
Redundant semicolons at the start of the expressions were removed

unam3:
## Description of Problem
Redundant semicolons at the start of the expressions.

## Proposed Solution
Remove them completely.

## Additional Information
All was found with `git grep -I ";\["`.

### DIFF
--- a/src/examples/src/crud/App/composition.js
+++ b/src/examples/src/crud/App/composition.js
@@ -15,7 +15,7 @@ export default {
     )
 
     watch(selected, (name) => {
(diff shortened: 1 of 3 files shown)


brc-dd:
https://prettier.io/docs/en/rationale.html#semicolons

AloisSeckar:
> https://prettier.io/docs/en/rationale.html#semicolons

Interesting how this clash with https://eslint.org/docs/latest/rules/no-extra-semi

skirtles-code:
See also #1858.

PR #2266, opened 2023-03-04
Remove excess words and fix minor language issues

peterhauke:
## Description of Problem
Fix minor language issues

## Proposed Solution
* Remove "at a time" which is redundant as "at most" is sufficient
* Replace "relevant tooling section" with the name of the section for better accessibility
* Make "Src" lowercase to be consistent with other use of tags and attributes in the Vue documentation, i.e. setup() and <script setup>

### DIFF
--- a/src/api/sfc-spec.md
+++ b/src/api/sfc-spec.md
@@ -36,13 +36,13 @@ export default {
 
 ### `<template>` {#template}
 
-- Each `*.vue` file can contain at most one top-level `<template>` block at a time.
+- Each `*.vue` file can contain at most one top-level `<template>` block.
 
 - Contents will be extracted and passed on to `@vue/compiler-dom`, pre-compiled into JavaScript render functions, and attached to the exported component as its `render` option.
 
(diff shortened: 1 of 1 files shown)


src/api/sfc-spec.md
```diff
@@ -109,7 +109,7 @@ Note that integration with various pre-processors may differ by toolchain. Check
 - [Vue CLI](https://cli.vuejs.org/guide/css.html#pre-processors)
 - [webpack + vue-loader](https://vue-loader.vuejs.org/guide/pre-processors.html#using-pre-processors)
 
-## Src Imports {#src-imports}
+## src Imports {#src-imports}
```
skirtles-code, line 112: If we're using lowercase for `src` because it's code, should we maybe wrap it in backticks too?

```suggestion
## `src` Imports {#src-imports}
```

skirtles-code (changes requested):

skirtles-code (approved):

PR #2268, opened 2023-03-05
Fix minor language errors

peterhauke:
## Description of Problem

Fix minor language errors.

Replace "dumb" with "foolish" as the use of "dumb" may be taken as offensive by some people even though most people do not see this word as offensive.

Make emoji plural. The plural of emoji is both emoji and emojis, but emojis is used elsewhere in this document.


NataliaTepluhina (approved):
@peterhauke thank you for the fixed!

PR #2270, opened 2023-03-06
Change <p> tag to <h2> tag to remove ambiguity

peterhauke:
## Description of Problem

A `<p>` tag is used in the tutorial and the name of the `ref()` in this tutorial is also "p". Using a tag and a variable with the same name can be little confusing. 

I thought of changing the variable name from "p" to "q", for example, but that would involve making changes to more files in this tutorial.

## Proposed Solution

Change `<p>` tag to `<h2>` tag to remove ambiguity

### DIFF
--- a/src/tutorial/src/step-9/App/template.html
+++ b/src/tutorial/src/step-9/App/template.html
@@ -1,1 +1,1 @@
-<p ref="p">hello</p>
+<h2 ref="p">hello</h2>
--- a/src/tutorial/src/step-9/description.md
+++ b/src/tutorial/src/step-9/description.md
@@ -5,7 +5,7 @@ So far, Vue has been handling all the DOM updates for us, thanks to reactivity a
 We can request a **template ref** - i.e. a reference to an element in the template - using the <a target="_blank" href="/api/built-in-special-attributes.html#ref">special `ref` attribute</a>:
 
 ```vue-html
-<p ref="p">hello</p>
+<h2 ref="p">hello</h2>
 ```
 
 <div class="composition-api">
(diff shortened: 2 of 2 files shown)


AloisSeckar:
I believe the `p` was chosen here on purpsoe to demonstrate the [Template refs](https://vuejs.org/guide/essentials/template-refs.html). Naturally you can name your template ref anything you want, but the `p` is there to point out that it refers to a `<p>` element.

peterhauke:
Yes, I realize now that the template ref can be assigned any name I want, but I think I was originally confused by the line "To access the ref, we need to declare a ref with matching name:" and I just got confused between the p tag and the other "p"s. 

I see the example in [Template Refs](https://vuejs.org/guide/essentials/template-refs.html) uses "input" as the ref name as well as the HTML tag.

skirtles-code:
I agree that the overuse of `p` is confusing (likewise the use of `input` in that other example), but using a ref called `p` for an `<h2>` just seems arbitrary.

I would rather have names that make sense as a pair. Like `<p>` and `ref="paragraph"`, or something like that.

peterhauke:
Whereas `paragraph` might work for `p`, I am not sure what could be used for `input`.

Reading further down the Template Refs page to [Refs inside `v-for`](https://vuejs.org/guide/essentials/template-refs.html#refs-inside-v-for), I notice that `itemRefs` is used as the variable name there, so I would suggest going in that direction:

How about one of the following (or something similar):
1. `pElementRef` and `inputElementRef`
2. `elementRef` and `elementRef`
3. `elemRef` and `elemRef`

I would go with the 1. even though the variables are a bit long.

yyx990803:
closed via 32284

PR #2277, opened 2023-03-08
Fix a typo in ElasticHeader.vue

midiidev:
firction -> friction

## Description of Problem
`friction` was spelled as "`firction`" causing editing the property to have no effect.

## Proposed Solution
Change the spelling to `friction`

## Additional Information
Also changed the spelling in the playground example


skirtles-code (approved):
Thanks for the fix!

While testing this I noticed that the example in the docs is broken. It doesn't spring back when released. It works fine in the Playground. That doesn't seem to be related to your change, though.

midiidev:
> Thanks for the fix!
> 
> While testing this I noticed that the example in the docs is broken. It doesn't spring back when released. It works fine in the Playground. That doesn't seem to be related to your change, though.

before my fix, this was still happening. I honestly am not sure what that problem is

PR #2282, opened 2023-03-11
Fix animated ElasticHeader demo

skirtles-code:
The ElasticHeader demo is supposed to ping back when the drag ends, but currently it just freezes in place.

This bug was introduced by a recent change to remove Reactivity Transforms, 1bc0434c0d81f4cd45d96874d0b8982d6e8c9175.

Specifically, the problem is on line 44/45, where the first argument passed to `dynamics.animate` needs to be a reactive object. The animation is driven by changing that object, so passing a throwaway wrapper object doesn't work.

I reverted the code back to the Reactivity Transforms version and then changed `$ref` to `reactive`.

Fixed demo: <https://deploy-preview-2282--vuejs.netlify.app/guide/extras/animation.html#state-driven-animations>

### DIFF
--- a/src/guide/extras/demos/ElasticHeader.vue
+++ b/src/guide/extras/demos/ElasticHeader.vue
@@ -1,20 +1,19 @@
 <script setup>
-import { ref, computed } from 'vue'
+import { reactive, computed } from 'vue'
 import dynamics from 'dynamics.js'
 
 const headerHeight = 120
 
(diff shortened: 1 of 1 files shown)


bencodezen (approved):

bencodezen:
Fantastic investigation into this and fixing it @skirtles-code!

PR #2291, opened 2023-03-20
docs: remove React for mobx

wxsms:
## Description of Problem

mobx is not only for react, it also works on other frameworks, like https://github.com/mobxjs/mobx-vue, so `state management library` might be enough I think. Thanks.


skirtles-code:
While MobX can be used independently of React, I think that is how it tends to be used in practice. In the context of that sentence, I think readers who aren't familiar with MobX could easily misunderstand if we don't qualify it in that way.

wxsms:
OK. I will close this one. Thanks for reply.

PR #2294, opened 2023-03-21
Add shorthand usages with object for `v-bind` and `v-on`

qmhc:
## Description of Problem

Examples of of `v-bind` and `v-on` have no shorthand object usage.

## Proposed Solution

Add usage to the example.

## Additional Information

I don't sure if this is a valid usage. But I think it also need to explain similarly if this isn't valid usage.

If it's valid usage I think it should be added to example.

Another dispute is about `v-slot` shorthand. Is the following usage valid?

```html
<MyComp>
  <template #="params">
    {{ params }}
  </template>
</MyComp>
```


skirtles-code:
I'm going to close this, as I believe this is intentionally not documented.

PR #2298, opened 2023-03-23
the parameter hasn't be used

xxhf:
## Description of Problem
There is no example for the parameter oldQuestion. 
## Proposed Solution
Or it's not necessary just there?


AloisSeckar:
I see your point, because tools like eslint will complain about unused params, but for the sake of the example I think it is worth having it here, so devs are aware how does the syntax looks like.

Maybe lets find some use for it in the basic example and then add a note that unused params can be safely omitted.

PR #2302, opened 2023-03-27
fix: wrong stable version

philfontaine:
## Description of Problem

image

Previous code was pointing to the last release on github including prerelease, so was showing v3.3.0-alpha.5 for instance

## Proposed Solution

The `/latest` endpoint should point to the correct version.

> The latest release is the most recent non-prerelease, non-draft release, sorted by the created_at attribute. The created_at attribute is the date of the commit used for the release, and not the date when the release was drafted or published.

[Docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release)

### DIFF
--- a/src/about/releases.md
+++ b/src/about/releases.md
@@ -8,8 +8,8 @@ import { ref, onMounted } from 'vue'
 const version = ref()
 
 onMounted(async () => {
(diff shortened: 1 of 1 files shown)


skirtles-code (approved):
Thanks!

PR #2304, opened 2023-03-28
fix: Update padding to improve layout consistency

Debbl:
## Description of Problem
The text is breaking incorrectly across lines.

image

## Proposed Solution
https://github.com/vuejs/docs/blob/859e90a1f2683b0d8312b43ad7f38fd9940dff05/src/api/ApiIndex.vue#L161
```css
.api-group {
  padding: 24px 28px;
}
```
image

### DIFF
--- a/src/api/ApiIndex.vue
+++ b/src/api/ApiIndex.vue
@@ -158,7 +158,7 @@ h3 {
   margin-bottom: 20px;
   background-color: var(--vt-c-bg-soft);
   border-radius: 8px;
-  padding: 28px 32px;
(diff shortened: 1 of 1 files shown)


skirtles-code:
This problem seems to have been introduced by an upstream change to the theme, which allows for lines to wrap in the middle of words, <https://github.com/vuejs/theme/pull/79>. The content box has always been too small for the text, but nobody noticed because it wasn't wrapping.

skirtles-code (approved):
Thanks for the PR!

In my testing this seems to work well.

PR #2308, opened 2023-04-01
feat: forms.md added docs references in tip note

SanariSan:
## Description of Problem

Highlighted references for both Options/Composition api styles missing links to corresponding docs sections in tip note

## Proposed Solution

Added links:
- https://vuejs.org/api/options-state.html#data
- https://vuejs.org/api/reactivity-core.html#reactivity-api-core

## Additional Information

Options
Options-fix

Composition
Composition-fix

### DIFF
--- a/src/guide/essentials/forms.md
+++ b/src/guide/essentials/forms.md
@@ -44,7 +44,7 @@ In addition, `v-model` can be used on inputs of different types, `<textarea>`, a
 - `<select>` use `value` as a prop and `change` as an event.
 
 ::: tip Note
-`v-model` will ignore the initial `value`, `checked` or `selected` attributes found on any form elements. It will always treat the current bound JavaScript state as the source of truth. You should declare the initial value on the JavaScript side, using <span class="options-api">the `data` option</span><span class="composition-api">reactivity APIs</span>.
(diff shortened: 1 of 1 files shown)


src/guide/essentials/forms.md
```diff
@@ -44,7 +44,7 @@ In addition, `v-model` can be used on inputs of different types, `<textarea>`, a
 - `<select>` use `value` as a prop and `change` as an event.
 
 ::: tip Note
-`v-model` will ignore the initial `value`, `checked` or `selected` attributes found on any form elements. It will always treat the current bound JavaScript state as the source of truth. You should declare the initial value on the JavaScript side, using <span class="options-api">the `data` option</span><span class="composition-api">reactivity APIs</span>.
+`v-model` will ignore the initial `value`, `checked` or `selected` attributes found on any form elements. It will always treat the current bound JavaScript state as the source of truth. You should declare the initial value on the JavaScript side, using <span class="options-api">the [`data`](https://vuejs.org/api/options-state.html#data) option</span><span class="composition-api">[reactivity APIs](https://vuejs.org/api/reactivity-core.html#reactivity-api-core)</span>.
```
skirtles-code, line 47: Thanks for the PR!

The links should just use paths, rather than full URLs:

```suggestion
`v-model` will ignore the initial `value`, `checked` or `selected` attributes found on any form elements. It will always treat the current bound JavaScript state as the source of truth. You should declare the initial value on the JavaScript side, using <span class="options-api">the [`data`](/api/options-state.html#data) option</span><span class="composition-api">[reactivity APIs](/api/reactivity-core.html#reactivity-api-core)</span>.
```

skirtles-code (changes requested):

Issue #1809, opened 2022-06-28
`Inject()` 3rd argument default value is not `true`.


niceplugin:
`Inject()` 3rd argument default value is not `true`.

However, the [API Docs](https://vuejs.org/api/composition-api-dependency-injection.html#inject) `type` description says `true`.

This can be confusing for many novice developers.

I did not know if this guide document was described incorrectly or if it was an issue of vuejs core, so I wrote it here.

See also: [Playground](https://sfc.vuejs.org/#eNp9kMFugzAMhl/FyoVWgkTbbqhF2nvkQpnpQMWJbNNNqnj3mcFh2qTdYuf3ly9+uNec/X1GV7uTdDxkBUGdcxNpmHJihUckgIFG7DTSAj2nCQqbKCJF6hKJQgvnPXEoFEWLEg5HODfw9PxyXGMhwC1dAT+zZQQ0AdsjTGughMusey2G6mfqdEjkN3q6obfZQ2ugU9gUG1e6za6a2uxHSWT+36Jxv5Do6k197ZnuWkf3rpqlDkH6bv31KD7xNdjJ80w6TOhRpurC6UOQDRxd+YMRrHlHrhjpDRn5P+av6B/uil1soW75ArNvhdE=)

NataliaTepluhina:
@niceplugin thank you for reporting this! This looks like a bug in Vue core rather than a documentation issue (the default value of the 3rd argument should be `true` by default). Would you mind opening an issue in https://github.com/vuejs/core/issues instead?

niceplugin:
Thank you, I will do that.

edison1105:
@NataliaTepluhina

edison1105:
> @niceplugin thank you for reporting this! This looks like a bug in Vue core rather than a documentation issue (the default value of the 3rd argument should be `true` by default). Would you mind opening an issue in [https://github.com/vuejs/core/issues](https://github.com/vuejs/core/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) instead?

I feel this is a ducumentation issue.
see https://github.com/vuejs/core/issues/6194#issuecomment-1169433928

NataliaTepluhina:
@edison1105 I still disagree with this being a documentation problem. I can see the source code but I believe from the UX point of view evaluating the default value as a factory would be better and this should be the default behavior. Please see https://github.com/vuejs/core/issues/6194#issuecomment-1169817540 for the context.

PR #2317, opened 2023-04-12
Update provide-inject.md explaining `inject`s third parameter

davidkrijgsman:
Adds `true` as third parameter in the example indicating the default value should be treated as a factory function and adds explanation in text

## Description of Problem
Example seems incorrect

## Proposed Solution
Update the example


skirtles-code:
Related: #1809.

Issue #2137, opened 2022-12-25
Sentry sponsor link is broken


hugo-t-b:
The Sentry sponsor link at [sponsors.vuejs.org/data.json](https://sponsors.vuejs.org/data.json) ([sentry.io/for/vue?utm_source=vuejs.org&utm_medium=paid-community](https://sentry.io/for/vue?utm_source=vuejs.org&utm_medium=paid-community)) doesn't start with `https://`, so the Sentry links on the docs site don't work (as they open a non-existent page within the docs site).

yyx990803:
This has been fixed at the data source.

PR #2326, opened 2023-04-18
fix: #2137 Prefixing urls with "https://" for sponsors.

ackzell:
## Description of Problem
Just as described in #2137 
The data coming from the endpoint for the Sentry sponsor doesn't have the proper URL.
image

ie. it is missing the protocol part.
See the surrounding sponsor urls:
image

## Proposed Solution
Since we don't have access to update the data (that I am aware of). We can add a quick check and prefix the url with the protocol before rendering it. 

Quick and dirty, but it works 😅 

## Additional Information
Works on my ~machine~ Stackblitz instance!
sentry fix

### DIFF
--- a/.vitepress/theme/components/SponsorsGroup.vue
+++ b/.vitepress/theme/components/SponsorsGroup.vue
@@ -57,7 +57,7 @@ function track(interest?: boolean) {
       <a
         v-for="{ url, img, name } of data[tier]"
         class="sponsor-item"
-        :href="url"
+        :href="url.startsWith('https') ? url : `https://${url}`"
         target="_blank"
         rel="sponsored noopener"
         @click="track()"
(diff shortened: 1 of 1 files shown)


AloisSeckar:
what if some data comes in as `http://...`?

ackzell:
It will break 🤓

.vitepress/theme/components/SponsorsGroup.vue
```diff
@@ -57,7 +57,7 @@ function track(interest?: boolean) {
       <a
         v-for="{ url, img, name } of data[tier]"
         class="sponsor-item"
-        :href="url.startsWith('https') ? url : `https://${url}`"
+        :href="/^https?:\/\//.test(url) ? url : `https://${url}`"
```
ackzell, line 60: @AloisSeckar accounting for the urls with insecure urls :)

yyx990803:
This has been fixed at the data source, thanks for the PR!

PR #2337, opened 2023-04-27
Update composition-api.md

tinohager:
fix missing comma in the code example


kiaking:
👋 This is type definition so we don't need comma 👀 It's same as. Thanks for the contribution though 💪 

```ts
interface Emit {
  (e: 'change', id: number): void
  (e: 'update', value: string): void
}

// Same as this kind of stuff on the same page
interface Book {
  title: string
  year?: number
}
```

AloisSeckar:
this is literally the first time ever I see a type definition without either `,` or `;` 👀  
always learning new things :)

PR #2338, opened 2023-04-28
Update composition-api-faq.md

leoelz:
The links for the examples (immutable data, state machines and RxJS) should be consistent and point to the guide's definition

## Description of Problem
RxJS link is pointing to the RxJS github readme, different than immutable data and state machines links that are poining tho the guide's definitions.

## Proposed Solution
Point RxJS link to the guide's definition.


NataliaTepluhina (approved):
@leoelz thank you for the fix!

PR #2352, opened 2023-05-03
Update render-function.md

louiss0:
These docs don't tell devs how to type functional components. This is the pre 3.3.0 way but it's just an example people need to know how to type functional components

## Description of Problem

As said in the above message  developers are not introduced to typing functional components at all. This is not good. Without his information. They will not be able to type them well. 

## Proposed Solution

Please give devs the right information on how to types functional components. My solution is a pre 3.3 solution. But either way people need this info. Especially the post 3.3 solution.


Heniker:
What would the post 3.3 solution be? The first example in the docs is the same code I came up with and it seems to work fine.

PR #2353, opened 2023-05-04
docs: use semantically correct tag for date

andrewvasilchuk:
## Description of Problem

It's more semantically correct to use `<time>` tag instead of `<span>` for dates.

## Proposed Solution

Let's use `<time>` tag.


NataliaTepluhina (approved):
@andrewvasilchuk thank you!

PR #2354, opened 2023-05-05
docs: add copying versions of the sort and reverse methods

MrNightingale:
## Description of Problem
Currently we're proposing to create a copy of array but modern browsers (except FireFox) are already support ES2023 new array methods that returns new array instead of mutating existing

## Proposed Solution
We can propose to the users to use new copying methods instead of creating array copy. 

## Additional Information
[Can I use](https://caniuse.com/mdn-javascript_builtins_array_tosorted)

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -432,3 +432,9 @@ Be careful with `reverse()` and `sort()` in a computed property! These two metho
 - return numbers.reverse()
 + return [...numbers].reverse()
 ```
+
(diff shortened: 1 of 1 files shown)


yyx990803:
These methods are *very* new and not yet supported in Firefox. I would avoid it until there is universal support in mainstream browsers, otherwise we will have to add caveats to avoid users accidentally using the syntax without proper polyfills.

MrNightingale:
@yyx990803 May I reopen PR since Firefox starts to support these two array methods?

image

image

Issue #2359, opened 2023-05-08
[bug] can not stop calling the 'update' method when elapsed > duration, in the timer example.


wangzhilongh:
It will call the 'update' method very quickly,  even when it doesn't make sense.

PR #2360, opened 2023-05-08
fix: add conditional judgment  when call update in the timer example. #2359

wangzhilongh:
…, in the timer example. #2359

### DIFF
--- a/src/examples/src/timer/App/composition.js
+++ b/src/examples/src/timer/App/composition.js
@@ -1,27 +1,41 @@
-import { ref, onUnmounted } from 'vue'
-
+import { ref, onUnmounted, computed } from 'vue'
 export default {
   setup() {
     const duration = ref(15 * 1000)
     const elapsed = ref(0)
 
-    let lastTime = performance.now()
+    let lastTime
     let handle
+
     const update = () => {
-      const time = performance.now()
-      elapsed.value += Math.min(time - lastTime, duration.value - elapsed.value)
-      lastTime = time
-      handle = requestAnimationFrame(update)
+      elapsed.value = performance.now() - lastTime
+      if (elapsed.value >= duration.value) {
+        cancelAnimationFrame(handle)
+      } else {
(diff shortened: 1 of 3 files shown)


wangzhilongh:
> @wangzhilongh could you please elaborate a little bit more on why this change is necessary?

1.When we open [https://vuejs.org/examples/#timer](url) , the 'update' method will be called very quickly. And it can not stop even if the elapsed time is bigger than the duration time. It just a waste of performance, but it doesn't make any sense.

image

So I think it is nessesary to stop call the 'update' method timely.  

2. In this case, the progress rate can easily be used to do some logical, or to control the details of the page. So I make a computed property 'progressRate' and it can be convenient for everyone to use. 

As a way to solve this problem, I pull this request and hope it can be merged.  
Of course, due to my limited ability, so the code may not be very appropriate. If you have any suggestions, welcome to exchange. And if you have a better solution, I am very happy to see and learn.

wangzhilongh:
> @wangzhilongh could you please elaborate a little bit more on why this change is necessary?

1.When we open [https://vuejs.org/examples/#timer](url) , the 'update' method will be called very quickly. And it can not stop even if the elapsed time is bigger than the duration time. It just a waste of performance, but it doesn't make any sense.

image

So I think it is nessesary to stop call the 'update' method timely.  

2. In this case, the progress rate can easily be used to do some logical, or to control the details of the page. So I make a computed property 'progressRate' and it can be convenient for everyone to use. 

As a way to solve this problem, I pull this request and hope it can be merged.  
Of course, due to my limited ability, so the code may not be very appropriate. If you have any suggestions, welcome to exchange. And if you have a better solution, I am very happy to see and learn.

NataliaTepluhina:
@wangzhilongh thank you for the detailed explanation and especially for the screenshots provided! Now I see that `update` is being called after timer should have been finished indeed.

NataliaTepluhina (approved):

PR #2369, opened 2023-05-14
Fix defineOptions secrtion according to vue 3.3

cbfsocial:
The new defineOptions macro allows declaring component options directly in <script setup>, without requiring a separate <script> block according to vue 3.3


NataliaTepluhina:
@cbfsocial thank you for a contrubution! Since 3.3 is still very fresh, would you mind leaving the previous example but also adding "since 3.3 you can also use `defineOptions` directly in script setup.." (and an updated example).

It's a temporary measure but without it we will most likely have people who didn't update to the latest version creating issues with "this does not work"

NataliaTepluhina:
Closing this one in favor of https://github.com/vuejs/docs/pull/2370

PR #2374, opened 2023-05-19
fix: removed anchor

jay-es:
The `#typescript-only-features` section in `src/api/sfc-script-setup.md` was removed at 914630259ee9c64539071d61b092b9590912eba7.

### DIFF
--- a/src/guide/typescript/overview.md
+++ b/src/guide/typescript/overview.md
@@ -175,10 +175,6 @@ const count = ref(1)
 </template>
 ```
 
-See also:
-
-- [TypeScript-only Features in `<script setup>`](/api/sfc-script-setup.html#typescript-only-features)
-
 ### TypeScript in Templates {#typescript-in-templates}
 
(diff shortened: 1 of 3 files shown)


src/guide/typescript/overview.md
```diff
@@ -175,10 +175,6 @@ const count = ref(1)
 </template>
 ```
 
-See also:
-
-- [TypeScript-only Features in `<script setup>`](/api/sfc-script-setup.html#typescript-only-features)
```
Jinjiang, line 180: I think this link can be replaced by "Using `<script setup>` (with TypeScript)" section: https://vuejs.org/guide/typescript/composition-api.html#using-script-setup

What do you think? :-)

jay-es: @Jinjiang Thank you for your review!

"TypeScript-only Features in `<script setup>`" is now not only props.
I don't think it's necessary to put links to all the features here, so it's okay to delete this.

Jinjiang: Yeah, fair enough. Thanks for the explanation.

Jinjiang (approved):

PR #2378, opened 2023-05-21
Update introduction.md

SourabhRanka:
## Description of Problem
In document it is written as incrementally adoptable.

## Proposed Solution
Whereas it should be written as incrementally adaptable.

## Additional Information
Adopt is to take something over, and to adapt is to change something to suit your needs. Here we mean vuejs is progressive in sense it will adapt as per our needs.

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -77,7 +77,7 @@ The rest of the documentation assumes basic familiarity with HTML, CSS, and Java
 
 ## The Progressive Framework {#the-progressive-framework}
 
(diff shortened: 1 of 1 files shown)


AloisSeckar:
I understand the word **"adoptable"** in this context as _"It is quite easy to learn it step-by-step from the small-yet-working examples to complex scenarios."_

I don't think it is a typo and your suggestion is significantly chaging the meaning (although it is probably also true)

Jinjiang:
I agree with @AloisSeckar that "adoptable" is more exact here.

NataliaTepluhina:
Thank you for your attempt to fix it! However, the `adoptable` was written intentionally, it's not a typo.

PR #2392, opened 2023-05-31
docs(inject/provide): add example for usage in setup context

OrbisK:
# Update provide-inject.md

## Description of Problem

inject and provide are not well documented in the setup context. I struggled for a long time to figure out that when I provide a ref it is also injected as a ref. 

## Proposed Solution

added one example.

## Additional Information

If this is accepted, it might be advisable to add examples for defaults and typescript. Especially also concerning BP.


NataliaTepluhina:
This PR has had no changes for a long time so I am closing it as a stale.

PR #2396, opened 2023-06-09
Rewrite reactivity fundamentals CAPI to start with refs

### LINKED ISSUE #2328: Confusing variable name under Reactivity Fundamentals
URL: https://vuejs.org/guide/essentials/reactivity-fundamentals.html#ref-unwrapping-in-templates

The example code for "Ref Unwrapping in Templates" uses `object` as a variable name:

```ts
const object = { foo: ref(1) }
```

 As `object` is also a TypeScript keyword, this confused me deeply, especially when the variable is referenced again a bit further down:

```ts
const { foo } = object
```

This looks like it is trying to instantiate a new object in an incredibly odd way.

Could we rename the variable in this example to something less confusing, such as `refObject`?

yyx990803:
[Preview](https://deploy-preview-2396--vuejs.netlify.app/guide/essentials/reactivity-fundamentals.html) (**Make sure to toggle API preference to Composition API**)

Also closes #2328

### DIFF
--- a/src/guide/essentials/reactivity-fundamentals.md
+++ b/src/guide/essentials/reactivity-fundamentals.md
@@ -115,9 +115,11 @@ export default {
 <div>{{ count }}</div>
 ```
 
-Notice that we did **not** need to append `.value` when using the ref in the template. For convenience, refs are automatically unwrapped when used inside templates. You can also mutate a ref directly in event handlers:
+Notice that we did **not** need to append `.value` when using the ref in the template. For convenience, refs are automatically unwrapped when used inside templates (with a few [caveats](#caveat-when-unwrapping-in-templates)).
 
-```vue-html
+You can also mutate a ref directly in event handlers:
+
+```vue-html{1}
 <button @click="count++">
   {{ count }}
 </button>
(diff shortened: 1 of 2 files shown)


src/guide/essentials/reactivity-fundamentals.md
```diff
@@ -67,98 +67,153 @@ When you access `this.someObject` after assigning it, the value is a reactive pr
...
 }
 ```
 
 ```vue-html
-<div>{{ state.count }}</div>
+<div>{{ count }}</div>
 ```
 
-Similarly, we can declare functions that mutate reactive state in the same scope and expose them as methods alongside the state:
+Notice that we did **not** need to append `.value` when using the ref in the template. For convenience, refs are automatically unwrapped when used inside templates. You can also mutate a ref directly in event handlers:
```
kiaking, line 119: Maybe link to template unwrapping caveat here? Not 100% confident but nested ref not being unwrapped might be surprising for user if we start from teaching `ref` 👀 

Something like:

```
Note: A `ref` nested inside object will not be unwrapped. Learn more at...
```

PR #2399, opened 2023-06-11
API style switch enhancements

yyx990803:
- Default style to Composition API. First-time users without a saved preference will see a tooltip (see screenshot below).

- Support specifying desired style via URL with `?api=composition` or `?api=options`

- Auto switch to proper API style if URL's target header is only available for a given API style. For example, when visiting `/guide/essentials/reactivity-fundamentals.html#ref` with saved preference set to Options API, the page will be displayed in Composition API. A tooltip will also be displayed to let the user be aware that the displayed style is different from their saved preference.

- API style choice enforced via URL persists for that session. This means:
  - If the user navigates to another page without reloading, the API choice remains the same as what was specified by the initial URL.
  - If the user reloads the page, the API choice is reset to the saved preference from localStorage. 

Screenshot 2023-06-11 at 5 27 32 PM

### DIFF
--- a/.vitepress/theme/components/PreferenceTooltip.vue
+++ b/.vitepress/theme/components/PreferenceTooltip.vue
@@ -1,34 +1,125 @@
 <script setup lang="ts">
-import { onMounted, ref } from 'vue'
-import { preferComposition, preferCompositionKey } from './preferences'
-import { useData } from 'vitepress'
+import { onMounted, ref, watch } from 'vue'
+import {
+  inBrowser,
+  preferComposition,
+  preferCompositionKey
+} from './preferences'
+import { useData, type Header } from 'vitepress'
 
 const show = ref(false)
 const { page } = useData()
 
+type Source = 'url-query' | 'url-header' | 'default'
+let source: Source | false =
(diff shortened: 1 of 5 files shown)


.vitepress/theme/components/PreferenceTooltip.vue
```diff
@@ -0,0 +1,227 @@
...
+  preferCompositionKey
+} from './preferences'
+import { useData, type Header } from 'vitepress'
+
+const show = ref(false)
+const { page } = useData()
+
+type Source = 'url-query' | 'url-header' | 'default'
+let source: Source | false =
+  inBrowser && localStorage.getItem(preferCompositionKey) === null
```
arxpw, line 15: Is this logic a bit too long in length for a ternery here? Think an if statement would make this cleaner?

icarusgk: I think it reads more naturally with this ternary 🤔

lmiller1990:
Options API is still neat, but I think now is a good time to move to Composition API as the default. Looks good to me.

martinszeltins:
100% support this! The Composition API makes for a much cleaner and maintainable code than Options API.

PR #2403, opened 2023-06-12
chore: defining default values for Props using withDefaults

Alfred-Skyblue:
Define props default values using withDefaults.


Alfred-Skyblue:
Sorry, I forgot to modify the title when making changes to the withDefaults in another repository. Thank you for reminding me.

PR #2404, opened 2023-06-12
I changed 'an' to 'a' before 'SPA' because 'S' is not a vowel letter

mojjjtaba:


brc-dd:
Actually vowel/consonant letter in spelling don't matter. It's the sound that matters. In this case, SPA begins with the /ɛs/ sound (pronounced like es-pee-ae). So, `an` is correct here.

yyx990803:
Usage of a vs. an is based on sound, not letter. SPA is an acronym is it is pronounced as "es-pee-ay" which starts with a vowel sound (e).

PR #2413, opened 2023-06-19
onBeforeUnmount to onUnmounted for consistency

lioumens:
## Description of Problem

The rest of this documentation page uses `onUnmounted` for removing event listeners, including the example this composable directly references. The API also suggests using `onUnmounted` for this purpose.

## Proposed Solution

Update the example to use `onUnmounted` as well, for consistency with the rest of the page.


bencodezen (approved):
Thanks for the PR @lioumens!

PR #2422, opened 2023-06-26
docs: remove duplicate defineOptions script setup example

hwkr:
removes an old example that used to document using defineOptions without a setup but now says the same thing twice

## Description of Problem

The docs recommend using the defineOptions macro twice. Looks like it was changed in [PR #2384](https://github.com/vuejs/docs/pull/2384) to always suggest always using `<script setup>` but it didn't remove the second example. 

## Proposed Solution

Remove the first example. (Could remove the other too though 🤷‍♀️ )

### DIFF
--- a/src/guide/components/attrs.md
+++ b/src/guide/components/attrs.md
@@ -79,17 +79,7 @@ If you do **not** want a component to automatically inherit attributes, you can 
 
 <div class="composition-api">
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
Thank you for spotting and fixing this!

PR #2428, opened 2023-07-02
docs(faq): remove outdated information about consider to use vue 2

arashsheyda:
## Description of Problem
removed the section in the FAQ page that mentioned the consider to use Vue 2 due to the lack of stable versions for major ecosystem projects like Nuxt or Vuetify. The information was outdated as there are now stable versions available.


NataliaTepluhina (approved):
@arashsheyda thank you for the fix!

PR #2431, opened 2023-07-06
Add TypeScript With Slots section

tdgao:
## Description of Problem

Missing a typing slots section in Vue slots doc. This section is relevant to Typescript users who are looking to type their slots.

## Proposed Solution

Add the missing section with a link to the defineSlots() docs


NataliaTepluhina (changes requested):
@tdgao thank you for your contribution! However, we don't usually provide TS options in the `guide/components` section. Maybe `guide/typescript` would be a better option?

PR #2436, opened 2023-07-08
docs: change `@vnode-mounted` to `@vue:mounted`

ChaominRuan:
## Description of Problem

image

`@vnode-*` hooks in templates are deprecated. Use the `vue:` prefix instead.

For example, `@vnode-mounted` should be changed to `@vue:mounted`. 

`@vnode-*` hooks support will be removed in 3.4.

## Proposed Solution

Replace `@vnode-mounted` with `@vue:mounted`

### DIFF
--- a/src/examples/src/cells/Cell/template.html
+++ b/src/examples/src/cells/Cell/template.html
@@ -4,7 +4,7 @@
     :value="cells[c][r]"
     @change="update"
     @blur="update"
-    @vnode-mounted="({ el }) => el.focus()"
+    @vue:mounted="({ el }) => el.focus()"
   >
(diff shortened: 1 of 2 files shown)


nazarepiedady:
@ChaominRuan I recommend you use backticks to highlight some important parts from your commit or pull request description. I am talking about something like this `@vnode-mounted` and `@vue:`

ChaominRuan:
> @ChaominRuan I recommend you use backticks to highlight some important parts from your commit or pull request description. I am talking about something like this `@vnode-mounted` and `@vue:`

Thanks for the suggestion, it has been revised. 🚀

NataliaTepluhina (approved):

PR #2437, opened 2023-07-08
refactor: use `@vue:mounted` inside `SpreadSheetCell.vue`

nazarepiedady:
replace `@vnode-mounted` with `@vue:mounted` because it will be removed in version 3.4, so this way we will not receive the deprecation warning.

## Description of Problem

`SpreadSheatCell.vue` component was warning about the deprecated usage of the `@vnode-mounted` hook.

## Proposed Solution

Replace `@vnode-mounted` with `@vue:mounted`

## Additional Information

`@vnode-*` hooks support will be removed in version 3.4.

Every usage of `@vnode-*` hooks should be replaced by `vue:` prefix in the place of `@vnode` then after the colon, we should write the hook name.

### DIFF
--- a/src/guide/extras/demos/SpreadSheetCell.vue
+++ b/src/guide/extras/demos/SpreadSheetCell.vue
@@ -22,7 +22,7 @@ function update(e) {
       :value="cells[c][r]"
       @change="update"
       @blur="update"
-      @vnode-mounted="({ el }) => el.focus()"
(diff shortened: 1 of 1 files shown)


nazarepiedady:
I will close this pull request because the problem is already being solved in the #2436 pull request.

nazarepiedady:
I will close this pull request because the problem is already being solved in the #2436 pull request.

PR #2441, opened 2023-07-20
docs: transition hook add comment

btea:
## Description of Problem

The hook `onEnterCancelled` of `transition` component lacks a comment description.

### DIFF
--- a/src/guide/built-ins/transition.md
+++ b/src/guide/built-ins/transition.md
@@ -362,7 +362,8 @@ function onEnter(el, done) {
 
 // called when the enter transition has finished.
 function onAfterEnter(el) {}
-// called when the enter transition is cancaled before completion.
+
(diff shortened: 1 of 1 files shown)


src/guide/built-ins/transition.md
```diff
@@ -362,6 +362,7 @@ function onEnter(el, done) {
 
 // called when the enter transition has finished.
 function onAfterEnter(el) {}
+// called when the enter transition is cancaled before completion.
```
brc-dd, line 365: there is a typo -- `cancaled`

```suggestion

// called when the enter transition is cancelled before completion.
```

btea: Thanks.

NataliaTepluhina (approved):

PR #2447, opened 2023-07-22
Change the initialization value to match the example output

Jerffersonferreira:


NataliaTepluhina (approved):
@Jerffersonferreira thank you for the fix!

PR #2455, opened 2023-07-29
Use `npm create` instead of `npm init`

sodatea:
For consistency across package managers.

`yarn` and `pnpm` only support the `create` subcommand. `init` means something different for them.

So using `create` leads to the least confusion if the user uses multiple package managers.


bencodezen (approved):
Nice suggestion @sodatea! Approving and merging!

PR #2475, opened 2023-08-12
fix: fix typo

vinsonw2:
## Description of Problem
it should read 'directive'
## Proposed Solution
as proposed change
## Additional Information
none


bencodezen (approved):
Thanks for catching this!

PR #2477, opened 2023-08-14
chore: Update newsletters.md

schalkneethling:
## Description of Problem
The Vue.js Feed newsletter seems discontinued.

## Proposed Solution
Remove the link.

## Additional Information

From what I can tell, the Vue.js Feed newsletter is no longer maintained. It used GetRevue, which was axed by Twitter.


bencodezen:
Thanks for the PR! 

Given that it's a popular resource for the community and the team is still active in sharing info, we don't want to prematurely remove this resource given no official announcement has been made just yet.

If that changes though, feel free to submit a PR with the announcement for reference and we'll be happy to update the list accordingly.

PR #2495, opened 2023-09-05
perf: remove `resize` event listener when `unmounted`

Alfred-Skyblue:
Remove event listeners from the `Repl` when `unMounted` to prevent memory leaks.


NataliaTepluhina (approved):
@Alfred-Skyblue thank you for this addition!

PR #2497, opened 2023-09-08
Update built-in-components.md. Rename <teleport> to <Teleport>

FreshHead:
## Description of Problem
I think teleport component name should be capitalized on this page. Sincere sorry if I'm wrong.
## Proposed Solution
Renaming <teleport... into <Teleport...

### DIFF
--- a/src/api/built-in-components.md
+++ b/src/api/built-in-components.md
@@ -294,17 +294,17 @@ Renders its slot content to another part of the DOM.
   Specifying target container:
 
   ```vue-html
-  <teleport to="#some-id" />
-  <teleport to=".some-class" />
-  <teleport to="[data-teleport]" />
+  <Teleport to="#some-id" />
+  <Teleport to=".some-class" />
+  <Teleport to="[data-teleport]" />
   ```
 
   Conditionally disabling:
 
   ```vue-html
-  <teleport to="#popup" :disabled="displayVideoInline">
+  <Teleport to="#popup" :disabled="displayVideoInline">
     <video src="./my-movie.mp4">
-  </teleport>
+  </Teleport>
   ```
 
 - **See also** [Guide - Teleport](/guide/built-ins/teleport)


src/api/built-in-components.md
```diff
@@ -294,17 +294,17 @@ Renders its slot content to another part of the DOM.
...
+  <Teleport to="#some-id" />
+  <Teleport to=".some-class" />
+  <Teleport to="[data-teleport]" />
   ```
 
   Conditionally disabling:
 
   ```vue-html
-  <teleport to="#popup" :disabled="displayVideoInline">
+  <Teleport to="#popup" :disabled="displayVideoInline">
```
NataliaTepluhina, line 305: **suggestion**: according to [styleguide](https://vuejs.org/style-guide/rules-strongly-recommended.html#component-name-casing-in-templates), self-closing component tags can be capitalized, while opening/closing tags should be lowercase. Let's keep the fixes above and remove this one

NataliaTepluhina (changes requested):

src/api/built-in-components.md
FreshHead: Sorry, I know that you know better, but I did not see this line in the your link :/
I just see:
" In most projects, component names should always be PascalCase in [Single-File Components](https://vuejs.org/guide/scaling-up/sfc) and string templates - but kebab-case in in-DOM templates."
Also lowercase is used to distinguish built-in elements such as <slot>  and <template> 
https://vuejs.org/api/built-in-special-elements.html#component

FreshHead: I am confused. I read it like: "In Single-File Components use PascalCase and kebab in in-DOM templates". Not self-enclosings components should be capitalized and components with content in slot should be lowercased. For me code snippet bellow can easily be in SFC: 
`<Teleport to="#popup" :disabled="displayVideoInline">
  <video src="./my-movie.mp4">
</Teleport>`
It is even more confusing that in the teleport doc page Teleport component is always capitalized: https://vuejs.org/guide/built-ins/teleport.html#basic-usage
And for me lowercasing Teleport component is clashed with convention to distinguish special elements and components:
image

PR #2502, opened 2023-09-12
Update introduction.md

tuujnhj:
add another example


NataliaTepluhina:
Hi @tuujnhj! Could you please provide more context on why this change is necessary?

bencodezen:
Thanks for the PR @tuujnhj!

To keep the example to a minimal, it doesn't make sense to add another example to this page. 

Appreciate the suggestion though!

Issue #2503, opened 2023-09-14
Conflicting naming convention of custom event. camelCase or kebab-case?


leochoo:
Potential conflict in doc: Which naming convention should I use for v-model  custom event for Vue 3?

https://vuejs.org/guide/components/v-model.html -> This uses `camelCase`
https://vuejs.org/guide/components/events.html -> This recommends `kebab-case`
https://eslint.vuejs.org/rules/v-on-event-hyphenation.html -> but this enforces `kebab-case`

AloisSeckar:
The name of the emit defined in a component should be camelCase, but when being listened to in iside parent's template, then kebab-case should be used according to the convention.

The third code snippet in https://vuejs.org/guide/components/v-model.html is inaccurate. Would you like to open a PR? Or I'll do later.

leochoo:
Yes I will open the PR! Thanks for confirming!

brc-dd:
Technically it's correct. `v-model` gets compiled to `update:modelValue` not `update:model-value`, although the latter might be recommended by style guides. (refer JS tab here - https://play.vuejs.org/#eNp9kbFuwjAQhl/FuiWtRGFgQ6FSixjo0FYto5coXEKoY1v2OY0U5d17NiJ0QIy+/7vzd/YAL9bOu4CwgtyXrrEkPFKwz1I3rTWOxCAcVmIUlTOtyBjNpJa6NNpHtHDlcY89iXXEHrLsUep8cZ7EM/hA2FpVEPJJiHwTPJl2p20g0T215oBqLeE6R4JYMJkvpjaYAXm+rmrq+ckbzaZDHCWhNK1tFLoPSw3rSFiJlMSsUMr8vqUauYCzS708Yvlzo37yfaxJ+HTo0XUoYcqocDWyWYy33+/Jcgp5haCYvhN+oTcqRMcz9hr0gbX/ccl2l9670fXeb3tC7S9LRdFIjomXwH+wubP6VXc5X6Y+qUcY/wDG1qNa)

AloisSeckar:
I think for the sake of clarity is better to use kebab-case everywhere in templates, even if it is not stricly necessary in SFCs that are being compiled. If you don't make it a habbit, it may once bite you back while [using in-DOM templates](https://vuejs.org/guide/essentials/component-basics.html#dom-template-parsing-caveats).

Also while using linter, you probably wont get away from that. 

And so it feels better to have it right in the docs. Kudos for @leochoo for spotting it.

[...]

PR #2504, opened 2023-09-14
Update v-model.md naming convention

leochoo:
## Description of Problem
camelCase is being used in the example, not following the convention.

## Proposed Solution
kebab-case should be used when event is being listened to inside parent's template.

## Additional Information
#2503

### DIFF
--- a/src/guide/components/v-model.md
+++ b/src/guide/components/v-model.md
@@ -22,7 +22,7 @@ When used on a component, `v-model` instead expands to this:
 ```vue-html
 <CustomInput
   :modelValue="searchText"
-  @update:modelValue="newValue => searchText = newValue"
+  @update:model-value="newValue => searchText = newValue"
 />
 ```
 
--- a/src/guide/components/v-model.md
+++ b/src/guide/components/v-model.md
@@ -21,7 +21,7 @@ When used on a component, `v-model` instead expands to this:
 
 ```vue-html
 <CustomInput
-  :modelValue="searchText"
+  :model-value="searchText"
   @update:model-value="newValue => searchText = newValue"
 />
 ```


AloisSeckar (changes requested):
while we are suggesting the changes, the prop itself should also be hypethenated here:

image

src/guide/components/v-model.md
```diff
@@ -22,7 +22,7 @@ When used on a component, `v-model` instead expands to this:
 ```vue-html
 <CustomInput
   :modelValue="searchText"
```
AloisSeckar, line 24: should be :model-value="searchText"

leochoo:
@AloisSeckar 
Should we also change this part?
```
1. Bind the `value` attribute of a native `<input>` element to the `modelValue` prop
2. When a native `input` event is triggered, emit an `update:modelValue` custom event with the new value
```

Hesitant because they explain the use of camelCase in [Component Events](https://vuejs.org/guide/components/events.html#emitting-and-listening-to-events)

> Like components and props, event names provide an automatic case transformation. Notice we emitted a camelCase event, but can listen for it using a kebab-cased listener in the parent. As with [props casing](https://vuejs.org/guide/components/props#prop-name-casing), we recommend using kebab-cased event listeners in templates.

AloisSeckar:
> @AloisSeckar Should we also change this part?

No, this is ok. The attribute VALUES inside child component must comply with the props and emits definition, which is:

```
<script setup>
defineProps(['modelValue'])
defineEmits(['update:modelValue'])
</script>
```

But the corresponding attribute NAMES in parent should be kebab-cased to avoid possible clashes.

Because from the JavaScript point of view "modelValue" and "modelvalue" are two distinct things, while html attributes names are [case-insensitive by definition](https://www.w3.org/TR/2010/WD-html-markup-20101019/documents.html#case-insensitivity).

leochoo:
Thanks for confirming! I think we can merge this after your approval now!

NataliaTepluhina (approved):

PR #2515, opened 2023-09-25
feat: add errors reference page

sxzz:
Preview: https://deploy-preview-2515--vuejs.netlify.app/errors/#10

relate https://github.com/vuejs/core/pull/9165

### DIFF
--- a/.vitepress/errors.data.ts
+++ b/.vitepress/errors.data.ts
@@ -1,0 +1,8 @@
+import { defineLoader } from 'vitepress'
+import { errorMessages } from '@vue/compiler-core'
+
+export default defineLoader({
+  load() {
+    return { errorMessages }
+  }
+})
--- a/.vitepress/errors.data.ts
+++ b/.vitepress/errors.data.ts
@@ -1,8 +1,0 @@
-import { defineLoader } from 'vitepress'
-import { errorMessages } from '@vue/compiler-core'
-
-export default defineLoader({
-  load() {
-    return { errorMessages }
-  }
-})
--- a/src/errors/index.md
+++ b/src/errors/index.md
@@ -1,0 +1,28 @@
+<script setup>
+import { data } from '../../.vitepress/errors.data.ts'
+const hash = location.hash.slice(1)
+</script>
+
+# Error Reference {#error-reference}
+
+<table>
+  <thead>
+    <tr>
+      <th>Code</th>
+      <th>Message</th>
+    </tr>
+  </thead>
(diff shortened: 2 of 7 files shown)


src/errors/index.md
```diff
@@ -0,0 +1,28 @@
+<script setup>
+import { data } from '../../.vitepress/errors.data.ts'
+const hash = location.hash.slice(1)
```
brc-dd, line 3: This is breaking SSR, move it to onMounted?

.vitepress/errors.data.ts
```diff
@@ -0,0 +1,8 @@
+import { defineLoader } from 'vitepress'
+import { errorMessages } from '@vue/compiler-core'
+
+export default defineLoader({
+  load() {
+    return { errorMessages }
```
yyx990803, line 6: 53 (or the largest in the keys) is an extension point which is empty, we should filter out empty keys here

.vitepress/errors.data.ts
```diff
@@ -0,0 +1,8 @@
+import { defineLoader } from 'vitepress'
+import { errorMessages } from '@vue/compiler-core'
```
yyx990803, line 2: We should be including runtime errors here as well. In fact, runtime errors are more important than compiler errors because most of the time compilers run in dev mode.

yyx990803 (changes requested):

.vitepress/errors.data.ts
sxzz: Runtime errors didn't be exported. We should do it first.

sxzz: `ErrorTypeStrings` is not exported in `runtime-core`, but exporting will increase the bundle size of vue itself at the same time. Because vue re-exported all of `runtime-core`.

Can we have a better way of exporting? How about exporting errors in another package entry, like `@vue/runtime-core/errors`

sodatea (changes requested):
Here's how React's Error Decoder is implemented:
https://github.com/facebook/react/tree/17.0.2/scripts/error-codes

Note this sentence:
> This file is append-only, which means an existing code in the file will never be changed/removed.

Unless we do something similar to ensure the existing error codes would never change, we may need a version number in the page URL.

.vitepress/errors.data.ts
sxzz: Added runtime errors.

yyx990803:
resolved conflicts and included in #2625

PR #2527, opened 2023-10-07
141 ways of using vuemd ≈

andreacampaci:
## Description of Problem
translation


brc-dd:
Ah, wrong repo. Create the PR against https://github.com/vuejs-translations/docs-it

PR #2537, opened 2023-10-17
Fix `watch` examples

jay-es:
## Description of Problem
The example code for `watch` may cause race conditions.
~~However, adding cleanup process makes the code complex and difficult to understand.
So I added a note.~~
↓
Disable the input field until the request is resolved.

### DIFF
--- a/src/guide/essentials/watchers.md
+++ b/src/guide/essentials/watchers.md
@@ -46,6 +46,10 @@ export default {
 <p>{{ answer }}</p>
 ```
 
+:::info
+The above code may cause race conditions, but the cleanup process is omitted for simplicity.
+:::
+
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@jay-es maybe we could disable the input field until the request is resolved? This way we would be able to skip the cleanup process and also wouldn't require the `info` badge?

jay-es:
@NataliaTepluhina Thank you for your advice. I will try it!

jay-es:
@NataliaTepluhina I fixed. Could you please review it again when you have a moment?:pray:

NataliaTepluhina:
@jay-es thank you for adding the disabled state! I've updated your code a little to use the reactive property for `disabled` state instead of direct DOM manipulation through a ref. Also, let's not care about focusing a field here since at this point we haven't explained template refs to users yet. Could you please check the changes and let me know if you're okay with them? If yes, I'll proceed to merge

jay-es:
@NataliaTepluhina Much smarter code!
Thank you for fixing and please merge.

PR #2541, opened 2023-10-20
Accurate countdown timer

skiff26:
## Description of Problem
N/A

## Proposed Solution
N/A

## Additional Information
Added accurate countdown timer to examples.


bencodezen:
This example is very similar to https://vuejs.org/examples/#timer which already exists. Thank you for the PR though!

PR #2557, opened 2023-10-31
docs: consistent colons in input labels

andrewvasilchuk:


bencodezen (approved):
Thanks for the PR! The consistency is a good catch 👍

PR #2568, opened 2023-11-13
feat: add defineModel

sxzz:
- [x] defineModel → https://vuejs.org/api/sfc-script-setup.html
- [ ] useModel → Composition API add page: Two-way Binding
- [ ] make sure to update "component v-model" section
  - https://vuejs.org/guide/components/v-model.html


yyx990803:
Cherry-picked into #2625

PR #2575, opened 2023-11-19
added toReversed example

aprs00:
## Description of Problem
We should include `toReversed()` option of returning copy of the original array.

## Proposed Solution
```diff
return numbers.reverse()
return [...numbers].reverse()

+ // Or you can use toReversed() method which returns 
+ // a reversed copy of the original array
+ return numbers.toReversed()
```

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -431,4 +431,8 @@ Be careful with `reverse()` and `sort()` in a computed property! These two metho
 ```diff
 - return numbers.reverse()
 + return [...numbers].reverse()
+
+// Or you can use toReversed() method which returns 
+// a reversed copy of the original array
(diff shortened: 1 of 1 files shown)


brc-dd:
Browser compatibility of toReversed is still very limited.

AloisSeckar:
I just wonder - what does "very limited" means? All major browser have it since february...
https://caniuse.com/?search=toReversed

When does new ES feature qualify for being "compatible enough"?

brc-dd:
> When does new ES feature qualify for being "compatible enough"?

Currently anything in ES2020 - the default target used by vite and what's specified in vue tsconfig.

toReversed/toSorted are ES2023 features which weren't even landed in Node until 2-3 months back.

PR #2582, opened 2023-11-29
Update forms.md

Perlover:
I think it's misprint and it's fix.


bencodezen (approved):
Thanks for the PR! This is definitely a grammatical error.

PR #2583, opened 2023-11-30
docs: added events as props and typing with object based emits

pikax:
## Description of Problem
Document typing parameters when using `emits` options syntax. 

Also document of declaring `emits` via `props`.

### DIFF
--- a/src/guide/components/events.md
+++ b/src/guide/components/events.md
@@ -177,7 +177,7 @@ export default {
 
 </div>
 
-The `emits` option and `defineEmits()` macro also support an object syntax, if using typescript you can type arguments, which allows us to perform runtime validation of the payload of the emitted events:
+The `emits` option and `defineEmits()` macro also support an object syntax. If using TypeScript you can type arguments, which allows us to perform runtime validation of the payload of the emitted events:
 
 <div class="composition-api">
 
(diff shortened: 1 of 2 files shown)


src/guide/components/events.md
```diff
@@ -176,14 +177,14 @@ export default {
 
 </div>
 
-The `emits` option and `defineEmits()` macro also support an object syntax, which allows us to perform runtime validation of the payload of the emitted events:
+The `emits` option and `defineEmits()` macro also support an object syntax, if using typescript you can type arguments, which allows us to perform runtime validation of the payload of the emitted events:
```
NataliaTepluhina, line 180:
```suggestion
The `emits` option and `defineEmits()` macro also support an object syntax. If using TypeScript you can type arguments, which allows us to perform runtime validation of the payload of the emitted events:
```

src/guide/components/events.md
```diff
@@ -287,3 +288,17 @@ export default {
 ```
 
 </div>
+
+## Events as Props {#events-props}
+
+You may also declare and pass `events` as `props`, by prefixing the event name with `on` + Capitalized name.
```
NataliaTepluhina, line 294:
```suggestion
You may also declare and pass `events` as `props`, by prefixing the capitalized event name with `on`
```

src/guide/components/events.md
```diff
@@ -287,3 +288,17 @@ export default {
...
+
+:::info
```
NataliaTepluhina, line 296: **suggestion**: we're trying to reduce the number of highlighted blocks. Maybe we can merge those two in one block?

NataliaTepluhina (changes requested):
@pikax thank you for the PR! I've added a few styling suggestions for your consideration

src/guide/components/events.md
pikax: the `info` could be remove the warning is something that should stay since is not something we don't want to recommend :D https://github.com/vuejs/core/pull/8025

NataliaTepluhina: @pikax let's remove the `info` then :)

PR #2585, opened 2023-11-30
docs: add composition api to style-guide

Shinigami92:
## Description of Problem

We had issues with some juniors in our company and told them to read the style-guide section about how to handle props and emits between parent and child components,
but then we found out that there are no composition api examples in that section

The juniors only know Vue 3 and don't know the "old" options-api

## Proposed Solution

Add composition-api style-guide examples

## Additional Information

~~I'm using TypeScript here, please let me know if I should rewrite the examples to plain JavaScript~~
All docs around the style-guide uses JavaScript, so I used JS as well

### DIFF
--- a/src/style-guide/index.md
+++ b/src/style-guide/index.md
@@ -4,10 +4,6 @@ outline: deep
 
 # Style Guide {#style-guide}
 
-:::warning Status Notice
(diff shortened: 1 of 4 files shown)


NataliaTepluhina (approved):
@Shinigami92 thank you for the great contribution! Looks great to me, approved 👍🏻

PR #2590, opened 2023-12-05
style: fix Special Sponsor loading

the-pesar:
## Description of Problem
Special Sponsor section is really messy
Screenshot from 2023-12-05 22-06-36

### DIFF
--- a/.vitepress/theme/components/Home.vue
+++ b/.vitepress/theme/components/Home.vue
@@ -42,28 +42,26 @@ onMounted(async () => {
     </p>
   </section>
 
-  <section id="special-sponsor">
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):

Jinjiang:
FYI: in Chinese doc site, we added a sponsorship welcome link with a `v-else-if="data"` dir if the sponsor list is empty.

https://github.com/vuejs-translations/docs-zh-cn/blob/main/.vitepress/theme/components/Home.vue#L89:L91

which I think might be a better practice. Also the template structure could be more reused (since now to keep the link on Chinese doc site, the template structure becomes different from the English one).

Thanks.

PR #2596, opened 2023-12-08
docs: add `watchEffect` notes

Alfred-Skyblue:
## Description of Problem

I have observed that users often make some mistakes when using `watchEffect`, such as asynchronously accessing reactive variables, resulting in dependencies not being collected, or having an `if` statement with a false condition during the initial execution, leading to uncollected dependencies. These errors can cause `watchEffect` to malfunction. In this pull request, I have added some notes to help users avoid these mistakes.

## Additional Information

My English is not very good. You can correct my grammar or inaccurate wording at will. I hope it can improve the documentation. Thank you.

### DIFF
--- a/src/api/reactivity-core.md
+++ b/src/api/reactivity-core.md
@@ -231,6 +231,9 @@ Takes an object (reactive or plain) or a [ref](#ref) and returns a readonly prox
 
 Runs a function immediately while reactively tracking its dependencies and re-runs it whenever the dependencies are changed.
 
+:::info Usage Note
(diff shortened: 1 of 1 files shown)


fabruex:
Hi, I saw this pull request and I would add this point from last comment by LinusBorg in [https://github.com/vuejs/core/discussions/9773](https://github.com/vuejs/core/discussions/9773), I think it's very important and clear:

> reactive dependencies are re-collected on each run of an effect. This is so that a) new dependencies, that might not have been accessed/read from in a previous run, can be read, and b) old dependencies, that are no longer being read from in the latest run, can be discarded.

PR #2600, opened 2023-12-10
Update docs on usage of external types in defineProps

franklin-tina:
## Description of Problem
The documentation currently lacks explicit reference to the capability of `defineProps` to incorporate types from an external file. 
Also, users should be aware that adding `TypeScript` as a peer dependency is necessary for this feature to work.

## Proposed Solution
Add documentation for when types for `defineProps` are coming from an external file

### DIFF
--- a/src/guide/typescript/composition-api.md
+++ b/src/guide/typescript/composition-api.md
@@ -50,7 +50,7 @@ const props = defineProps<Props>()
 </script>
 ```
 
-This also works if `Props` is imported from an external source. This feature requires TypeScript to be a peer dependency of vue.
+This also works if `Props` is imported from an external source. This feature requires TypeScript to be a peer dependency of Vue.
 
 ```vue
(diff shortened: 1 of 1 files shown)


src/guide/typescript/composition-api.md
```diff
@@ -50,6 +50,20 @@ const props = defineProps<Props>()
...
+
+```vue
+<script setup lang="ts">
+import type { Props } from './foo'
+
+const props = defineProps<Props>()
+</script>
+```
+
+:::info
```
NataliaTepluhina, line 63: **suggestion**: we're trying to avoid the excessive usage of `info` blocks. Could you please move the note about TS being necessary up above the code snippet, and remove the `info` wrapper?

NataliaTepluhina (approved):
Thank you for the fix, @franklin-tina! PR looks great to me, approved 👍🏻

PR #2605, opened 2023-12-13
feat: add pnpm, yarn and bun support in quick-start

the-pesar:
Like [Nuxt installation](https://nuxt.com/docs/getting-started/installation) to have a better DX, I added pnpm, yarn and bun installation commands to the quick-start page.

### DIFF
--- a/src/guide/quick-start.md
+++ b/src/guide/quick-start.md
@@ -2,6 +2,10 @@
 footer: false
 ---
 
+<script setup>
+import { VTCodeGroup, VTCodeGroupTab } from '@vue/theme'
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@the-pesar thank you, this looks great! A little nitpick - do we want to simplify `yarn`-related instructions to have `yarn` instead of `yarn install` and `yarn dev` instead of `yarn run dev`?

NataliaTepluhina (approved):
@the-pesar thank you for the fix! PR looks great to me, approved 👍🏻

PR #2607, opened 2023-12-14
docs: recommend to use new ES2023 Array method

chenxeed:
## Description of Problem

The new ES2023 has new array method to modify the array without mutating the original value. Here, we are still showing the traditional method to copy the original array before running the array method like `sort` or `reverse`.

## Proposed Solution

The new method has 12% higher performance: https://jsperf.app/yipiqe

Thus, we can introduce and encourage user to use the newer array method as well.

## Additional Information

Question: Shall we put more information that this method may not work if the user is using older browser and their scripts are not compiled to support latest browser? ex: Typescript >v5.2

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -432,3 +432,10 @@ Be careful with `reverse()` and `sort()` in a computed property! These two metho
 - return numbers.reverse()
 + return [...numbers].reverse()
 ```
+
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@chenxeed thank you for the contribution! I would probably wait with this addition - because we would need to mention the older browser support and it would overload the `Essentials` docs (we try to keep them as simple and short as possible)

PR #2611, opened 2023-12-16
List transition example fix

AloisSeckar:
## Description of Problem
In the [List with transitions example](https://vuejs.org/examples/#list-transition), when you hit `button`, the array will revert to original `[1, 2, 3, 4, 5]`, but the `id` for newly inserted items remains.

So then 7 is inserted instead of logical 6 and so on. You can end up with stuff like that:
image

## Proposed Solution
Would be more logical to reset the `id` value as well.

I added the command to do it in each API variant.

## Additional Information

### DIFF
--- a/src/examples/src/list-transition/App/composition.js
+++ b/src/examples/src/list-transition/App/composition.js
@@ -14,7 +14,7 @@ export default {
 
     function reset() {
       items.value = getInitialItems()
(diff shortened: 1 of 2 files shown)


NataliaTepluhina (approved):
Thank you, it's a nice little improvement 👍🏻

PR #2615, opened 2023-12-20
fix: change redirection link

jiannanxiboy:
Change redirection link from https://github.com/vuequery to https://github.com/phanan/vuequery


brc-dd:
@phanan Hey, your other details are outdated too 😅

PR #2622, opened 2023-12-22
fix: the link address is incorrect after configuring the base URL

YanxinTang:
## Description of Problem

Then link address is incorrect after configuring the base URL

When configuring the base URL like below: 
```json
{
    "base": "/base/"
}
```
Correct: `'/base/some-link'`
Incorrect: `'/some-link'`

## Proposed Solution

use `withBase` prepend base URL into the link

### DIFF
--- a/.vitepress/theme/components/Home.vue
+++ b/.vitepress/theme/components/Home.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { onMounted } from 'vue'
+import { withBase } from 'vitepress'
 import SiteMap from './SiteMap.vue'
 // import NewsLetter from './NewsLetter.vue'
 import { load, data, base } from './sponsors'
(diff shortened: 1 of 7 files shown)


brc-dd:
And why should we support that here? We aren’t using any base in the site. And I don’t think any translation project has that either.

YanxinTang:
> And why should we support that here? We aren’t using any base in the site. And I don’t think any translation project has that either.

For some security reasons, I cannot access the internet in the work environment.

The private internal docs site in my working environment requires supporting baseURL to deploy. So I have to handle the same stuff after updating to the latest doc. I want to solve from the source to avoid repeating meaningless work. And, I believe there's someone who needs this too.

This PR does not bring new things but it will improve the experience of personal deployment.

yyx990803:
@YanxinTang in that case I think it's better to maintain this in your personal fork as this PR does add unnecessary complexity which is something that we have to deal with in the long run.

PR #2625, opened 2023-12-26
3.4 updates

yyx990803:
Includes:

- https://github.com/vuejs/docs/pull/2515
- https://github.com/vuejs/docs/pull/2568

### DIFF
--- a/src/api/reactivity-core.md
+++ b/src/api/reactivity-core.md
@@ -52,7 +52,7 @@ Takes a getter function and returns a readonly reactive [ref](#ref) object for t
   ```ts
   // read-only
   function computed<T>(
-    getter: () => T,
+    getter: (oldValue: T | undefined) => T,
     // see "Computed Debugging" link below
     debuggerOptions?: DebuggerOptions
   ): Readonly<Ref<Readonly<T>>>
@@ -60,7 +60,7 @@ Takes a getter function and returns a readonly reactive [ref](#ref) object for t
   // writable
   function computed<T>(
     options: {
-      get: () => T
+      get: (oldValue: T | undefined) => T
       set: (value: T) => void
     },
     debuggerOptions?: DebuggerOptions
@@ -112,6 +112,7 @@ Takes a getter function and returns a readonly reactive [ref](#ref) object for t
   - [Guide - Computed Properties](/guide/essentials/computed)
   - [Guide - Computed Debugging](/guide/extras/reactivity-in-depth#computed-debugging)
   - [Guide - Typing `computed()`](/guide/typescript/composition-api#typing-computed) <sup class="vt-badge ts" />
+  - [Guide - Performance - Computed Stability](/guide/best-practices/performance#computed-stability) <sup class="vt-badge" data-text="3.4+" />
 
 ## reactive() {#reactive}
 
--- a/src/api/reactivity-core.md
+++ b/src/api/reactivity-core.md
@@ -361,6 +361,7 @@ Watches one or more reactive data sources and invokes a callback function when t
     flush?: 'pre' | 'post' | 'sync' // default: 'pre'
     onTrack?: (event: DebuggerEvent) => void
     onTrigger?: (event: DebuggerEvent) => void
+    once?: boolean // default: false (3.4+)
   }
   ```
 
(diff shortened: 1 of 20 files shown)


src/error-reference/errors.data.ts
```diff
@@ -0,0 +1,17 @@
+import { defineLoader } from 'vitepress'
+import { errorMessages } from 'vue/compiler-sfc'
+// @ts-expect-error internal api
+import { ErrorTypeStrings } from 'vue'
```
yyx990803, line 4: pending a fix in core to make this work for production builds.

src/api/reactivity-core.md
```diff
@@ -52,15 +52,15 @@ Takes a getter function and returns a readonly reactive [ref](#ref) object for t
   ```ts
   // read-only
   function computed<T>(
-    getter: () => T,
+    getter: (oldValue: T | undefined) => T,
```
Shinigami92, line 55: Is it correct that this is `: T | undefined` instead of `?: T`, if I remember correctly there might be a TS compiler options so that `| undefined` means that it needs to explicitly pass a variable that can be undefined, instead of not passing a variable at all.  
However, because this is a callback, it is not affected by this compile option.

src/api/sfc-script-setup.md
```diff
@@ -227,6 +227,77 @@ const props = withDefaults(defineProps<Props>(), {
...
+// declares "modelValue" prop, consumed by parent via v-model
+const model = defineModel()
+// OR: declares "modelValue" prop with options
+const model = defineModel({ type: String })
+
+// emits "update:modelValue" when mutated
+model.value = 'hello'
+
+// declares "count" prop, consumed by parent via v-model:count
+const count = defineModel('count')
```
Shinigami92, line 246: _some random thoughts_

Is it possible to declare multiple models like that?
```js
// v-model:count
const count = defineModel('count')
// v-model:x
const x = defineModel('x')
```
Or will this result in an error?
What happens if someone accidentally redeclared two times the same?
```js
const count = defineModel('count')
const count2 = defineModel('count')
```
Should this be handled by a linter? Will there be a Vue runtime error?

src/api/sfc-script-setup.md
```diff
@@ -227,6 +227,77 @@ const props = withDefaults(defineProps<Props>(), {
...
+  }
+})
+```
+
+### Usage with TypeScript <sup class="vt-badge ts" /> {#usage-with-typescript}
+
+Like `defineProps` and `defineEmits`, `defineModel` can also receive type arguments to specify the types of the model value and the modifiers:
+
+```ts
+const modelValue = defineModel<string>()
```
Shinigami92, line 290: Why is it that way? It feels like it behaves the opposite of `ref` 🤔 
Here is an example:


first line: when I leave out the types at `ref`, it results in `Ref<undefined>` (legit to me 👍)
second line: when I explicitly use `<boolean>`, `undefined` is not a valid value (also legit to me 👍)

so I would assume that `defineModel<string>()` results in `string` and `defineModel<string | undefined>()` results in `string | undefined` and I don't need a `required` at all (or suddenly mix generics-TS and options object, where as e.g. `defineProps` explicitly forbids this mixing)

Shinigami92 (approved):

src/api/reactivity-core.md
yyx990803: This function is not passed by the user but provided by the framework, so the argument will always be present.

src/api/sfc-script-setup.md
yyx990803: Because it accepts parent value via a prop, so it behaves like a prop (can be `undefined` unless explicitly required)

src/api/sfc-script-setup.md
yyx990803: First case is supported. Second case will throw a compile-time error.

Shinigami92: Could we add docs that different `defineModel`s are supported and can be used that way? I think it is really common to have multiple `v-model:*` but it works differently as to `defineProps` or `defineEmits` where you pass multiple values at once, but for `defineModel` you _suddenly_ need multiple calls.

src/api/sfc-script-setup.md
Shinigami92: ah I see 🤔
should this be also documented, maybe with a `:::tip`?

NataliaTepluhina (approved):

src/api/sfc-script-setup.md
yyx990803: It is already shown in the component v-model guide.

Issue #2627, opened 2023-12-26
Tutorial - Declarative Rendering - Google Chrome Browser Warning


kathleenwestPayItGov:
URL: https://vuejs.org/tutorial/#step-2
API Preferences: Options, HTML

The tutorial code should be updated from 
`import { createApp } from 'vue'`
to 
`import { createApp } from 'vue/dist/vue.esm-bundler.js'`
or include a side note on the page for the code change to be compatible with Google Chrome.

This is the warning, however the html does not render on the page unless this is fixed

`tutorial.html:7 [Vue warn]: Component provided template option but runtime compilation is not supported in this build of Vue. Configure your bundler to alias "vue" to "vue/dist/vue.esm-bundler.js". 
  at <App>
warn2 @ chunk-EFQI2HPI.js?v=4a8bb179:1518
finishComponentSetup @ chunk-EFQI2HPI.js?v=4a8bb179:8905
setupStatefulComponent @ chunk-EFQI2HPI.js?v=4a8bb179:8820
setupComponent @ chunk-EFQI2HPI.js?v=4a8bb179:8752
mountComponent @ chunk-EFQI2HPI.js?v=4a8bb179:7150
processComponent @ chunk-EFQI2HPI.js?v=4a8bb179:7116
patch @ chunk-EFQI2HPI.js?v=4a8bb179:6589
render2 @ chunk-EFQI2HPI.js?v=4a8bb179:7883
mount @ chunk-EFQI2HPI.js?v=4a8bb179:5311
app.mount @ chunk-EFQI2HPI.js?v=4a8bb179:10889
(anonymous) @ tutorial.html:7
Show 10 more frames
Show less`

brc-dd:
Can you try hard reloading the page once? I'm not seeing any such warning.

kathleenwestPayItGov:
@brc-dd in order to re-create the browser warning issue, you will need to copy the code to an independent html file and run it locally.

kathleenwestPayItGov:
This is the working version for your reference:
https://github.com/kathleenwestPayItGov/my-first-vue-project/blob/main/tutorial.html

NataliaTepluhina:
[...]

PR #2641, opened 2024-01-02
Side note about HTML-mode in Tutorial

AloisSeckar:
## Description of Problem

A follow-up to #2627

As a newcommer (and then me as well) got confused about using Vue directly in HTML without a build step, I think there should be more info about this topic to make Tutorial more friendly and help new developers to avoid pitfalls.

## Proposed Solution

Add "Tip" section with some clarification.

The box is placed inside `Tutorial Setting Details` inside `/tutorial/#step-1` which is folded by default and only shows when user actively clicks on it.

## Additional Information

I am not sure about the form, so I am more than happy to read the feedback and change the wording as necessary.

### DIFF
--- a/src/tutorial/src/step-1/description.md
+++ b/src/tutorial/src/step-1/description.md
@@ -24,10 +24,10 @@ If you are an experienced developer coming from Vue 2 or other frameworks, there
 <div class="html">
 
 :::tip
-If you're about to use HTML-mode without a build step in your own applications, make sure you either change imports to 
+If you're about to use HTML-mode without a build step in your own applications, make sure you either change imports to:
 
 ```js
-import { * } from vue/dist/vue.esm-bundler.js
+import { ... } from 'vue/dist/vue.esm-bundler.js'
 ```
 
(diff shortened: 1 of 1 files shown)


src/tutorial/src/step-1/description.md
```diff
@@ -21,6 +21,29 @@ If you are an experienced developer coming from Vue 2 or other frameworks, there
 
 - You can also switch between SFC-mode or HTML-mode. The former will show code examples in <a target="_blank" href="/guide/introduction.html#single-file-components">Single-File Component</a> (SFC) format, which is what most developers use when they use Vue with a build step. HTML-mode shows usage without a build step.
 
+:::tip
```
brc-dd, line 24: hide this on sfc:

```suggestion
:::tip {:class="'html'"}
```

AloisSeckar: Your suggestion didn't work for me for some reason, but I wrapped the part inside <div class="html"> which seems to do the job.

PR #2644, opened 2024-01-03
Simplified language around generics in composition API docs

mattpocock:
Guidelines used: https://www.totaltypescript.com/no-such-thing-as-a-generic


NataliaTepluhina:
Thank you for your contribution! While I agree with the sentiment of the article, I would still keep the current usage of `generic` in Vue docs.

PR #2662, opened 2024-01-13
Document long-form checkbox bindings

bmulholland:
I couldn't find anywhere in the docs that spelled this out. I expected to use `:value`, not `:checked`, because that's what's listed on https://vuejs.org/guide/components/v-model.html#under-the-hood

In the end I had to look on StackOverflow instead of the docs. This is one of the places I went to look before I did that, so hopefully this is a good place for others to find it.

### DIFF
--- a/src/guide/essentials/forms.md
+++ b/src/guide/essentials/forms.md
@@ -126,6 +126,12 @@ Single checkbox, boolean value:
   <label for="checkbox-demo">{{ checked }}</label>
 </div>
 
+The v-model is equivalent to:
+
+```vue-html
+<input type="checkbox" id="checkbox" :checked="checked" @input="checked = $event.target.checked" />
+```
+
 <div class="composition-api">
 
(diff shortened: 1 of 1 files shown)


hexiaokang (approved):

NataliaTepluhina:
@bmulholland thank you for your contribution! It is documented on the very top of the document:

https://vuejs.org/guide/essentials/forms.html#form-input-bindings

Screenshot 2024-01-17 at 16 29 35

bmulholland:
@NataliaTepluhina Well sure, but then it's docs feedback: I didn't read that, and I looked in several places. IMO, details on the specifics of checkboxes belongs in the checkbox-specific section, https://vuejs.org/guide/essentials/forms.html#checkbox, which is one place I looked.

More broadly, if I have feedback about how usable the docs are, is that desirable on your end? Closing the PR without discussion of the underlying feedback leaves me feeling like it's not. I would love to help out and contribute to make the docs more usable for others -- but don't want to feel like it's a fight to do so.

PR #2668, opened 2024-01-15
feat: setup twoslash for codeblocks

antfu:
This is working now

### Short-term

- `twoslash` is not stable yet: https://github.com/twoslashes/twoslash
- Using `shikiji`'s beta https://shikiji.netlify.app/packages/twoslash, which is wait for `twoslash`
- Once `shikiji` v0.10 released and VitePress upgrade to it, the pnpm overrides can be removed

### Long-term

- `__VUE_OPTIONS_API__: true` is required at this moment because [`floating-vue`](https://floating-vue.starpad.dev/) is still using Options API
- `shamefully-hoist=true` is required at this moment for `twoslash` to resolve the types correctly

----

And later we need to revise code blocks across the docs to add the `twoslash` flag.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+# This is needed for TwoSlash to resolve the types correctly
+shamefully-hoist=true
--- a/.vitepress/config.ts
+++ b/.vitepress/config.ts
@@ -4,6 +4,7 @@ import { defineConfigWithTheme } from 'vitepress'
 import type { Config as ThemeConfig } from '@vue/theme'
 import baseConfig from '@vue/theme/config'
 import { headerPlugin } from './headerMdPlugin'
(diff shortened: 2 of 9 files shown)


antfu:
I think it's ready now. The only thing is that if we should bring Options API or wait for Floating Vue to adopt to Composition API only

NataliaTepluhina:
@antfu thank you for working on this! Looks great ✨ 

Could you please resolve the conflicts? I'll merge the PR and add `twoslash` to the codeblocks

brc-dd:
I'll release vitepress by tomorrow. It will have shiki@v1. We can remove resolutions from package.json here.

antfu:
Should be good to go now

yyx990803:
I think we need to be a bit selective on where to use these, let me take a pass on the rest of the docs so it doesn't become distractive.

PR #2670, opened 2024-01-16
docs: add explanation for scoped slots with render functions

dbzx10299:


NataliaTepluhina (approved):
@dbzx10299 thank you for the contribution!

PR #2673, opened 2024-01-18
Wrong comment in the description of the setup

GaetanoPiazzolla:
## Description of Problem
This comment should be referencing the setup() method of the OptionsAPI() 

## Proposed Solution
Changed comment.

## Additional Information


GaetanoPiazzolla:
Oh, all right. thank you for the explanation! I find it so confusing that Vue provides two ways of doing ALMOST the same thing.

PR #2678, opened 2024-01-22
Update event-handling.md to avoid horizontal scrolling

leoelz:
comment lines are being splitted to avoid horizontal scrolling as in a comment right after

## Description of Problem
ScreenCap horizontal scrolling
the comment is in one line and right after there is another comment in 3 lines that avoid horizontal scrolling

## Proposed Solution
Split the comment in two lines

## Additional Information


bencodezen (approved):
I definitely agree that being consistent is the better approach.

Thanks for the PR!

Issue #2688, opened 2024-02-07
typo in tutorial - step 2


thanosolo:
As I was going through the tutorial I noticed a sentence that seems incorrect:

https://vuejs.org/tutorial/#step-2

`State that can trigger updates when changed are considered reactive.`

This should either be `States that can trigger updates when changed are considered reactive.` or `State that can trigger updates when changed is considered reactive.`

NataliaTepluhina:
Closing an issue as the related PR was merged

PR #2689, opened 2024-02-07
Fix: typo of step 2

thanosolo:
Changed "are considered" to "is considered".

## Description of Problem
Description says `State that can trigger updates when changed are considered reactive`
## Proposed Solution
Description should say `State that can trigger updates when changed is considered reactive`
## Additional Information
An alternative would be `States that can trigger updates when changed are considered reactive`. However, in context, I feel like my proposed solution is more appropriate.

## Related Issue (Open)
https://github.com/vuejs/docs/issues/2688


bencodezen (approved):

bencodezen:
Thanks for the PR @thanosolo!

PR #2703, opened 2024-02-19
fix: Script Setup Does Not Populate Name Property

incutonez:
#[10348](https://github.com/vuejs/core/issues/10348):
- Adding clarification on how to access `name` property in script setup

## Description of Problem

In `script setup`, the `name` property does not get created, per [here](https://github.com/vuejs/core/issues/10348#issuecomment-1951604644).

## Proposed Solution

In order to help clarify what can be used, it makes sense to add a reference to `__name` to help devs that need access to the component's name in `script setup`.

## Additional Information

[Original PR](https://github.com/vuejs/core/commit/9734b31c312244a2b5c5cf83c75d7b34076a0c4b) that introduced `__name`

### DIFF
--- a/src/api/options-misc.md
+++ b/src/api/options-misc.md
@@ -29,7 +29,7 @@ Explicitly declare a display name for the component.
   There is one case where `name` is explicitly necessary: when matching against cacheable components in [`<KeepAlive>`](/guide/built-ins/keep-alive) via its `include / exclude` props.
 
   :::tip
-  Since version 3.2.34, a single-file component using `<script setup>` will automatically infer its `name` option based on the filename, removing the need to manually declare the name even when used with `<KeepAlive>`.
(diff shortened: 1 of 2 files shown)


yyx990803:
I don't think this should be added because the inference is an internal mechanism. Can you clarify your use case here?

incutonez:
> I don't think this should be added because the inference is an internal mechanism. Can you clarify your use case here?

Absolutely!  I use script setup, and I need to be able to reference the name of the component programmatically... according to the docs, I should be able to use `name`, but I can't because script setup sets `__name`, and it wasn't clear which property to use.

flogehring:
> I use script setup, and I need to be able to reference the name of the component programmatically...

What are you doing with the name?

incutonez:
> What are you doing with the name?

For display purposes, retrieving it for an `include`/`exclude` list in `KeepAlive`, and other use cases that apply for the Options API.

lakkvak:
I have the same problem. Is there a solution?

softboy99:
I have the same problem. Is there a solution? how about dynamic name relation with route.name? we still use vite-plugin-vue-setup-extend-plus. But which cannot deal with dynamic route related name

PR #2709, opened 2024-02-23
docs: proposal for two new immutable methods for Vue's list rendering

dev-itsheng:
## Additional Information

JavaScript provides two new immutable methods on arrays to solve the problem of legacy functions with side effects. These two methods also apply to Vue's list rendering optimizations as well, so I've modified the documentation to add suggestions for them.

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -426,9 +426,10 @@ methods: {
 </ul>
 ```
 
-Be careful with `reverse()` and `sort()` in a computed property! These two methods will mutate the original array, which should be avoided in computed getters. Create a copy of the original array before calling these methods:
+Be careful with `reverse()` and `sort()` in a computed property! These two methods will mutate the original array, which should be avoided in computed getters. Create a copy of the original array before calling these methods, alternatively, you can use their corresponding immutable methods, [`toReversed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) and [`toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted):
 
 ```diff
 - return numbers.reverse()
 + return [...numbers].reverse()
++ return numbers.toReversed()
 ```


brc-dd:
Note that these are ES2023 features not available in over 10% browsers and require manually configuring tsconfig.

dev-itsheng:
> Note that these are ES2023 features not available in over 10% browsers and require manually configuring tsconfig.

I have three considerations.

1. most users will be using the latest version of Chrome and they will get the expected results when learning the code in the documentation.

2. If the user is using TypeScript in their project, they will configure tsconfig, and babel similarly.

3. give users the necessary hints that these two new APIs can solve the problem of array side effects.

AloisSeckar:
> Note that these are ES2023 features not available in over 10% browsers and require manually configuring tsconfig.

Although I agree that it should at least come with a comment and warning about potential issues, I am not sure you're right in the part about requiring manual configuration.

I just scaffolded fresh Vue project starter by `npm create vue@latest` and I can use `Array.toReversed` without any further config. Target is set to `ESNext` in default Vue's tsconfig that is being extended.

brc-dd:
> Target is set to ESNext in default Vue's tsconfig that is being extended

Not in tsconfig.dom.json. You probably are trying inside some ts file. It should show error in sfc.

AloisSeckar:
> Not in tsconfig.dom.json.

This file is extending general `tsconfig.json` and not changing the `target` setting... 🤷

brc-dd:
Target doesn't matter compilerOptions.lib does.

In newly created vue app:

image

AloisSeckar:
@dev-itsheng 

> 2. if the user is using TypeScript in their project, they will have already learned to configure tsconfig, and babel will be similar.

This is a terribly wrong consideration, imo. Nowadays tools allow people using them without deeper understanding of the principles behind. I think it is generally a good thing, because it allows awesome content being created by almost everyone, but you should keep that in mind and not assuming others know what you do.

brc-dd:
> without complaining during build time

Yeah, the type-check command (vue-tsc) will throw error, build won't because vite doesn't check types.

src/guide/essentials/list.md
```diff
@@ -426,9 +426,10 @@ methods: {
 </ul>
 ```
 
-Be careful with `reverse()` and `sort()` in a computed property! These two methods will mutate the original array, which should be avoided in computed getters. Create a copy of the original array before calling these methods:
+Be careful with `reverse()` and `sort()` in a computed property! These two methods will mutate the original array, which should be avoided in computed getters. Create a copy of the original array before calling these methods, alternatively, you can use their corresponding immutable methods, [`toReversed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) and [`toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted):
```
bencodezen, line 429:
```suggestion
Be careful with `reverse()` and `sort()` in a computed property! These two methods will mutate the original array, which should be avoided in computed getters. Create a copy of the original array before calling these methods. Alternatively, you can use their corresponding immutable methods, [`toReversed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) and [`toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted):
```

dev-itsheng:
@brc-dd @AloisSeckar 

I think you are right, so I have added the new background of these two functions and solutions to common problems (editor  or browser) in the main body of the document and code comments.

PR #2711, opened 2024-02-23
fix typo in reactivity in depth section

dbzx10299:


dbzx10299:
I'm not sure what happened here, I created a new pull request so it should work now.

Issue #2694, opened 2024-02-10
Example in "Immutable Data" consists of function no longer callable


tfulanchan:
Concerning the example in [Immutable Data](https://vuejs.org/guide/extras/reactivity-in-depth.html#immutable-data) ,

```
//here
import produce from 'immer'
import { shallowRef } from 'vue'

export function useImmer(baseState) {
  const state = shallowRef(baseState)
  const update = (updater) => {
    state.value = produce(state.value, updater)
  }

  return [state, update]
}
```
 immer.js has undergone changes in its export patterns and the example in the Playground maybe outdated and requires attention.

![Screenshot 2024-02-10 062917](https://github.com/vuejs/docs/assets/60479975/16e25eea-d0af-45f5-b1a7-aa746922c1c1)
Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/immer.js?v=46fe6564' does not provide an export named 'default' (at immer.js:1:8)

PR #2719, opened 2024-03-01
Fix outdated Immer import in "Immutable Data"

ashc0:
## Description of Problem
  The "Immutable Data" section contains a code example that uses Immer to achieve data immutability. However, the code example uses the outdated Immer import method, `import produce from "immer"`. The latest version of Immer recommends using the named import, `import { produce } from "immer"`.
## Additional Information
[2694](https://github.com/vuejs/docs/issues/2694)


NataliaTepluhina (approved):
Thank you for the fix!

PR #2722, opened 2024-03-01
Update computed.md

SpartaSixZero:
The title should be Computed Properties instead of Computed Caching

## Description of Problem
The title is incorrect

## Proposed Solution
Fixed the title


NataliaTepluhina:
Thank you for your contribution! However, the `Computed caching` title is intended here

PR #2725, opened 2024-03-03
Update forms.md

suleiman-hamza:
Selecting multiple options for mouse users on their desktop

## Description of Problem
No description on how mouse users can choose multiple options from the <select> multiple
## Proposed Solution
I created a note describing how to select multiple options

### DIFF
--- a/src/guide/essentials/forms.md
+++ b/src/guide/essentials/forms.md
@@ -293,6 +293,9 @@ Multiple select (bound to array):
     <option>C</option>
   </select>
 </div>
+:::tip Note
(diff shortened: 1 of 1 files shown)


brc-dd:
1. That tip block doesn't get rendered properly. (You need to add a newline after closing div.)
2. This is not specific to vue, but rather how native select element works. I don't think we should start specifying [everything in MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#usage_notes) again in vue docs.

suleiman-hamza:
Yeah, it makes sense to keep Vue docs specific to Vue, Thanks ;)

PR #2735, opened 2024-03-05
docs: fix typo in reactivity in depth

dbzx10299:


brc-dd:
I don’t think it’s a typo. ‘automagically’ is a valid word which makes sense in that context:

> (especially in relation to the operation of a computer process) automatically and in a way that seems ingenious, inexplicable, or magical

dbzx10299:
Yes you are right, good catch.

PR #2746, opened 2024-03-12
Update config.ts to restore Style-Guide to nav

kheiner:
It appears the style-guide nav was removed two years ago in preparation for migration from v2 to v3 but never restored.

## Description of Problem

Style Guide was not easily searchable in docs.

## Proposed Solution

Restore Style-Guide path to nav


yyx990803:
Thanks, but this is currently intentional because the style guide is still out of date in many places and we haven't gotten to fully update it for Composition API yet.

kheiner:
I'll check it out and see if I'm able to contribute but I do see updates in #2585.

PR #2749, opened 2024-03-13
docs: fix typo - count should be model

EtherNoteWorld:
## Description of Problem
When I read the documentation, I am unclear about the source of this count.

### DIFF
--- a/src/guide/components/v-model.md
+++ b/src/guide/components/v-model.md
@@ -27,7 +27,7 @@ The parent can then bind a value with `v-model`:
 
 ```vue-html
 <!-- Parent.vue -->
-<Child v-model="count" />
+<Child v-model="model" />
 ```
 
 The value returned by `defineModel()` is a ref. It can be accessed and mutated like any other ref, except that it acts as a two-way binding between a parent value and a local one:
--- a/src/guide/components/v-model.md
+++ b/src/guide/components/v-model.md
@@ -27,7 +27,7 @@ The parent can then bind a value with `v-model`:
 
 ```vue-html
 <!-- Parent.vue -->
-<Child v-model="model" />
+<Child v-model="countModel" />
 ```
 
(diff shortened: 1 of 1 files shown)


brc-dd (approved):

brc-dd:
Actually, the name doesn't matter in parent component. It can be `count`, `foo`, `bar` or anything else. This change might be bit confusing for new users as they might think that the variable name should be `model` too. I'm fine with either one though.

AloisSeckar:
I object against this chage as it goes against clean code principles. Naming the variable `model` makes sense inside the component itself, because there it has clear purpose - it _"models the actual data value for this component"_. But from parent's perspective you want to know, what data you are passing to your child, what they represent. Therefore `count` makes sense, if you are displaying, let's say amount items in shoping cart, but `model` is nothing. If you use two child components, will you name your variables `model1` and `model2`? How about having ten children?

But maybe we should make clearer, why `count` is used. I will think about some wording, if noone else does.

bencodezen:
Agreed with @AloisSeckar. I'll propose some changes in the PR to accommodate the general momentum we seem to be heading on.

bencodezen (changes requested):
Naming the model `model` can lead to future confusion. Suggesting changes shortly.

src/guide/components/v-model.md
```diff
@@ -27,7 +27,7 @@ The parent can then bind a value with `v-model`:
 
 ```vue-html
 <!-- Parent.vue -->
-<Child v-model="count" />
+<Child v-model="model" />
```
bencodezen, line 30:
```suggestion
<Child v-model="countModel" />
```

bencodezen: @EtherNoteWorld Can you update the PR to update the variables so that it reflects `countModel` rather than just `model`?

NataliaTepluhina (approved):

PR #2750, opened 2024-03-14
docs: fix update theme property to a reactive property

GmavionR:
Update theme to reactive  property

## Description of Problem

Theme should be a reactive property before you can use v-bind in css.

### DIFF
--- a/src/api/sfc-css-features.md
+++ b/src/api/sfc-css-features.md
@@ -183,9 +183,10 @@ The syntax works with [`<script setup>`](./sfc-script-setup), and supports JavaS
 
 ```vue
 <script setup>
-const theme = {
-  color: 'red'
-}
+import { ref, reactive } from 'vue'
+const theme = reactive({
+    color: 'red',
+})
 </script>
 
 <template>


AloisSeckar:
Hello, and thank you for your interest in improving Vue.js docs. But your assumption is wrong. Static values can be bound easilly - see [live demo here](https://play.vuejs.org/#eNp9UT1PwzAQ/SuWl4BUkgGmECoB6gADIGD0EpwjSXFsy76EoCr/nbOjlCJV3Xzv4/yeveO31qZDDzznhZeutcg8YG/XQkujPTJsoAN2w3ZCMyaNMi5niYMqEXoSushmE8lpQOisKhFoYqyw6waUMkUWlhXZAUmjxx8FzEtjoSLE/ts/XHy0ujpL4t1pBJPz6+XC4CQLX3Ekv/5s63TrjaYGcYfg0nS2VeCeLbbUQfB83h64khJ9P0YMXQ+rBZcNyK8j+NaPARP8xYEHN4Dgew5LVwPO9ObtCUY678nOVL0i9QnyFbxRfcg4y+56XVHsA11M+9BZ47DV9bvfjAjaL6VC0KCcol5w+sb7E9X/4l6mV9FHL8qnX4VEq+4=).

Using `reactive` would become useful to allow dynamic runtime changes of the bound color. But this is not what this simple demo is focusing at, so I think it would be rather confusing than helpful to include it here.

brc-dd:
I think having reactive will be more consistent with the options API example above it, and the description below that states that the styles are reactively updated if the source value changes.  We can probably make the example more useful than just keeping it simple because I don't think any one would use CSS data binding for a static value.

PS: remove extra `ref` import

src/api/sfc-css-features.md
```diff
@@ -183,9 +183,10 @@ The syntax works with [`<script setup>`](./sfc-script-setup), and supports JavaS
 
 ```vue
 <script setup>
-const theme = {
-  color: 'red'
-}
+import { ref, reactive } from 'vue'
```
bencodezen, line 186:
```suggestion
import { ref } from 'vue'
```

src/api/sfc-css-features.md
```diff
@@ -183,9 +183,10 @@ The syntax works with [`<script setup>`](./sfc-script-setup), and supports JavaS
...
+const theme = reactive({
```
bencodezen, line 187:
```suggestion
const theme = ref({
```

bencodezen (approved):
Great catch @GmavionR! I modified the example you submitted down to only import `ref` and use that instead of `reactive` since this is the one we'd like to see the community standardize on if possible.

Thank you for the PR!

PR #2751, opened 2024-03-15
docs: added conditional slots section in slots guide

MrHus:
## Description of Problem

The guide section of slots does not show how to do conditional slots.

## Proposed Solution

I've added a section on how to use slots conditionally with a card
example and a link to a playground.

Here you can see it in action:

https://deploy-preview-2751--vuejs.netlify.app/guide/components/slots.html#conditional-slots

### DIFF
--- a/src/guide/components/slots.md
+++ b/src/guide/components/slots.md
@@ -296,6 +296,35 @@ function BaseLayout(slots) {
 }
 ```
 
+## Conditional Slots
+
+Sometimes you want to render something based on whether or not a slot is present. 
+
+You can use the [$slots](https://vuejs.org/api/component-instance.html#slots) property in combination with a [v-if](https://vuejs.org/guide/essentials/conditional.html#v-if) to achieve this.
+
+In the example below we define a Card component with two conditional slots: `header` and `footer`.
+When the header / footer is present we want wrap them to provide additional styling:
+
+```vue-html
+<template>
+  <div class="card">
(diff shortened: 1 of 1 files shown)


src/guide/components/slots.md
```diff
@@ -296,6 +296,36 @@ function BaseLayout(slots) {
 }
 ```
 
+## Conditional Slots
+
+Sometimes you want to render something based on whether or not a slot is present. 
+
+You can use the [$slots](https://vuejs.org/api/component-instance.html#slots) component
+instance in combination with a [v-if](https://vuejs.org/guide/essentials/conditional.html#v-if) to achieve this.
```
NataliaTepluhina, line 304: **suggestion**: I believe this requires a little rephrasing

```suggestion
You can use the [$slots](https://vuejs.org/api/component-instance.html#slots) property in combination with a [v-if](https://vuejs.org/guide/essentials/conditional.html#v-if) to achieve this.
```

NataliaTepluhina (changes requested):

src/guide/components/slots.md
MrHus: I've changed It to your recommendation!

MrHus:
@NataliaTepluhina is there anything more you'd like me to change?

NataliaTepluhina:
@MrHus Thank you for the fixes! The phrasing looks good to me now 👍🏻

NataliaTepluhina (approved):

PR #2768, opened 2024-03-25
Update reactivity-in-depth.md

vostersc:
Replace the word mutation with the word reassign so the example is accurate.

## Description of Problem
The example used in the beginning of the Reactivity docs isn't mutating. Primitives aren't mutated in JavaScript. The action in the example is making a reassignment.

## Proposed Solution
Replace mutate with the correct word. Replace the other instance of mutation with a word that could mean both reassign and mutate.

## Additional Information
This problem likely occurs in other places.

### DIFF
--- a/src/guide/extras/reactivity-in-depth.md
+++ b/src/guide/extras/reactivity-in-depth.md
@@ -31,7 +31,7 @@ A0 = 2
 console.log(A2) // Still 3
 ```
 
-When we mutate `A0`, `A2` does not change automatically.
+When we reassign `A0`, `A2` does not change automatically.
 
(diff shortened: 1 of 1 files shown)


AloisSeckar:
I am not strongly against, but the term "mutate" is derived from "mutable / immutable", which is commonly used in programming to determine whether you can or cannot change the object. So I don't think it is really necessary to change it.

vostersc:
I get that! It just seems like precision would be especially important in a discussion like this. In JavaScript they aren't used interchangeably, though I am sure there is a language where it is. If the terms were interchangeable in JavaScript it would confuse developers who are new to the language. Because of this, you'll notice the distinction between mutation and assignment maintained in the React docs. That being said I totally understand if making these changes isn't a priority.

PR #2784, opened 2024-04-04
Update introduction.md

HP8585:
you missed the <script> and <template> tags which can be misleading for new coder who are gonna start learning vue but will be puzzled.. please make it right!

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -32,6 +32,7 @@ Here is a minimal example:
 <div class="options-api">
 
 ```js
+<script>
 import { createApp } from 'vue'
 
 createApp({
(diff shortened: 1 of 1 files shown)


AloisSeckar:
Hi and thanks for suggestion. Assuming SFC syntax, your proposal would make sense. But Single-File Components are only one (though most common) way of using Vue. You can also have `in-DOM templates` directly in HTML and your logic can be placed in standalone `.js` file. Because of that is seems more logical to only annotate the codeblocks with `js` and `template` like it is, not to force newcommer into SFC syntax immideately. It is introduced very soon after.

src/guide/introduction.md
```diff
@@ -32,6 +32,7 @@ Here is a minimal example:
 <div class="options-api">
 
 ```js
+<script>
```
NataliaTepluhina, line 35: Thank you for your suggestion! However, we cannot simply put `<script>` tag in `js` code block. We'd need to have `vue-html` block (we can merge it with the next one) and then put the tag there:

```vue-html
<script>
 // script part
 </script>

<template>
  // template part
</template>

NataliaTepluhina (changes requested):

PR #2785, opened 2024-04-04
feat(Suspense): add suspensible prop information

ferferga:
## Description of Problem

``suspensible`` prop was added in https://github.com/vuejs/core/pull/6736 in Vue 3.3 but there's no mention of it in the documentation. I tried to simplify a bit further the description of the original PR (which is a great explanation and I didn't think it needed much changing besides maybe a few nitpicks which might be more understandable by the uninitiated).


NataliaTepluhina (approved):
Thank you for the great addition to docs!

PR #2792, opened 2024-04-08
fix: don't trigger onHashChange if path is not the same

### LINKED ISSUE #2791: can't navigate to other url
when in tutorial page(https://vuejs.org/tutorial/#step-1 page), I can't navigate to other page

brc-dd:
closes #2791

### DIFF
--- a/src/examples/utils.ts
+++ b/src/examples/utils.ts
@@ -208,8 +208,15 @@ export function resolveNoBuildExample(
 }
 
 export function onHashChange(cb: () => void) {
-  window.addEventListener('hashchange', cb)
+  if (typeof document === 'undefined') return
+  const { pathname } = window.location
+  function check() {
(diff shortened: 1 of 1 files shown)


src/examples/utils.ts
```diff
@@ -208,8 +208,15 @@ export function resolveNoBuildExample(
 }
 
 export function onHashChange(cb: () => void) {
-  window.addEventListener('hashchange', cb)
+  if (typeof document === 'undefined') return
+  const { pathname } = window.location
+  function handler() {
+    if (window.location.pathname === pathname) {
```
brc-dd, line 214: Technically we could have also use .newURL / .oldURL on event argument, but looks like vitepress is dispatching only generic event object for hashchange. I'll fix it there and refactor this later.

PR #2798, opened 2024-04-15
Vuejsde conf banner september 16 22 2024

antony-k1208:
Thanks again!


.vitepress/config.ts
```diff
@@ -626,13 +626,6 @@ export default defineConfigWithTheme<ThemeConfig>({
         defer: ''
       }
     ],
-    [
```
yyx990803, line 629: no need to remove this script here - the banner injection is controlled externally and will be off during that period

yyx990803:
I can't see the banner in the preview - did you enable the banner component in `theme/index.ts`?

PR #2801, opened 2024-04-16
feat: update isProxy() param type

OnlyWick:


src/api/reactivity-utilities.md
```diff
@@ -237,7 +237,7 @@ Checks if an object is a proxy created by [`reactive()`](./reactivity-core#react
 - **Type**
 
   ```ts
-  function isProxy(value: unknown): boolean
+  function isProxy(value: any): boolean
```
NataliaTepluhina, line 240: Could you please provide more context on why `any` here is preferrable to `unknown`?

OnlyWick: In version `3.4.22`, I have changed it to `any`.

https://github.com/vuejs/core/blob/b4b856b3f0ebe937eacb6945d13b2561037057b5/packages/reactivity/src/reactive.ts#L332

sxzz (approved):

src/api/reactivity-utilities.md
brc-dd: Why was it changed in core though?

OnlyWick: avoid assertion

brc-dd: All of the other three methods there use assertion :/

OnlyWick: In my impression, the refactoring has already changed the style in minor, I didn't expect it to not change yet. Let's unify them then.

https://github.com/vuejs/core/blob/bb5c31e61480bae971df6817d4a65a511bd8fa90/packages/reactivity/src/ref.ts#L42

OnlyWick: In my impression, the refactoring has already changed the style in minor, I didn't expect it to not change yet.

https://github.com/vuejs/core/blob/bb5c31e61480bae971df6817d4a65a511bd8fa90/packages/reactivity/src/ref.ts#L42

OnlyWick: I feel that avoiding assertions altogether is feasible, more concise, and the results are the same

PR #2802, opened 2024-04-16
chore: remove line breaks in sentences

jay-es:
## Description of Problem
Line breaks in sentences make translation into other languages difficult.


bencodezen (approved):
Agreed. Thanks for the PR!

PR #2804, opened 2024-04-16
Update list.md - description of iterating Objects with v-for

destradafilm:
Update list.md to reflect a better description of result for iterating Objects in v-for

`iterating v-for="value in myObject" is based on the result of calling Object.values()`

## Description of Problem
Minor nit - just reading the documentation and see that iterating an Object with `v-for` is more like `Object.values` than `Object.keys` 

## Proposed Solution
switch `Object.keys()` for `Object.values()`

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -134,7 +134,7 @@ You can also use `of` as the delimiter instead of `in`, so that it is closer to 
 
 ## `v-for` with an Object {#v-for-with-an-object}
 
-You can also use `v-for` to iterate through the properties of an object. The iteration order will be based on the result of calling `Object.keys()` on the object:
+You can also use `v-for` to iterate through the properties of an object. The iteration order will be based on the result of calling `Object.values()` on the object:
 
 <div class="composition-api">
 


AloisSeckar:
`Object.keys` and `Object.values` both return the same order of items on the same object. I think it is better to keep `keys` in the docs, because when you talk about "properties of an object", you would typically want to know their names (keys) rather than plain values without a context of their meaning...

src/guide/essentials/list.md
```diff
@@ -134,7 +134,7 @@ You can also use `of` as the delimiter instead of `in`, so that it is closer to
 
 ## `v-for` with an Object {#v-for-with-an-object}
 
-You can also use `v-for` to iterate through the properties of an object. The iteration order will be based on the result of calling `Object.keys()` on the object:
+You can also use `v-for` to iterate through the properties of an object. The iteration order will be based on the result of calling `Object.values()` on the object:
```
brc-dd, line 137: The keyword here is "iteration order" and as [AloisSeckar](https://github.com/AloisSeckar) wrote above, both return elements in same order. What you get in `v-for="_____ in obj"` is different thing and is shown by the example just below this. Also, `Object.keys()` is the thing we internally use.

destradafilm: ok - makes sense! Thanks for explaining and Thanks for having such great documentation!

PR #2806, opened 2024-04-18
Update slots.md

claushellsing:
Update documentation in order to indicate how to use conditional render of default slot


brc-dd:
Can you provide more context on why is this needed? Isn't it obvious from the example above that, that `$slots.foo` is needed to check for `<slot name="foo" />`? Regarding the name when name is not explicitly specified, the docs mention:

> A `<slot>` outlet without `name` implicitly has the name "default".

claushellsing:
You're right. According to the documentation, one can interpolate knowledge and assume a `$slots.default` exists to check if the slot is present. However personally, I tend to prefer explicitness. It's just a personal preference.

brc-dd:
Instead of two examples, can this be updated in the above example itself?

claushellsing:
Like this ?

In the example below we define a Card component with three conditional slots: `header`, `footer` and the `default` one.
When the header / footer / default is present we want wrap them to provide additional styling:

```vue-html
<template>
  <div class="card">
    <div v-if="$slots.header" class="card-header">
      <slot name="header" />
    </div>
    
    <div v-if="$slots.default" class="card-content">
      <slot />
    </div>
    
    <div v-if="$slots.footer" class="card-footer">
      <slot name="footer" />
    </div>
  </div>
</template>
```

brc-dd:
Yeah, looks fine to me.

NataliaTepluhina (approved):

src/guide/components/slots.md
```diff
@@ -302,8 +302,8 @@ Sometimes you want to render something based on whether or not a slot is present
 
 You can use the [$slots](/api/component-instance.html#slots) property in combination with a [v-if](/guide/essentials/conditional.html#v-if) to achieve this.
 
-In the example below we define a Card component with two conditional slots: `header` and `footer`.
-When the header / footer is present we want to wrap them to provide additional styling:
+In the example below we define a Card component with three conditional slots: `header`, `footer` and the `default` one.
+When the header / footer / default is present we want wrap them to provide additional styling:
```
NataliaTepluhina, line 306:
```suggestion
When the header / footer / default is present we want to wrap them to provide additional styling:
```

Issue #2528, opened 2023-10-09
Define "getter function"


nielsbom:
In the [docs](https://github.com/search?q=repo%3Avuejs%2Fdocs+%22getter+function%22&type=code) the term "getter function" seems to be an important concept but it's not defined anywhere that I've found.

Do y'all think it's an idea to define it?

bencodezen:
@nielsbom This is technically a fundamental JS concept, but I could see value in linking to docs since this is a more advanced concept for those new to JavaScript.

I'll issue a PR to help improve this.

nielsbom:
Ah, yeah I don't use those parts of JavaScript directly often. Saying something like "getter functions are a feature of JavaScript itself, see the docs" would already help.

PR #2817, opened 2024-04-23
docs (#2528): add references to help people new to getter functions i…

bencodezen:
## Description of Problem

Getter functions are referenced a lot in docs, but if you're new to JS, it is unclear whether this is a Vue thing or a JS thing.

## Proposed Solution

Similar to how acronyms are defined at the beginning of each article before being used throughout the rest, adding a link referencing to what getters / setters are can help provide clarity to people newer to JavaScript and haven't encountered these concepts yet.

## Additional Information

Related issue #2528

### DIFF
--- a/src/api/component-instance.md
+++ b/src/api/component-instance.md
@@ -212,7 +212,7 @@ Imperative API for creating watchers.
 
 - **Details**
 
-  The first argument is the watch source. It can be a component property name string, a simple dot-delimited path string, or a getter function.
+  The first argument is the watch source. It can be a component property name string, a simple dot-delimited path string, or a [getter function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get).
 
   The second argument is the callback function. The callback receives the new value and the old value of the watched source.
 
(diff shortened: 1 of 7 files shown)


brc-dd:
I don't think getter function in Vue's context is same as `get` syntax. It's more like any function that returns something. 

That MDN link just shows things like `const foo = { get bar() { return 1 }`. If one is just reading that link and will try to apply that to, say, `computed` it won't work.

AloisSeckar:
> I don't think getter function in Vue's context is same as `get` syntax.

Then perhaps a better approach would be to include it into [Glossary](https://vuejs.org/glossary/) under its own label? 🤔

bencodezen:
> I don't think getter function in Vue's context is same as `get` syntax. It's more like any function that returns something.
> 
> That MDN link just shows things like `const foo = { get bar() { return 1 } }`. If one is reading that link and will try to apply that to, say, `computed` it won't work.

Further down in the description, there's the following snippet:

> The getter syntax allows you to specify the getter function in an object initializer.

I believe this is the concept that we are referring to, though maybe I'm wrong.

NataliaTepluhina (approved):

nielsbom:
W00t!

PR #2822, opened 2024-04-25
Update class-and-style.md

anoopbk123:
In the inline style are it was written 13 px instead of 30px for the font size. I re write it to 30px


anoopbk123:
In the inline style are it was written 13 px instead of 30px for the font size. I re write it to 30px

brc-dd:
change the value in `options-api` block below that example too.

bencodezen (approved):
Fair point on the consistency. Thanks for the PR!

PR #2825, opened 2024-04-25
fix: <pre> syntax error in example code

xiaodong2008:
## Description of Problem

`<pre>` should not put inside `<p>`

References:
https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element
https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element
https://github.com/prettier/prettier/issues/15294

### DIFF
--- a/src/examples/src/form-bindings/App/template.html
+++ b/src/examples/src/form-bindings/App/template.html
@@ -16,7 +16,7 @@
 <label for="john">John</label>
 <input type="checkbox" id="mike" value="Mike" v-model="checkedNames">
 <label for="mike">Mike</label>
-<p>Checked names: <span>{{ checkedNames }}</span></p>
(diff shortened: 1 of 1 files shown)


src/examples/src/form-bindings/App/template.html
```diff
@@ -16,7 +16,8 @@ <h2>Multi Checkbox</h2>
 <label for="john">John</label>
 <input type="checkbox" id="mike" value="Mike" v-model="checkedNames">
 <label for="mike">Mike</label>
-<p>Checked names: <pre>{{ checkedNames }}</pre></p>
```
brc-dd, line 19: maybe just use `<span>` like other examples below this? 👀

xiaodong2008: You're right, putting this in one line with `<span>` is better. 😆

Issue #2821, opened 2024-04-24
Docs: Code blocks writing standards


xiaodong2008:
Refers to the pages in docs, I think that the first letter of text(like `button`) in code blocks should be capitalised.

For example, there is a violation of writing standards in [`/guide/components/attrs`](https://github.com/vuejs/docs/blob/main/src/guide/components/attrs.md?plain=1#L29), it should be write like this:

```diff
- <button class="large">click me</button>
+ <button class="large">Click Me</button>
```

bencodezen:
I think you're totally right @xiaodong2008! Appreciate the call out.

xiaodong2008:
Sure! I will fix this soon 🥰

PR #2828, opened 2024-04-26
docs #2821: fix writing mistakes in letter case

xiaodong2008:
## Description of Problem

Some letters should be capitalised, check below:

```diff
+ <button>Click Me</button>
+ <h1>Title</h1>
+ <p>This is a paragraph or sentence</p>
+ <p>Count: 0</p>
- <p>short text/word</p> // no need to capitalised
```

### DIFF
--- a/src/guide/introduction.md
+++ b/src/guide/introduction.md
@@ -63,7 +63,7 @@ createApp({
 ```vue-html
 <div id="app">
   <button @click="count++">
-    Count is: {{ count }}
+    Count: {{ count }}
   </button>
 </div>
 ```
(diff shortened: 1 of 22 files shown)


src/guide/introduction.md
```diff
@@ -128,7 +128,7 @@ export default {
 </script>
 
 <template>
-  <button @click="count++">Count is: {{ count }}</button>
+  <button @click="count++">Count: {{ count }}</button>
```
bencodezen, line 131: I understand why you're removing the `is`, but for the sake of the semantics of being explicit, let's not change this.

bencodezen (changes requested):
Thanks for the PR @xiaodong2008! I agree with most of the casing changes, but removing the `is` in the examples isn't related to letter casing and arguably is better as is since it is more explicit as a statement. 

Once you remove those changes, I'm happy to approve and merge these changes!

PR #2837, opened 2024-04-27
fix: wrong remark in guide step-11

xiaodong2008:
## Description of Problem

It should be `register child component`, not `register component`.

Also, refer to [here](https://github.com/vuejs/docs/blob/main/src/tutorial/src/step-11/App/options.js).


bencodezen (approved):
Thank you for the PR!

Issue #2810, opened 2024-04-19
Doc for `update:modelValue` or `update:*` event ?


steelywing:
> https://vuejs.org/guide/components/v-model.html#under-the-hood

`defineModel` will convert to an event named `update:modelValue`, I search the doc site but I cannot found doc for `update:modelValue` or `update:*` event, how the parent handle this event?

I think it will make the parent update the `modelValue` to the emit value, right?

brc-dd:
`v-model="foo"` is compiled to `:modelValue="foo" @update:modelValue="$event => (foo = $event)"`

steelywing:
> `v-model="foo"` is compiled to `:modelValue="foo" @update:modelValue="$event => (foo = $event)"`

Thanks for the quick reply, can we have a doc for this?

brc-dd:
IIRC it used to be there but I can't find it anymore.

NataliaTepluhina:
Closing as related PR was merged

PR #2839, opened 2024-04-29
docs: add docs for v-model `update:*` event

xiaodong2008:
Added #2810

### DIFF
--- a/src/guide/components/v-model.md
+++ b/src/guide/components/v-model.md
@@ -59,6 +59,7 @@ const model = defineModel()
 This is how you would implement the same child component shown above prior to 3.4:
 
 ```vue
+<!-- Child.vue -->
 <script setup>
 const props = defineProps(['modelValue'])
 const emit = defineEmits(['update:modelValue'])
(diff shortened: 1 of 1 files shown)


brc-dd:
Ah, wait this is documented, but it's currently only inside options API block (see later in the document). It should be shown on both composition and options API.

xiaodong2008:
> Ah, wait this is documented, but it's currently only inside options API block (see later in the document). It should be shown on both composition and options API. It should be similar to what it was earlier - https://github.com/vuejs/docs/blob/7f299d98443efc489115d5f6b64d02094660dbd0/src/guide/components/v-model.md?plain=1 .

Okay, I will update this.

xiaodong2008:
Wait, but composition API already have an `under the hood` part(not complete), and it's not similar to options API, I just completed the part of the parent component, so I think this change is acceptable

https://vuejs.org/guide/components/v-model#under-the-hood

NataliaTepluhina (approved):

PR #2845, opened 2024-04-30
fix: remove layout shift on the cute logo

prashantpalikhe:
## Description of Problem

Because there is not dimension specified for the logo, we get a massive layout shift 

## Proposed Solution

Add the aspect ratio to the cute logo's style


yyx990803:
Unfortunately this doesn't really remove the layout shift because the v-if is only toggled after mount / hydration.

This is because we want to avoid sending a request for the image if the URL does not have the `?uwu` query string, so it cannot be rendered during SSG.

I'll think of another way to get around this.

PR #2846, opened 2024-05-01
Fix: uwu logo style & alt text from Home.vue

yamanoku:
## Description of Problem

- Insert alternative text for the logo itself (because **this is a meaningful logo image**, not a background image)
- To prevent layout shift for the image, specify the `width` and `height`
- The `:global` specification is meaningless since CSS Modules are not being used


yyx990803:
FYI the original implementation was intentional because we don't want to load the image when the users didn't explicitly have `?uwu` in the URL...

PR #2852, opened 2024-05-03
Added a missing article

thenewdesign:
For grammatical correctness

## Description of Problem
Missing word

## Proposed Solution
Added word


NataliaTepluhina:
I don't think `Composition API` requires `the` article - at least, we're using it throughout the documentation without `the`.

Issue #2851, opened 2024-05-03
Add documentation for "VNode Lifecycle Events"


atilkan:
Hi,
"VNode Lifecycle Events" are documented in [migration docs](https://v3-migration.vuejs.org/breaking-changes/vnode-lifecycle-events#vnode-lifecycle-events) but there is no information about them in main docs. 

Can you add some information about them?

atilkan:
Duplicate issue => https://github.com/vuejs/core/discussions/9298

NataliaTepluhina:
As mentioned in the PR, we didn't document this feature in both Vue 2 and Vue 3 because it's not a recommended practice to listen to `@vue:` events on child components.

ngocsangyem:
Hi @NataliaTepluhina, please tell us why this is not a recommended practice. This approach is beneficial in many situations. Are you guys planning to remove it in breaking changes?

AloisSeckar:
I guess the reason is that it tightly couples child to parent and therefore reducing re-usability.

It is also obsfucating logical flow of the code - because why wouldn't you handle component's lifecycle hook inside the component itself?

atilkan:
> I guess the reason is that it tightly couples child to parent and therefore reducing re-usability.
> 
> It is also obsfucating logical flow of the code - because why wouldn't you handle component's lifecycle hook inside the component itself?

If someone uses it internally, that answers the question of why. 
Why is it a thing anyway?

PR #2860, opened 2024-05-08
docs: listening to lifecycle events

AloisSeckar:
## Description of Problem

This should be at least a partial response to #2851 documenting the built-in possibility to listen to Vue lifecycle events without explicitly declare corresponding emits. The feature is briefly mentioned in [Migration guide](https://v3-migration.vuejs.org/breaking-changes/vnode-lifecycle-events#vnode-lifecycle-events) but AFAIK not documented anywhere in the actual Vue 3 docs.

## Proposed Solution

I described a simple demo including Playground examples.

## Additional Information

I am happy to receive a revisions and I am ready to address any remarks as there can be more to cover on this topic I am not aware of yet.


NataliaTepluhina:
Thank you for creating the PR! However, we didn't document this intentionally since this is not a recommended practice to listen to `@vue:` events on the child components

PR #2862, opened 2024-05-09
Update template-syntax.md

HMZeeshanLatif:
Add style attribute in Dynamically Binding Multiple Attributes


NataliaTepluhina (approved):

AloisSeckar:
When this was added, shouldn't it be included in Options API example as well? Or is there special reason, why it was only included into Compositions API example?

PR #2873, opened 2024-05-14
api(built-in-directives): enhance-v-on-description

quiteeasy:
docs: Explain $event behavior with inline handlers

- Add a description of how the $event object is automatically set when using inline event handlers with v-on
- Clarify the need to explicitly pass $event when emitting events from child components to preserve the original event data
- Provide examples to illustrate the correct usage of $event in these scenarios

Related issue: https://github.com/vuejs/core/issues/10945

### DIFF
--- a/src/api/built-in-directives.md
+++ b/src/api/built-in-directives.md
@@ -198,6 +198,10 @@ Attach an event listener to the element.
 
   When listening to native DOM events, the method receives the native event as the only argument. If using inline statement, the statement has access to the special `$event` property: `v-on:click="handle('ok', $event)"`.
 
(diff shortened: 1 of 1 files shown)


src/api/built-in-directives.md
```diff
@@ -198,6 +198,10 @@ Attach an event listener to the element.
 
   When listening to native DOM events, the method receives the native event as the only argument. If using inline statement, the statement has access to the special `$event` property: `v-on:click="handle('ok', $event)"`.
 
+  When using an inline statement with the `$event` property, such as `v-on:click="handle('ok', $event)"`, the inline statement is transformed to `$event => handle('ok', $event)`. This means that the `$event` object is automatically passed as the first argument to the inline function.
```
NataliaTepluhina, line 201: We're trying to follow `show, not tell` principle in the docs, and the example below shows the behaviour you're explaining here:

```
 <button v-on:click="doThat('hello', $event)"></button>
```

I am not sure any additional clarification is needed here

PR #2878, opened 2024-05-20
chore: bump deps

brc-dd:
reverts #2858


skirtles-code:
It seems this change may have broken some pages in the docs. e.g. https://vuejs.org/api/sfc-script-setup.html currently won't load.

brc-dd:
I have partially reverted the changes and locked vitepress version back to 1.1.

.prettierignore
```diff
@@ -0,0 +1 @@
+*.vue
```
Jinjiang, line 1: Any special reason we added this? (sorry for being late 😅)

package.json
```diff
@@ -7,33 +7,24 @@
...
-      "ignoreMissing": [
-        "@algolia/client-search",
-        "react",
-        "react-dom",
-        "@types/react",
-        "search-insights"
-      ]
-    }
-  }
+  "packageManager": "pnpm@9.1.1"
```
Jinjiang, line 29: @brc-dd do you think we can skim this rule? Since I guess it breaks the translation deployments on Vercel like docs-cs https://github.com/vuejs-translations/docs-cs/issues/132#issuecomment-2122103463

.prettierignore
brc-dd: They aren't formatted currently, and prettier doesn't work properly with vue's template part.

package.json
brc-dd: We can fix it on v8 for now. Vercel supports corepack but I think they are blocked on something in pnpm v9. On newer node versions, this is automatically added each time you run a command and have corepack enabled locally (say, for other projects). Also, the engine fields of most of the translation repos are wrong. vite, vitepress, pnpm don't support node v14.

brc-dd: Or I think v9 is also supported via corepack (but detecting it is not supported). Try setting https://vercel.com/docs/deployments/configure-a-build#corepack `ENABLE_EXPERIMENTAL_COREPACK=1` in env variable.

PR #2883, opened 2024-05-28
Update props.md

LimiteAether:
add playground for props type as a custom class


NataliaTepluhina:
Hi! Thank you for the contribution but I am not entirely convinced this particular chapter needs a playground, as we don't see anything "in action"

PR #2885, opened 2024-05-30
Update custom-directives.md

thiagosaife:
Removing `prevVnode` params from hooks where they aren't available.

## Description of Problem
According to the docs: `prevVnode: the VNode representing the bound element from the previous render. Only available in the beforeUpdate and updated hooks.`
However we have `prevVnode` as a parameter on all the other hooks.
## Proposed Solution
Removed `prevVnode` from hooks where they are not allowed to be used.


NataliaTepluhina (approved):
Thank you for the fix@

PR #2890, opened 2024-06-03
Update v-model.md

Yordan-Ramchev:
## Description of Problem

In the **Basic Usage** section, **function update()** is not invoked and is not helping the example.

``` 
function update() {
  model.value++
}
```

## Proposed Solution

> When it is mutated by the child, it causes the parent bound value to be updated as well.

Maybe adding a click event to invoke the update would help clarify the example.


src/guide/components/v-model.md
```diff
@@ -19,7 +19,7 @@ function update() {
 </script>
 
 <template>
-  <div>Parent bound v-model is: {{ model }}</div>
+  <div @click="update">Parent bound v-model is: {{ model }}</div>
```
NataliaTepluhina, line 22: **suggestion**: I agree with the addition, but let's make it a bit more semantic by adding a button and binding the `update` to it:

```suggestion
  <div>Parent bound v-model is: {{ model }}</div>
  <button @click="update">Increment</button>
```

NataliaTepluhina (changes requested):

src/guide/components/v-model.md
Yordan-Ramchev: Yeah, that makes total sense :)

NataliaTepluhina: Thank you for the speedy fix! :)

NataliaTepluhina (approved):

PR #2895, opened 2024-06-14
docs: fix missing model modifier

Disservin:
## Description of Problem

The example is missing the `.capitalize` modifier.


NataliaTepluhina:
@Disservin thank you for the PR! The existing code is correct, we're adding modifier in the _parent_ component, not to the `input` in child component

Disservin:
oh yeah mh.. sorry dunno what i was thinking :D

PR #2905, opened 2024-06-25
docs(withModifiers): change type of withModifiers param

Aaron-zon:
## Description of Problem

change type of withModifiers param
update documentation to reflect the change in #10856, where the type of the second parameter of withModifiers was changed from string[] to ModifierGuardsKeys[]

[vue/core #10856](https://github.com/vuejs/core/pull/10856)

image


NataliaTepluhina (approved):
@Aaron-zon thank you for the update!

PR #2925, opened 2024-07-10
docs: fix SSG explanation

flogehring:
## Description of Problem

The docs state that with SSG data may not change between deploys.

## Proposed Solution

Rephrase it to say data may not change between requests instead.

### DIFF
--- a/src/guide/scaling-up/ssr.md
+++ b/src/guide/scaling-up/ssr.md
@@ -40,7 +40,7 @@ Before using SSR for your app, the first question you should ask is whether you 
 
 **Static Site Generation (SSG)**, also referred to as pre-rendering, is another popular technique for building fast websites. If the data needed to server-render a page is the same for every user, then instead of rendering the page every time a request comes in, we can render it only once, ahead of time, during the build process. Pre-rendered pages are generated and served as static HTML files.
 
-SSG retains the same performance characteristics of SSR apps: it provides great time-to-content performance. At the same time, it is cheaper and easier to deploy than SSR apps because the output is static HTML and assets. The keyword here is **static**: SSG can only be applied to pages consuming static data, i.e. data that is known at build time and does not change between deploys. Every time the data changes, a new deployment is needed.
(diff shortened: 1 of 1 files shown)


AloisSeckar:
I believe the current wording is correct. "Static" means that no matter how many **_requests_** you do, you'll always receive the very same file(s). To change anything in the content, you need to **deploy** a new version of application with updated files.

Purely static sites should be HTML + CSS + media files only. No database, no web services, no processing on server... 

But I think the term is also often used for JavaScript files, that are "static" from the server perspective (they are always served the same), but dynamic for the user - JavaScript on client allows interaction with the content. And this might cause the confusion.

flogehring:
> Purely static sites should be HTML + CSS + media files only. No backend, no database, no web services, no processing on server...

They **are** being processed on the server (or in CI). During **G**eneration in the build process, which may very well include backends and databases etc.  Purely static files are the results of that with SSG. 

The current text said, that the data cannot change between deploys. That sentence would only make sense to me, if you separate the meaning of the build process from the actual deployment as in hosting. If so, the wording should be updated further, to include that notion and thus make it more clear.

Considering the possibly dynamic data at build time I still think that my PR makes more sense than the current wording.

AloisSeckar:
If you make request to `/foo` on SSG site 1000 times, you will receive the very same `foo.html` every time. Data are not changed between **requests**. To change the content, you have to statically build and **deploy** a new version of the application.

flogehring:
Yes absolutely! 😊 

That is why the previous wording was wrong.

AloisSeckar:
Oh, now I am finally getting you. Both sentences are true, but the updated one sounds more explanatory. You're right, sorry.

flogehring:
> Oh, now I am finally getting you. Both sentences are true, but the updated one sounds more explanatory. You're right, sorry.

Glad we're on the same page.

Do you think it would help to change `consuming` into `providing` in the previous sentence?


>The keyword here is **static**: SSG can only be applied to pages ~~consuming~~ providing static data i.e. data that is known at build time and can not change between requests.

AloisSeckar:
> 
> Do you think it would help to change `consuming` into `providing` in the previous sentence?
> 
> > The keyword here is **static**: SSG can only be applied to pages ~consuming~ providing static data i.e. data that is known at build time and can not change between requests.

Now when you point it out, yes. Although we can say a site "consumes" input data to create a content, more natural way of thinking about the concept probably would be to say a site "provides" data to the user...

PR #2929, opened 2024-07-11
Fixed error in docs of List Rendering

thibautvs:
## Description of Problem

`<li v-for="value in myObject">` lists the values of `myObject`, not its keys as suggested via the `Object.keys()` analogy in the docs.

## Proposed Solution

Replace `Object.keys()` analogy to `Object.values()`


NataliaTepluhina (approved):
@thibautvs technically the sentence is correct - the iteration _order_ is based on `Object.keys`, while we do display the contents of `Object.values` or rather `obj[key]` in this case. But I think it's quite easy to misread the sentence here and in _most_ cases `Object.values` order matched `Object.keys` closely so I'll approve the change

PR #2950, opened 2024-07-27
fix grammar: step 2 description

sachinsuresh95:
## Description of Problem
Step-2 description
Grammar error in
"State that can trigger updates when changed are considered reactive"

## Proposed Solution
Change to 
"States that can trigger updates when changed are considered reactive"


NataliaTepluhina (commented):
I believe current version doesn't contain a mistake anymore: "state _is_ reactive".

PR #2951, opened 2024-07-27
docs: add typing global custom directives

hmd-ali:


NataliaTepluhina:
@hmd-ali thank you for the PR! The change looks good but could we please move it to the TypeScript part of the guide? The rest of the directives documentation is plain JS, so adding typing guide looks a bit off. We can add a link at the bottom of directives doc to TS doc too.

hmd-ali:
I just noticed your message @NataliaTepluhina. The TS sections are Overview,  ts with composition and ts with options api. Where should it go as it's for both apis?

NataliaTepluhina:
> I just noticed your message @NataliaTepluhina. The TS sections are Overview, ts with composition and ts with options api. Where should it go as it's for both apis?

I would probably split it in two and add to both APIs

PR #2975, opened 2024-08-20
v-model.number: clarify unexpected behavior

simon04:
Clarify unexpected behavior regarding un-parsable or empty input until/unless https://github.com/vuejs/core/issues/8214 and https://github.com/vuejs/core/pull/8215 are fixed.

### DIFF
--- a/src/guide/essentials/forms.md
+++ b/src/guide/essentials/forms.md
@@ -442,7 +442,9 @@ If you want user input to be automatically typecast as a number, you can add the
 <input v-model.number="age" />
 ```
 
-If the value cannot be parsed with `parseFloat()`, then the original value is used instead.
+:::warning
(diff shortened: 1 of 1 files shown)


src/guide/essentials/forms.md
```diff
@@ -442,7 +442,9 @@ If you want user input to be automatically typecast as a number, you can add the
 <input v-model.number="age" />
 ```
 
-If the value cannot be parsed with `parseFloat()`, then the original value is used instead.
+:::warning
```
NataliaTepluhina, line 445: **suggestion** could we please remove the `warning` wrapper here? We're intentionally trying to minimize the number of highlighted blocks

simon04: Done.

PR #2990, opened 2024-09-03
docs: remarks about minor versions

AloisSeckar:
follow up to b6339b62a9522eb195a24ef74ee6392d65d2bfd7

In mentioned commit a lot of "3.x+" badges were removed from docs and replaced with text "Only suported in 3.x+"

I found two more places, where 3.1 / 3.3 were still present, so I altered them in the same manner.

I also noticed sometimes the text is placed before function description and sometimes after. I took liberty to unify the usage and put all of them **before**, because I think this good to know in advance whether I can use it in my project or not.

### DIFF
--- a/src/api/application.md
+++ b/src/api/application.md
@@ -285,9 +285,9 @@ Provide a value that can be injected in all descendant components within the app
 
 ## app.runWithContext() {#app-runwithcontext}
 
-Execute a callback with the current app as injection context.
-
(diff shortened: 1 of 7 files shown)


yyx990803:
Thank you!

AloisSeckar:
> Thank you!

Glad I can help :) I am transfering the recent changes to czech translation, so I had good chance to notice it.

PR #2991, opened 2024-09-04
docs: update ref attribute example

Thy3634:
ref -> useTemplateRef

## Description of Problem

```ts
const p = ref() // [!code --]
const pRef = useTemplateRef('p') // [!code ++]
```

### DIFF
--- a/src/api/built-in-special-attributes.md
+++ b/src/api/built-in-special-attributes.md
@@ -60,7 +60,7 @@ Denotes a [template ref](/guide/essentials/template-refs).
   <script setup>
   import { ref } from 'vue'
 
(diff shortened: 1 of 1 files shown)


src/api/built-in-special-attributes.md
```diff
@@ -60,7 +60,7 @@ Denotes a [template ref](/guide/essentials/template-refs).
   <script setup>
   import { ref } from 'vue'
```
yyx990803, line 61: Should add import of `useTemplateRef` here.

yyx990803 (changes requested):

PR #3000, opened 2024-09-06
docs: add onWatcherCleanup example

Arman19941113:
## Description of Problem

Some guys might think, "Why not just use onCleanup to handle cleanup?".

## Proposed Solution

This example might be helpful.


NataliaTepluhina:
@Arman19941113 I am not sure I follow: does the added example explain how to abstract the existing method to a standalone util? If yes, I am not sure it's the API doc responsibility to provide an example for code composition 🤔

PR #3001, opened 2024-09-06
feat(guide/computed): add previous to computed

ferferga:
This was shipped silently in Vue 3.4. See [PR](https://github.com/vuejs/core/pull/9497)

### DIFF
--- a/src/guide/essentials/computed.md
+++ b/src/guide/essentials/computed.md
@@ -259,6 +259,112 @@ Now when you run `fullName.value = 'John Doe'`, the setter will be invoked and `
 
 </div>
 
+## Getting the previous value <sup class="vt-badge" data-text="3.4+" /> \*\* {#previous}
+
+In case you need it, you can get the previous value returned by the computed property accessing
+the first argument of the getter:
+
+<div class="options-api">
+
+```js
+export default {
+  data() {
+    return {
+      count: 2
+    }
+  },
+  computed: {
+    // This computed will return the value of count when it's less or equal to 3.
+    // When count is >=4, the last value that fulfilled our condition will be returned
+    // instead until count is less or equal to 3
+    alwaysSmall(previous) {
+      if (this.count >= 3) {
+        return this.count;
+      }
+
+      return previous;
+    }
+  }
+}
+```
+</div>
+
+<div class="composition-api">
+
+```vue
+<script setup>
+import { ref, computed } from 'vue'
+
+const count = ref(2)
+
+// This computed will return the value of count when it's less or equal to 3.
+// When count is >=4, the last value that fulfilled our condition will be returned
+// instead until count is less or equal to 3
+const alwaysSmall = computed((previous) => {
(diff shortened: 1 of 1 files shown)


src/guide/essentials/computed.md
```diff
@@ -259,6 +259,114 @@ Now when you run `fullName.value = 'John Doe'`, the setter will be invoked and `
...
+    return {
+      count: 2
+    }
+  },
+  computed: {
+    // This computed will return the value of count when it's less or equal to 3.
+    // When count is >=4, the last value that fulfilled our condition will be returned
+    // instead until count is less or equal to 3
+    alwaysSmall(previous) {
+      if (this.count >= 3) {
```
NataliaTepluhina, line 283: **question**: should this be `<=3` according to the explanation above?

src/guide/essentials/computed.md
```diff
@@ -259,6 +259,114 @@ Now when you run `fullName.value = 'John Doe'`, the setter will be invoked and `
...
+<script setup>
+import { ref, computed } from 'vue'
+
+const count = ref(2)
+
+// This computed will return the value of count when it's less or equal to 3.
+// When count is >=4, the last value that fulfilled our condition will be returned
+// instead until count is less or equal to 3
+const alwaysSmall = computed((previous) => {
+  if (count.value >= 3) {
```
NataliaTepluhina, line 306: Same note about <=3

NataliaTepluhina (changes requested):

ferferga:
Good catch @NataliaTepluhina! Changed the logic a few times while working on this because I wasn't convinced in the examples that I was providing and completely missed this!

I also added the closing `</script>` tag that was missing from one of the Composition API's examples in my last commit.

src/guide/essentials/computed.md
```diff
@@ -259,6 +259,115 @@ Now when you run `fullName.value = 'John Doe'`, the setter will be invoked and `
...
+export default {
+  data() {
+    return {
+      count: 2
+    }
+  },
+  computed: {
+    alwaysSmall: {
+      get(previous) {
+        if (this.count >= 3) {
```
NataliaTepluhina, line 330:
```suggestion
        if (this.count <= 3) {
```

src/guide/essentials/computed.md
```diff
@@ -259,6 +259,115 @@ Now when you run `fullName.value = 'John Doe'`, the setter will be invoked and `
...
+
+```vue
+<script setup>
+import { ref, computed } from 'vue'
+
+const count = ref(2)
+
+const alwaysSmall = computed({
+  get(previous) {
+    if (count.value >= 3) {
```
NataliaTepluhina, line 355:
```suggestion
    if (count.value <= 3) {
```

NataliaTepluhina:
Thank you @ferferga! I've updated also the examples for writable computeds and I think we're good to go now

PR #3002, opened 2024-09-06
fix(guide/suspense): missing v3.3+ badge in suspensible prop

ferferga:
## Description of Problem

The `suspensible` prop was added in Vue 3.3 and I forgot to add it to #2785


yyx990803:
We only use badges for features added in the latest minor now. For older versions, we use a list item at the top of that section instead. e.g. `- Only supported in 3.3+`. Would you mind updating this PR and also #3001?

ferferga:
@yyx990803 Both has been updated!

NataliaTepluhina (approved):

PR #3014, opened 2024-09-11
typo.

zhangenming:
## Description of Problem
I think it's a forget

### DIFF
--- a/src/guide/components/v-model.md
+++ b/src/guide/components/v-model.md
@@ -68,7 +68,7 @@ const emit = defineEmits(['update:modelValue'])
 
 <template>
   <input
-    :value="modelValue"
+    :value="props.modelValue"
     @input="emit('update:modelValue', $event.target.value)"
   />
 </template>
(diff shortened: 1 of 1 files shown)


AloisSeckar:
It is not forgotten. As long as the props are defined via `defineProps` macro inside `<script setup>`, they are automatically unwrapped and available directly in `<template>`. See for example [HERE](https://vuejs.org/guide/essentials/component-basics.html#passing-props). Your update would also work, but it is not necessary.

In the current example, the assignments into `props` and `emits` are de-facto useless, as we are not working with them further in `<script setup>` section. The code could be re-written as:

```vue
<script setup>
defineProps(['modelValue'])
defineEmits(['update:modelValue'])
</script>

<template>
  <input
    :value="modelValue"
    @input="$emit('update:modelValue', $event.target.value)"
  >
</template>
```

@NataliaTepluhina Do you think this will improve clarity? Or is it rather against coding style established in other examples?

bencodezen (approved):
Great catch. You're totally right. Thanks for the PR!

PR #3015, opened 2024-09-12
improves `api/sfc-script-setup` defineModel TS examples

mostafa-nematpour:
This PR improves the comment's of code for `defineModel with TypeScript`

https://vuejs.org/api/sfc-script-setup.html#usage-with-typescript


yyx990803:
This is intended annotation and commonly used syntax to visualize types with comments.

PR #3028, opened 2024-09-17
docs: update comment following #2994

AloisSeckar:
## Description of Problem

In #2994 update, one comment was omitted from the update, so now it would be confusing

## Proposed Solution

Added the directive change from `v-focus` to `v-highlight` in comment


AloisSeckar:
Fix was already made in #3027

PR #3031, opened 2024-09-19
docs: correct usage examples of `useTemplateRef`

DamianGlowala:
## Description of Problem

`useTemplateRef` cannot accept `null` and does not need union with `null` in generic param type, as `null` will be added to the return type internally.


DamianGlowala:
cc @NataliaTepluhina

PR #3055, opened 2024-10-09
Add docs addressing "$attrs no longer contains events declared in the emits option" issue

stevebauman:
## Description of Problem

There is a long-standing discussion from 2021 with comments as recent as September 6th on the disappearance of event listeners in `$attrs` when those events are declared in the `emits` property:

https://github.com/vuejs/rfcs/discussions/397

This is documented here, but with no solution proposed:

https://vuejs.org/api/options-state.html#emits

Screenshot 2024-10-09 at 11 17 47 AM

## Proposed Solution

Additional docs can be added for those who discover that this occurs so that developers can properly utilize this un-documented functionality in Vue (or, if the docs do exist, correct me if I'm wrong) that allows developers to create props for these event listeners that Vue will auto-populate if they conform to a specific naming convention.

## Additional Information

N/A

### DIFF
--- a/src/api/options-state.md
+++ b/src/api/options-state.md
@@ -433,20 +433,20 @@ Declare the custom events emitted by the component.
   }
   ```
 
-Listeners on events may be captured in components by defining them as properties with the naming scheme of `on{Event}`:
-
-> If an event contains a colon, use the naming scheme `on{Event}:{event}`.
-
-```js
-export default {
-  props: {
-    onCheck: Function,
-    'onClick:export': Function,
-  },
-
-  emits: ['check', 'click:export'],
-}
-```
+  Listeners on events may be captured in components by defining them as properties with the naming scheme of `on{Event}`:
+  
+  > If an event contains a colon, use the naming scheme `on{Event}:{event}`.
+  
+  ```js
+  export default {
+    props: {
+      onCheck: Function,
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@stevebauman 

If we want to add this part of the documentation, let's start with _why_ do we want/need to check on the event. Otherwise the use-case is not clear (I know there are examples in the thread, but we need some "you might find yourself in a need to check for a listener bound).

stevebauman:
@NataliaTepluhina Ok sounds good! Would you be able to share an example of the style you may be looking for from an existing doc page so I can write it in the format that you may be expecting? 🙏 

The existing docs on this API page don't share any "why"'s, only "how"'s...

NataliaTepluhina:
@stevebauman a good example can be a beginning of the Forms section in the Guide: https://vuejs.org/guide/essentials/forms.html - we show first how we could do it with our "known" tools and then introduce `v-model`. Here, we could say that "there might be a case where you need to know about passed listeners (case example)" and then proceed with explaining _how_. Personally, I never had the need to check for the existing listener so you might need to come with "real" example yourself

stevebauman:
Ok thanks @NataliaTepluhina! Marking as draft and will update this a few hours (away from computer atm) and will mark ready once good to go for re-review.

Issue #3058, opened 2024-10-11
Clarification on Multiple v-model Bindings in the Documentation


netosts:
## Feature Request

Currently, the documentation lacks detailed information on how to handle `v-model` updates, especially when dealing with multiple bindings. Understanding this is crucial for working with Vue's two-way data binding system, particularly using the `@update:model-value` event.

Most importantly, there is no mention of handling multiple `v-model` bindings updates—neither in the official documentation nor through external resources. To handle this in practice, one can use the following syntax:

```vue
<UserName
  v-model:first-name="first"
  v-model:last-name="last"
  @update:first-name.model-value="foo"
  @update:last-name.model-value="foo2"
/>
```

This approach enables handling updates for multiple bound values (e.g., `first-name` and `last-name`), but the process isn't well-documented, which can cause confusion for developers.

I propose adding a section in the documentation that clearly explains this pattern, with examples and guidelines on how to handle updates for multiple `v-model` bindings. This would greatly enhance the developer experience and provide clarity for a commonly needed feature.

AloisSeckar:
Thanks for the suggestion. However:
1. The construct you are talking about belongs to https://vuejs.org/guide/components/v-model.html#multiple-v-model-bindings where I beleive it already is, with greater level of detail too.
2. The purpose of `v-model` is to abstract developer from having to use `@update` event. You just declare `v-model` binding and it works out-of-the-box updating the values from within the child component to the parent. If you need to react to the change in parent with side effects, I'd rather bet on a watcher, because then it has clearer relation to the variable change, not to some component event.

PR #3059, opened 2024-10-11
docs: add section for handling multiple v-model bindings updates

netosts:
#3058

## Description of Problem
Currently, the documentation lacks detailed information on how to handle v-model updates, especially when dealing with multiple bindings. Understanding this is crucial for working with Vue's two-way data binding system, particularly using the @update:model-value event.

Most importantly, there is no mention of handling multiple v-model bindings updates—neither in the official documentation nor through external resources.

## Proposed Solution
I propose adding a section in the documentation that clearly explains this pattern, with examples and guidelines on how to handle updates for multiple v-model bindings. This would greatly enhance the developer experience and provide clarity for a commonly needed feature.


NataliaTepluhina:
@netosts thank you for the contribution! However, as @AloisSeckar already explained in https://github.com/vuejs/docs/issues/3058#issuecomment-2406610867, we do not really need to use `update:modelValue` with `v-model`. We _can_. just like we can _watch_ the model value and track the updates, but I wouldn't go as far as recommending this in the docs.

PR #3061, opened 2024-10-12
doc:add a warn about '~' in src-imports

Bayn-Web:
## Description of Problem
`\~` is not mentioned in src-imports ,which is different from the others.
## Proposed Solution
Add description.
## Additional Information
Nothing.

### DIFF
--- a/src/api/sfc-spec.md
+++ b/src/api/sfc-spec.md
@@ -136,6 +136,13 @@ Beware that `src` imports follow the same path resolution rules as webpack modul
 </unit-test>
 ```
 
+:::warning Note
+While using aliases in `src`,don't start with `~`(It can be used in `import` seq ), anything after it is interpreted as a module request. This means you can reference assets inside node modules:
+```vue
+<img src="~some-npm-package/foo.png">
+```
+:::
+
 ## Comments {#comments}
 
 Inside each block you shall use the comment syntax of the language being used (HTML, CSS, JavaScript, Pug, etc.). For top-level comments, use HTML comment syntax: `<!-- comment contents here -->`


Bayn-Web:
Mentioned#https://github.com/vuejs/core/issues/11367

src/api/sfc-spec.md
```diff
@@ -136,6 +136,13 @@ Beware that `src` imports follow the same path resolution rules as webpack modul
 </unit-test>
 ```
 
+:::warning Note
+While using aliases in `src`,don't start with `~`(It can be used in `import` seq ), anything after it is interpreted as a module request. This means you can reference assets inside node modules:
```
NataliaTepluhina, line 140:
```suggestion
While using aliases in `src`, don't start with `~`(it can be used in `import` seq ), anything after it is interpreted as a module request. This means you can reference assets inside node modules:
```

I think we need more clarification on `import seq`, it doesn't sound clear enough for new developers

NataliaTepluhina (changes requested):

src/api/sfc-spec.md
Bayn-Web: Should we add a code block to explain?
Like this:
```vue
// use as alias `~`
import XXX from "~/X"
```

NataliaTepluhina: Honestly, I feel like skipping it altogether will reduce the confusion 😅 We're speaking about `src` here so let's not mention global imports. I'll leave a suggestion

Bayn-Web: Ok, I'll create a new pr to clarify `import` behave in [`<script setup>`](https://vuejs.org/api/sfc-script-setup.html).

NataliaTepluhina (approved):

PR #3063, opened 2024-10-14
Update event-handling.md hyperlink

liangpengyv:
## Description of Problem

The original link `https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values` of MDN has expired.

## Proposed Solution

The new link address is `https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values`.


NataliaTepluhina:
Already fixed in https://github.com/vuejs/docs/pull/3045

PR #3070, opened 2024-10-16
add useCssModule example

runyasak:
## Description of Problem

The style module is an interesting feature in Vue.js, and the `useCssModule` helper is also quite useful. However, it seems like `useCssModule` lacks an example, which would make it easier for beginners to understand and use.

## Proposed Solution

I have added a simple example for `useCssModule` to demonstrate how to implement it.

### Screenshot result:
image

---

If you have any feedback or suggestions, please do not hesitate to let me know. 😁

### DIFF
--- a/src/api/sfc-css-features.md
+++ b/src/api/sfc-css-features.md
@@ -153,6 +153,26 @@ useCssModule()
 useCssModule('classes')
 ```
 
+- **Example**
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
Thank you for the added example!

PR #3071, opened 2024-10-21
docs: use badge instead of text for version support

HigherOrderLogic:
## Description of Problem
Use badge to show what version is the helper is available in for consistency.

## Proposed Solution
N/A

## Additional Information
N/A


Cr0zy07:
> We only use badges for features added in the latest minor now. For older versions, we use a list item at the top of that section instead. e.g. `- Only supported in 3.3+`.

See https://github.com/vuejs/docs/pull/3002#issuecomment-2352282817

NataliaTepluhina:
Thank you for your contribution! However, as @Cr0zy07 stated above, we only use badges for the latest minor version

PR #3073, opened 2024-10-22
Corrected and improved docs on type definitions for Custom Elements.

trusktr:
The doc was incorrect for two reasons:

- For Vue-based custom elements, the types passed into `GlobalComponents` need to be the Vue component types, not the custom element types, or the custom elements will not be type checked in Vue templates.
- The name of the elements need to have at least two words and a hyphen, otherwise with single words the custom elements will not work at all.

A few questions:

- Should we include working examples?
  - Here's an example for non-Vue custom elements: https://github.com/lume/element/tree/vue-type-issue/examples/kitchen-sink-vue
- Should we include something like the `DefineCustomElement` type helper in `vue`?
- Should we add type tests?

Having examples/helpers/tests could prevent the docs from being incorrect.

There's a conversation in [`#typescript` in Vue's Discord](https://discord.com/channels/325477692906536972/1298526984078364754) where we chatted about and worked on this.

The conversation includes the idea of updating `vue-tsc` so that Custom Element type definitions can instead be added to `IntrinsicElementAttributes`, which I tried at first but it currently does not work. Only built-in elements on `IntrinsicElementAttributes` work, so I resorted to using `GlobalComponents` with the `$props` and `$emit` hacks which results in element ref types exposing non-existent `$props` and `$emit` properties to the user.

So one more question is:

- Do we merge this update now so it is correct with the current state of `vue-tsc`? Or do we wait until `vue-tsc` can be updated so that hyphenated element names can successully be added to `IntrinsicElementAttributes`, and update these docs to show the proper way of defining element types with `IntrinsicElementAttributes`? cc @KazariEX @johnsoncodehk

### DIFF
--- a/src/guide/extras/web-components.md
+++ b/src/guide/extras/web-components.md
@@ -220,6 +220,8 @@ If the custom elements will be used in an application that is also using Vue, yo
 It is recommended to export the individual element constructors to give your users the flexibility to import them on-demand and register them with desired tag names. You can also export a convenience function to automatically register all elements. Here's an example entry point of a Vue custom element library:
 
 ```js
+// elements.js
+
 import { defineCustomElement } from 'vue'
 import Foo from './MyFoo.ce.vue'
 import Bar from './MyBar.ce.vue'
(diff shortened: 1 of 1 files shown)


trusktr:
This is not there yet. Firstly, custom elements are not Vue components, so trying to type them using Vue component types is not working well. `ref` is currently broken, and `ref.value` is currently `unknown` with this approach. 

`Component` and `DefineComponent` are for Vue components, not well suited for this.

I don't see yet how to hook a custom element in to be treated like a built in element. Augmenting the `IntrinsicElementAttribute` interface would be the intuitive simple way to do it, and making custom elements all be treated the same as builtin elements, but that does not work at all currently (Vue or the language server or something just completely ignores any custom elements added to that interface).

trusktr:
Alright, updated and ready for initial review.

NataliaTepluhina (approved):
@trusktr thank you for working on this, it's an extremely valuable and detailed addition to the documentation!

PR #3075, opened 2024-10-23
Update transition-group.md - add detail about prop moveClass

femans:
## Description of Problem
Prop spec of TransitionGroup is missing on component documentation page. 

## Proposed Solution
Added the missing information.

## Additional Information
While building a library which is making use of TransitionGroup, it took me a loooong minute to find out how to pass the move-class as a prop, (that is, I thought for a while it was not possible and it was only possible using CSS, which is annoying if you want to avoid that), because it was not written on the main documentation page. 
Now I checked again and found it in the API spec and found it after all, yey! I updated my library to include the option for this by using the correct props type.

### DIFF
--- a/src/guide/built-ins/transition-group.md
+++ b/src/guide/built-ins/transition-group.md
@@ -81,6 +81,7 @@ Now it looks much better - even animating smoothly when the whole list is shuffl
 [Full Example](/examples/#list-transition)
 
 ### Custom TransitionGroup classes
+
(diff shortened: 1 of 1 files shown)


src/guide/built-ins/transition-group.md
```diff
@@ -80,6 +80,9 @@ Now it looks much better - even animating smoothly when the whole list is shuffl
 
 [Full Example](/examples/#list-transition)
 
+### Custom TransitionGroup classes
```
NataliaTepluhina, line 83:
```suggestion
### Custom TransitionGroup classes

```

PR #3076, opened 2024-10-24
docs: fix the example in With Components section

white-black333:
the example in With Components section should be modified.


NataliaTepluhina:
I don't think this change is necessary, we're stating it's `MyComponent`'s template right above the code snippet

PR #3090, opened 2024-11-04
docs: add explanation for mouse button event modifiers

midskyey:
## Description of Problem

The section [Mouse Button Modifiers](https://vuejs.org/guide/essentials/event-handling.html#mouse-button-modifiers) does not contain sufficient explanation for what `.left`, `.right`, and `.middle` represent.

## Proposed Solution

Add a note block about "main", "secondary", "auxiliary", and different pointing devices.

### DIFF
--- a/src/guide/essentials/event-handling.md
+++ b/src/guide/essentials/event-handling.md
@@ -357,3 +357,7 @@ The `.exact` modifier allows control of the exact combination of system modifier
 - `.middle`
 
 These modifiers restrict the handler to events triggered by a specific mouse button.
+
+::: tip Note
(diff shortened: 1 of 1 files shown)


src/guide/essentials/event-handling.md
```diff
@@ -357,3 +357,7 @@ The `.exact` modifier allows control of the exact combination of system modifier
 - `.middle`
 
 These modifiers restrict the handler to events triggered by a specific mouse button.
+
+::: tip Note
```
NataliaTepluhina, line 361: **suggestion**: we're trying to reduce the number of highlighted blocks and use them only if absolutely necessary. Would you mind removing the wrapping `tip` and keeping the added information?

NataliaTepluhina (changes requested):

src/guide/essentials/event-handling.md
midskyey: Sure, but moved the emphasis to the beginning of the sentence. Or would it be better to just keep "`.left`" as the beginning?

NataliaTepluhina: @midskyey the changed emphasis looks good to me 👍🏻

NataliaTepluhina (approved):

PR #3092, opened 2024-11-05
Update the transition group

tisma95:
## Description of Problem

Replace the absolute url by relatve url  and add missing anchor.


NataliaTepluhina (approved):
@tisma95 thank you for fixing this!

PR #3096, opened 2024-11-06
docs: fix grammar

iiio2:
I think it would be `props` instead of `props.` with a full stop because it is just one line.


NataliaTepluhina:
Thank you for the pull request! Unfortunately, the period is intended there, we prefix the name with `props.` so `foo` becomes `props.foo`

iiio2:
Thanks @NataliaTepluhina .

PR #3112, opened 2024-11-23
fix: Incorrect URL when using the virtual path.

scccboom:


brc-dd:
We are not using a subpath base in any of the docs side or its translations. withBase will just return the value as it is.

PR #3117, opened 2024-11-25
docs(computed): remove semicolons

iiio2:
## Description of Problem
In the entire vue.js documentation, codes are written without semicolons. Having semicolons in code is not wrong, but for keeping consistency with the other codes in the vue.js doc, I think we can remove the semicolons in the given examples. Only here semicolons are used. I think it would be better to remove the semicolons here to maintain consistency with the other codes in vue.js doc.

### DIFF
--- a/src/guide/essentials/computed.md
+++ b/src/guide/essentials/computed.md
@@ -281,10 +281,10 @@ export default {
     // instead until count is less or equal to 3
     alwaysSmall(previous) {
       if (this.count <= 3) {
-        return this.count;
+        return this.count
       }
 
-      return previous;
+      return previous
     }
   }
 }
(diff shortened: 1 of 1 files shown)


src/guide/essentials/computed.md
```diff
@@ -328,13 +328,13 @@ export default {
     alwaysSmall: {
       get(previous) {
         if (this.count <= 3) {
-          return this.count;
+          return this.count
         }
 
         return previous;
```
AloisSeckar, line 334: sorry for late review (I noticed while implementing chages into translation) - the semicolon remained here

iiio2:
Thanks @AloisSeckar

PR #3120, opened 2024-12-01
docs: fix punctuation

iiio2:
I think instead of using semicolon(;), we should use full stop (.) like the third line.


NataliaTepluhina:
@iiio2 thank you for the PR! I disagree with the suggestion, we use semicolons on all the points that are not final, and the period on the last point.

PR #3122, opened 2024-12-03
Update teleport.md

darrelfrancis:
1. Removed the implication that teleport is focussed on displaying the element _outside the Vue application_. The real focus is on displaying outside the original component. 

2. Improved grammar generally and reduced unnecessary words.


NataliaTepluhina (approved):
@darrelfrancis thank you for the improvement! 👍🏻

PR #3124, opened 2024-12-04
docs(guide/components/slots): clarify slot presence phrasing

alexchexes:
## Description of Problem
The original phrasing "**_whether or not a slot is present_**" was potentially misleading, as it suggested checking the existence of the slot itself rather than whether content was passed to it. Updated the explanation to clarify this distinction for better readability and accuracy.

## Proposed Solution
The original phrase was replaced with "...**_whether or not content has been passed to a slot_**"

## Additional Information
-

### DIFF
--- a/src/guide/components/slots.md
+++ b/src/guide/components/slots.md
@@ -298,7 +298,7 @@ function BaseLayout(slots) {
 
 ## Conditional Slots {#conditional-slots}
 
-Sometimes you want to render something based on whether or not a slot is present. 
+Sometimes you want to render something based on whether or not content has been passed to a slot. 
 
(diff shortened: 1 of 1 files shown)


alexchexes:
Oh, I just noticed another related phrase right below the one I addressed in this PR:

```
When the header / footer / default is present we want to wrap them to provide additional styling:
```

The phrase "When the header/footer/default is present..." could also be interpreted as referring to the slot's definition, rather than whether content is provided for the slot. A clearer alternative:

```
When content for the header / footer / default is present, we wrap it to provide additional styling:
```

Let me know if I should open another PR for this.

AloisSeckar:
You shouldn't open a new PR. Make a new commit into this one. The proposed change is adressing the same thing.

alexchexes:
>Make a new commit into this one

done

NataliaTepluhina (approved):

PR #3128, opened 2024-12-11
docs: move `data-allow-mismatch` to built-in-special-attributes.md

KazariEX:
## Description of Problem

language-tools uses the documentation in Built-ins to provide the completion of special attributes, while currently `data-allow-mismatch` is located in another page.


brc-dd:
IMO it's better to keep it in the SSR page for better discoverability. And only add a see more link in the special attributes page. Maybe with a note that it is treated same as usual data-* attributes by the compiler unlike the other attributes in the page. It only affects renderer behavior.

KazariEX:
You are right, the current document structure is clearer. Will do special processing on document crawling at the language tools side.

PR #3129, opened 2024-12-12
docs: add missing "the"

G-Rath:
## Description of Problem

The current sentence does not make grammatical sense

## Proposed Solution

Introduce a "the"


bencodezen (approved):
Thanks for the PR!

PR #3169, opened 2025-02-05
fix typo in faq.md

MohcineDev:
fix a typo in faq.md file


AloisSeckar:
Thanks for your PR.

However, _"this very something"_ is actually an English language concept of putting emphasis on the subject - https://www.quora.com/How-can-the-use-of-very-in-this-very-day-or-that-predicted-these-very-events-be-explained-to-a-second-language-speaker-of-English

+ you have used `npm` for building the project locally and posted your generated `package-lock.json` along with the PR, which is not desired, as this project uses `pnpm` as default package manager.

PR #3172, opened 2025-02-09
docs: update "Static Hoisting" to "Cache Static"

Arman19941113:
According to this PR https://github.com/vuejs/core/pull/11067 (which changes node hoisting to caching per instance), I believe the "Static Hoisting" section is outdated.


NataliaTepluhina (approved):

AloisSeckar:
Don't wanna sound as a grumbler, but this section has [a reference from Glossary section](https://vuejs.org/glossary/#hoisting), which now:

1. has broken link as the anchor was changed
2. is possibly outdated because it is not called "static hoisting" anymore

I will also try to make a fix

AloisSeckar:
Looks good to me, thanks for quick response 👍 

My "complaint" wasn't meant towards you but more to the fact the inconsistency wasn't spotted before getting merged.

PR #3175, opened 2025-02-11
docs: Clarify inject data when multiple parent provide the same key

kfina-planxy:
## Description of Problem
It was unclear in the docs, how the situation is handled if multiple parents provide the same key.

### DIFF
--- a/src/guide/components/provide-inject.md
+++ b/src/guide/components/provide-inject.md
@@ -169,6 +169,8 @@ export default {
 }
 ```
 
+If multiple parents provide data with the same key, inject will resolve to the first parent in its parent chain.
+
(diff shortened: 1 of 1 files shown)


AloisSeckar:
I am just translating this into cs.vuejs.org and this sentence doesn't clarified it much for me. Is it "first" going up from the root component or closest to the injecting child?

kfina-planxy:
Closest to the child:
Starting from the child that injects you go to its parent and check if its parent provides that key. If so, use that key, otherwise go to the its parent and check again.

Example:
```
    [ A ] provides: x=1
   /     \
[ C ]   [ B ] provides: x=2
x=1        \
           [ C ]
            x=2
```

PR #3182, opened 2025-02-20
docs(glossary): update "hoisting" and "cache static"

Arman19941113:
The glossary update was missed; see #3172.


AloisSeckar (approved):
This looks like a great solution to me - the explanation for "hoisting" in general is preserved and there is a new key for updated Vue-related page.

PR #3192, opened 2025-03-07
docs: fix grammar

SirDaev:
## Description of Problem
"Complexity" is the singular subject of the sentence, but the sentence uses the plural verb "grow".

## Proposed Solution
We should use "grows" instead of "grow".

## Additional Information
n/a


bencodezen (approved):
Good catch! Thanks @SirDaev!

PR #3204, opened 2025-03-26
docs: fix transition component casing

ocmrz:
## Description of Problem

On line 303 in the Transition document, it should be `<Transition>` instead of `<transition>`

## Proposed Solution

as above


bencodezen (approved):
Agreed. This should be consistent with the rest of the docs. Thanks @ocmrz!

PR #3205, opened 2025-03-26
docs: fix blockquote language

ocmrz:
## Description of Problem

On line 333 in the Transition page, `vue-html` should be used instead of `html`

## Proposed Solution

as above


bencodezen (approved):
Good catch. Thanks @ocmrz!

PR #3207, opened 2025-03-27
docs:add typing support documentation for plugins

jh-leong:
Added new section "Typing Plugins" in Composition API guide


src/guide/typescript/composition-api.md
```diff
@@ -476,3 +476,53 @@ const openModal = () => {
 ```
 
 Note that with `@vue/language-tools` 2.1+, static template refs' types can be automatically inferred and the above is only needed in edge cases.
+
+## Typing Plugins {#typing-plugins}
```
yyx990803, line 480: Since plugins are not Composition API specific, it feels a bit off to be placing it here. It's probably better to just move it directly into `plugins.md`.

yyx990803 (changes requested):

PR #3213, opened 2025-04-03
docs(guide): Props Stability

sofuxro:
related to [#13157](https://github.com/vuejs/core/issues/13157)

The example is missing `@click`  on the `ListItem` (this is how selected is triggered - even if the click event comes from the child, in more complex cases). 

To make it general and not be forced to move the event inside the child I would suggest using `v-memo` for this prop stability example.


yyx990803:
I don't think we should bring in `v-memo` in this example, it makes it much harder to understand. `v-memo` is a really advanced concept that should be discussed alone (which already has an item below), and also not needed in vapor mode.

PR #3228, opened 2025-04-23
Add Helix editor

hongquan:
## Description of Problem

Just want to introduce Helix editor, which has LSP built-in support (without any plugin) to Vue users.

## Additional Information

https://helix-editor.com/

### DIFF
--- a/src/guide/scaling-up/tooling.md
+++ b/src/guide/scaling-up/tooling.md
@@ -100,6 +100,9 @@ If you are looking for a lighter-weight alternative for no-build-step usage, che
 
 - Other IDEs that support the [Language Service Protocol](https://microsoft.github.io/language-server-protocol/) (LSP) can also leverage Volar's core functionalities via LSP:
 
+  - [Helix](https://helix-editor.com/) supports out-of-the-box, as long as you install [`@vue/language-server`](https://www.npmjs.com/package/@vue/language-server), the Vue language server. To work with Vue + TypeScript, you also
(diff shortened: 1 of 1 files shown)


src/guide/scaling-up/tooling.md
```diff
@@ -100,6 +100,9 @@ If you are looking for a lighter-weight alternative for no-build-step usage, che
 
 - Other IDEs that support the [Language Service Protocol](https://microsoft.github.io/language-server-protocol/) (LSP) can also leverage Volar's core functionalities via LSP:
 
+  - [Helix](https://helix-editor.com/) supports out-of-the-box, as long as you install [`@vue/language-server`](https://www.npmjs.com/package/@vue/language-server), the Vue language server. To work with Vue + TypeScript, you also
```
bencodezen, line 103: @hongquan You say that it supports Vue out of the box, but the proposed changes mention a number of prerequisites. For us to include this on the list, there would need to be a separate dedicated docs page to match the rest of the mentioned editors.

Example:

```
Helix Editor support via [helix-plugin](...).
```

bencodezen (changes requested):

src/guide/scaling-up/tooling.md
hongquan: Hi @bencodezen The prerequisites are just the neutral LSP servers, and please note that they are under `@vue` namespace, they are not made specifically for Helix. They are not Helix plugins.

PR #3230, opened 2025-04-30
Fix: provide/inject in plugins.md

leoelz:
plugins allow us to use `provide` to provide a function or attribute plugin, users are able to `inject` the plugin function or attribute

## Description of Problem
It says: _Plugins also allow us to use `inject` to provide a function or attribute to the plugin's users..._ and the example shows that `provide` is used. Also after the example it says _Plugin users will now be able to inject the plugin options..._

## Proposed Solution
Replace `inject` with `provide`

### DIFF
--- a/src/guide/reusability/plugins.md
+++ b/src/guide/reusability/plugins.md
@@ -97,7 +97,7 @@ Use global properties scarcely, since it can quickly become confusing if too man
 
 ### Provide / Inject with Plugins {#provide-inject-with-plugins}
 
-Plugins also allow us to use `provide` to provide a function or attribute to the plugin's users. For example, we can allow the application to have access to the `options` parameter to be able to use the translations object.
+Plugins also allow us to use `provide` to give plugin users access to a function or attribute. For example, we can allow the application to have access to the `options` parameter to be able to use the translations object.
 
 ```js{10}
 // plugins/i18n.js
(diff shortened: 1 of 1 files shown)


src/guide/reusability/plugins.md
```diff
@@ -97,7 +97,7 @@ Use global properties scarcely, since it can quickly become confusing if too man
 
 ### Provide / Inject with Plugins {#provide-inject-with-plugins}
 
-Plugins also allow us to use `inject` to provide a function or attribute to the plugin's users. For example, we can allow the application to have access to the `options` parameter to be able to use the translations object.
+Plugins also allow us to use `provide` to provide a function or attribute to the plugin's users. For example, we can allow the application to have access to the `options` parameter to be able to use the translations object.
```
bencodezen, line 100:
```suggestion
Plugins also allow us to use `provide` to give plugin users access to a function or attribute. For example, we can allow the application to have access to the `options` parameter to be able to use the translations object.
```

bencodezen: What do you think of this @leoelz?

I wanted to remove the duplicate use of "provide" which could be confusing to readers.

leoelz: Excellent Ben! I had the same feeling.

bencodezen (approved):

bencodezen:
Thanks for the PR @leoelz!

PR #3234, opened 2025-05-13
Fix: misplaced semicolon in crud example

joceqo:
## Description of Problem

this is more a typo

## Additional Information

this is really not something important, but I spotted it while going through the docs and example, I guess it can be clearer if fixed


joceqo:
Indeed there is no semicolon in the file 👍

bencodezen (approved):
Good catch @joceqo! Thanks for the PR!

PR #3249, opened 2025-06-16
Update composition-api.md 

owaiskhan310:
## Description of Problem
I was going through the documentations and found an error in the documentations. In the  "Typing Component Template Refs" section of "TS with composition API", there is a first code snippet below the description of "Typing Component Template Refs". In this code snippet when we are passing the ref in the component there is a typo error in the name of the ref. We have declared the ref as "compRef " and passed the ref as "comp". This needs to be fixed to ref="compRef".

## Proposed Solution
current:
<component :is="Math.random() > 0.5 ? Foo : Bar" ref="comp"

expected:
<component :is="Math.random() > 0.5 ? Foo : Bar" ref="compRef"

### DIFF
--- a/src/guide/typescript/composition-api.md
+++ b/src/guide/typescript/composition-api.md
@@ -432,7 +432,7 @@ const compRef = useTemplateRef<FooType | BarType>('comp')
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


owaiskhan310:
image

owaiskhan310:
mistakenly opened this issue

PR #3262, opened 2025-07-08
docs: add addendum on useId usage with computed

bencodezen:
## Description of Problem

There is currently no documentation around how `useId()` can cause errors in `computed()` due to conflicting instances.

## Proposed Solution

Since this is something people will likely run into, a warning callout is proposed to help users avoid this problem going forward.


brc-dd (approved):
LGTM. Though the ModelRef formatting can be reverted if prettier didn't update it 👀

brc-dd:
Though more that I think, using useId with computed won't make much sense for anyone. Like if you have a static suffix/prefix, then why not just concat it `` const ariaDescribedBy = `${id}-description` ``. And if you don't have a static string, then why not use computed to generate a unique key from other refs/props. And if you don't have a unique static string, then why not just use useId alone. Maybe we should just remove the example 😅. The warning itself is fine.

bencodezen:
Thanks for the review @brc-dd!

PR #3263, opened 2025-07-11
docs: sync vite need node version

btea:
## Description of Problem

The vite version currently supported by create-vue is 7, and the node version required by vite is as follows. In order to ensure that the created project is OK, you need to synchronize the node version information.

https://github.com/vitejs/vite/blob/588b3b5a44eef35bcca0e3079df579d50a528421/package.json#L6

### DIFF
--- a/src/guide/quick-start.md
+++ b/src/guide/quick-start.md
@@ -23,7 +23,7 @@ import { VTCodeGroup, VTCodeGroupTab } from '@vue/theme'
 :::tip Prerequisites
 
 - Familiarity with the command line
-- Install [Node.js](https://nodejs.org/) version 18.3 or higher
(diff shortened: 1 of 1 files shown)


src/guide/quick-start.md
```diff
@@ -23,7 +23,7 @@ import { VTCodeGroup, VTCodeGroupTab } from '@vue/theme'
 :::tip Prerequisites
 
 - Familiarity with the command line
-- Install [Node.js](https://nodejs.org/) version 18.3 or higher
+- Install [Node.js](https://nodejs.org/) version `^20.19.0 || >=22.12.0`
```
NataliaTepluhina, line 26: **question**: would any version higher than 20.19 work? It's better to avoid using logical operators in quick start and try to phrase it with words

PR #3264, opened 2025-07-12
docs: clarify HTML validator warnings for Vue-specific directives

adityaranjan08:
## What This Does

This adds a tip box to the `template-syntax.md` guide explaining that Vue-specific attributes like `v-if`, `@click`, and `v-bind` may appear as invalid in HTML validators or IDEs. These are Vue template directives and are expected to be handled by Vue's compiler.

## Why

Fixes [vuejs/vue#4428](https://github.com/vuejs/vue/issues/4428), which notes that new users may find these attributes invalid when writing Vue templates. This update clarifies that this is expected behavior and not a bug.

Thanks!


bencodezen:
Closing as the PR is stale and the attached issue is also old. If this is still an issue, please open a new issue PR. Thanks!

PR #3268, opened 2025-07-20
:lipstick: update reverse array example to modern syntax

lregaladohdez:
## Description of Problem

This is not a problem but a stylistic suggestion.

The example currently uses `[...array].reverse()`, which can be simplified using a more modern and concise syntax.

## Proposed Solution

Suggest using the simpler:

`array.toReversed()`

This achieves the same result reversing the array without mutating the original, but in a cleaner and more modern way.

## Additional Information

array.toReversed() was introduced in ECMAScript 2023

### DIFF
--- a/src/guide/essentials/list.md
+++ b/src/guide/essentials/list.md
@@ -436,5 +436,5 @@ Be careful with `reverse()` and `sort()` in a computed property! These two metho
 
 ```diff
 - return numbers.reverse()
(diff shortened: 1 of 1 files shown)


brc-dd:
closing for same reason as #2575

toReversed is a baseline 2023 feature and vite's default target is baseline-widely-available. If you use toReversed it won't work in some browsers without a polyfill.

PR #3273, opened 2025-08-05
Update ways-of-using-vue.md

WMcKibbin:
Change language of petite-vue to reflect its current status

## Description of Problem
The current documentation does not accurately reflect the status of petite-vue. I've been misled a few times reading the docs thinking that this was something that the Vue team was actively maintaining. 

## Proposed Solution
Use past tense language to describe this distribution and provide context and recommendation around its purpose and usage.

## Additional Information
This language can and probably should be altered and modified by the Vue team to properly meet their needs.


bencodezen (approved):
@WMcKibbin Thanks for the PR! I've modified the language a little since the team doesn't have an explicit recommendation as far as not using it.

The cautionary warning you've suggested should be enough to give people pause and then they can make their own decision as far as whether they want to build with it or not.

Appreciate the PR!

PR #3283, opened 2025-08-25
feat: add Vue Plugins to the resources list

Baroshem:
Added https://www.vue-plugins.org/ to the resources list. 

If the placement or the text is wrong, please let me know so that I will adjust it :)


Baroshem:
@yyx990803 @NataliaTepluhina 

Heyo guys! Let me know if we can merge it or if there is anything that should be changed :)

PR #3293, opened 2025-09-15
Fix typo: remove duplicate word "be"

stuartwiener:
## Description of Problem

In [the warning at the end of the useId() section](https://vuejs.org/api/composition-api-helpers.html#useid) the word "be" is duplicated. 

## Proposed Solution

Change the copy from

> `useId()` should be not be called inside a `computed()` property...

to 

> `useId()` should not be called inside a `computed()` property...


bencodezen (approved):
Good catch. Thanks @stuartwiener!

PR #3360, opened 2026-03-07
chore: migrate to oxfmt

ubugeeei:
just migrated, I've not confirm yet.
If anyone has some free time, I’d appreciate it if you could take a look first.


ubugeeei:
I'll take it on another

PR #3389, opened 2026-05-07
docs: typescript essential style guide

serkodev:
## Description of Problem
The original Essential Style Guide was designed for JavaScript. In this PR, we will create a Vue Essential Style Guide specifically for TypeScript.

## Proposed Solution
- Please visite `/style-guide/rules-essential-ts.html` to this guide
- Updated style guide to TypeScript friendly
- Added more guides:
  - Access DOM or component instances
  - Avoid destructuring from reactive() directly
  - Watching reactive state
  - Prefer computed over watcher-assigned derived refs
  - Use eager watchers instead of duplicate initial calls

Relared #3370

### DIFF
--- a/src/style-guide/index.md
+++ b/src/style-guide/index.md
@@ -4,42 +4,44 @@ outline: deep
 
 # Style Guide {#style-guide}
 
-::: warning Note
-This Vue.js Style Guide is outdated and needs to be reviewed. If you have any questions or suggestions, please [open an issue](https://github.com/vuejs/docs/issues/new).
+::: warning Status
+This section is under active revision. Priority A: Essential is currently the maintained section of the guide. Priority B, C, and D are retained as legacy reference content.
 :::
 
-This is the official style guide for Vue-specific code. If you use Vue in a project, it's a great reference to avoid errors, bikeshedding, and anti-patterns. However, we don't believe that any style guide is ideal for all teams or projects, so mindful deviations are encouraged based on past experience, the surrounding tech stack, and personal values.
+This is the official style guide for Vue-specific code. It is meant to help teams make sound architectural, semantic, and maintainability decisions in modern Vue applications. However, no style guide is ideal for every team or project, so mindful deviations are still encouraged when they are deliberate and well understood.
 
-For the most part, we also avoid suggestions about JavaScript or HTML in general. We don't mind whether you use semicolons or trailing commas. We don't mind whether your HTML uses single-quotes or double-quotes for attribute values. Some exceptions will exist however, where we've found that a particular pattern is helpful in the context of Vue.
+This guide focuses on a small set of human-facing Vue rules. It is not intended to mirror every rule that lint tooling may enforce. Tools such as [eslint-plugin-vue](https://eslint.vuejs.org/), formatters, and other static analysis can still cover broader mechanical correctness and consistency checks.
 
-Finally, we've split rules into four categories:
+For the most part, we avoid suggestions about JavaScript or HTML in general. We do not try to settle choices like semicolons, trailing commas, or quote style unless they become specifically important in the context of Vue.
+
+## How to use this guide {#how-to-use-this-guide}
+
+- Start with the [essential rules](./rules-essential). They are the primary maintained part of this guide and focus on component contracts, explicit data flow, component boundaries, and the Vue mental model.
+- Treat Priority B, C, and D as legacy reference pages. They may still contain useful conventions, but they are not part of the current maintained scope.
+- If your team uses [eslint-plugin-vue](https://eslint.vuejs.org/), expect tooling to enforce additional syntax, correctness, and consistency rules that are not all repeated here.
 
 ## Rule Categories {#rule-categories}
 
-### Priority A: Essential (Error Prevention) {#priority-a-essential-error-prevention}
+### Priority A: Essential (Maintained) {#priority-a-essential-error-prevention}
 
-These rules help prevent errors, so learn and abide by them at all costs. Exceptions may exist, but should be very rare and only be made by those with expert knowledge of both JavaScript and Vue.
+These rules focus on component contracts, explicit data flow, component boundaries, and the Vue mental model. They are written to guide human decisions, not to duplicate every lint rule.
 
 - [See all priority A rules](./rules-essential)
 
-### Priority B: Strongly Recommended {#priority-b-strongly-recommended}
+### Priority B: Strongly Recommended (Legacy Reference) {#priority-b-strongly-recommended}
 
-These rules have been found to improve readability and/or developer experience in most projects. Your code will still run if you violate them, but violations should be rare and well-justified.
+This page is preserved for historical reference. Its recommendations may still be useful, but they are not part of the current maintained scope.
 
 - [See all priority B rules](./rules-strongly-recommended)
 
-### Priority C: Recommended {#priority-c-recommended}
+### Priority C: Recommended (Legacy Reference) {#priority-c-recommended}
 
-Where multiple, equally good options exist, an arbitrary choice can be made to ensure consistency. In these rules, we describe each acceptable option and suggest a default choice. That means you can feel free to make a different choice in your own codebase, as long as you're consistent and have a good reason. Please do have a good reason though! By adapting to the community standard, you will:
-
-1. Train your brain to more easily parse most of the community code you encounter
-2. Be able to copy and paste most community code examples without modification
-3. Often find new hires are already accustomed to your preferred coding style, at least in regards to Vue
+This page is also preserved as legacy reference content. It may help teams that want additional conventions, but it is not part of the current maintained scope.
 
 - [See all priority C rules](./rules-recommended)
 
-### Priority D: Use with Caution {#priority-d-use-with-caution}
+### Priority D: Use with Caution (Legacy Reference) {#priority-d-use-with-caution}
 
-Some features of Vue exist to accommodate rare edge cases or smoother migrations from a legacy code base. When overused however, they can make your code more difficult to maintain or even become a source of bugs. These rules shine a light on potentially risky features, describing when and why they should be avoided.
+This page remains available as historical guidance on risky patterns, but it is not part of the current maintained scope.
 
 - [See all priority D rules](./rules-use-with-caution)
--- a/src/style-guide/index.md
+++ b/src/style-guide/index.md
@@ -12,7 +12,9 @@ This is the official style guide for Vue-specific code. It is meant to help team
 
 This guide focuses on a small set of human-facing Vue rules. It is not intended to mirror every rule that lint tooling may enforce. Tools such as [eslint-plugin-vue](https://eslint.vuejs.org/), formatters, and other static analysis can still cover broader mechanical correctness and consistency checks.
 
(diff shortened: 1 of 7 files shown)


src/style-guide/rules-essential-ts.md
```diff
@@ -0,0 +1,614 @@
...
+
+In Vue 3.5+, you can destructure typed props directly from `defineProps()`.
+
+```ts
+const { status } = defineProps<{
+  status: string
+}>()
+
+watchEffect(() => {
+  console.log(foo)
```
ubugeeei: As a supplement, it might be worth mentioning somewhere that Language Tools can show hints when inlay hints are enabled.

Destructuring can look surprising because, from a JavaScript semantics perspective, it appears to break reactivity by disconnecting the reference. However, this is a special semantic provided by the Vue compiler, allowing default values to be written concisely. I’d like to help avoid confusion around that point.

[https://blog.vuejs.org/posts/vue-3-5#reactive-props-destructure](https://blog.vuejs.org/posts/vue-3-5#reactive-props-destructure)

I wrote an article about this in Japanese a while ago:

[https://zenn.dev/comm_vue_nuxt/articles/reactive-props-destructure#%E8%A8%80%E8%AA%9E%E3%83%84%E3%83%BC%E3%83%AB%E3%81%AE%E6%94%AF%E6%8F%B4%E3%81%AB%E3%81%A4%E3%81%84%E3%81%A6](https://zenn.dev/comm_vue_nuxt/articles/reactive-props-destructure#%E8%A8%80%E8%AA%9E%E3%83%84%E3%83%BC%E3%83%AB%E3%81%AE%E6%94%AF%E6%8F%B4%E3%81%AB%E3%81%A4%E3%81%84%E3%81%A6)

src/style-guide/rules-essential-ts.md
```diff
@@ -0,0 +1,614 @@
...
+
+In Vue 3.5+, you can use JavaScript's native default value syntax to declare default values for the props.
+
+```ts
+const { status = 'syncing' } = defineProps<{
+  status?: string
+}>()
+```
+
+In Vue 3.4 and earlier, use `withDefaults()` to declare default values for props.
```
ubugeeei, line 84: It does depend on the version, but in the Motivation section of the Props Destructure RFC written by Evan, it says:

https://github.com/vuejs/rfcs/discussions/502#discussion-5140019

> Succinct and native-like syntax for default values and local alias. Big DX improvement over withDefaults().

So I don’t think we need to intentionally mention `withDefaults` in the style guide.

It became stable in 3.5, but it should have been possible to opt in even before that.

src/style-guide/rules-essential-ts.md
```diff
@@ -0,0 +1,614 @@
...
+<script setup lang="ts">
+const emit = defineEmits<{
+  submit: [payload: { email: string }]
+  close: []
+}>()
+</script>
+
+<template>
+  <button @click="emit('submit', { email: 'john@example.com' })">Submit</button>
+  <button @click="emit('close')">Close</button>
```
ubugeeei: nits
```suggestion
  <button @click="emit('submit', { email: 'john@example.com' })">Submit</button>
  <button type="button" @click="emit('close')">Close</button>
```

src/style-guide/rules-essential-ts.md
```diff
@@ -0,0 +1,614 @@
...
+Usage in parent component:
+
+```vue [Parent.vue]
+<script setup lang="ts">
+const isOpen = ref(false)
+</script>
+
+<template>
+  <Child v-model="isOpen" />
+</template>
```
ubugeeei, line 229: nits

```suggestion
<script setup lang="ts">
const isOpened = ref(false)
</script>

<template>
  <Child v-model="isOpened" />
</template>
```

src/style-guide/rules-essential-ts.md
```diff
@@ -0,0 +1,614 @@
...
+
+<br />
+
+In Vue 3.5+, use [`useTemplateRef()`](https://vuejs.org/api/composition-api-helpers.html#usetemplateref) to access DOM or component instances.
+
+```vue
+<script setup lang="ts">
+import { useTemplateRef, onMounted } from 'vue'
+
+const el = useTemplateRef<HTMLInputElement>('my-input')
```
ubugeeei: Type inference should work in simple cases like this.
https://vuejs.org/guide/typescript/composition-api.html#typing-template-refs

```suggestion
const el = useTemplateRef('my-input')
```

src/style-guide/rules-essential-ts.md
```diff
@@ -0,0 +1,614 @@
...
+
+</div>
+
+<div class="style-example style-example-good">
+<h3>Good</h3>
+
+```ts
+const userId = ref(1)
+
+watch(userId, (id) => loadUser(id), { immediate: true })
```
ubugeeei, line 615: This is not the main point, but seeing `loadUser` may suggest something async.
Combined with `watch`, it could be interpreted as recommending code that may cause **race conditions**, so a more motivational example might be better. (For example, opening and closing a dialog)

Previously, there was actually an example like that in the official tutorial, and it caused a bit of discussion.

ubugeeei (changes requested):
I left a few comments for now, mainly about the content itself!

As for the overall structure of the documentation, I still have some doubts, so I’d like to discuss that on Discord instead. (My thoughts aren’t fully formed enough yet to leave comments.)

src/style-guide/rules-essential-ts.md
bencodezen: Agreed! We should remove this and just keep the 3.5+ mention in case people are wondering why it doesn't work in an older build (which they probably need to update rather than keep)

src/style-guide/index.md
```diff
@@ -6,36 +6,40 @@ outline: deep
...
 
-Finally, we've split rules into four categories:
+For the most part, we avoid suggestions about JavaScript or HTML in general. We will not set explicit guidelines around formatting decisions like semicolons, quote styles, etc. These are decisions that are personal to each team and we recommend setting your own conventions with tools like [Prettier](https://prettier.io/) or [Oxfmt](https://oxc.rs/docs/guide/usage/formatter.html).
+
+Our goal is to provide guidance around patterns that are important within the context of Vue.
+
+## How to use this guide {#how-to-use-this-guide}
+
+- Start with the [essential rules](./rules-essential). They are the primary maintained part of this guide and focus on component contracts, explicit data flow, component boundaries, and the Vue mental model.
+- Treat Priority B, C, and D as legacy reference pages. They may still contain useful conventions, but they are not part of the current maintained scope.
```
bencodezen, line 18:
```suggestion
```

The rest of the rules are still helpful guidelines for a lot of people in the community, so I don't think it's worth deprecating at this point. If needed, we should consider discussing them rule by rule to rather than deprecate them all at once.

src/style-guide/index.md
```diff
@@ -6,36 +6,40 @@ outline: deep
...
+- If your team uses [eslint-plugin-vue](https://eslint.vuejs.org/), expect tooling to enforce additional syntax, correctness, and consistency rules that are not all repeated here.
 
 ## Rule Categories {#rule-categories}
 
-### Priority A: Essential (Error Prevention) {#priority-a-essential-error-prevention}
+### Priority A: Essential (Maintained) {#priority-a-essential-error-prevention}
```
bencodezen, line 23:
```suggestion
### Priority A: Essential (Error Prevention) {#priority-a-essential-error-prevention}
```

I'd like to keep this theming to the style guides as a way to prevent developers from avoiding common mistakes that can cause errors in applications.

src/style-guide/index.md
```diff
@@ -6,36 +6,40 @@ outline: deep
...
 
-These rules help prevent errors, so learn and abide by them at all costs. Exceptions may exist, but should be very rare and only be made by those with expert knowledge of both JavaScript and Vue.
+These rules focus on component contracts, explicit data flow, component boundaries, and the Vue mental model. They are written to guide human decisions, not to duplicate every lint rule.
 
 - [See all priority A rules](./rules-essential)
 
-### Priority B: Strongly Recommended {#priority-b-strongly-recommended}
+### Priority B: Strongly Recommended (Legacy Reference) {#priority-b-strongly-recommended}
```
bencodezen, line 29:
```suggestion
### Priority B: Strongly Recommended {#priority-b-strongly-recommended}
```

Same idea. Let's not deprecate the entire list at this point. They're still useful for the community.

src/style-guide/index.md
```diff
@@ -6,36 +6,40 @@ outline: deep
...
 
-These rules have been found to improve readability and/or developer experience in most projects. Your code will still run if you violate them, but violations should be rare and well-justified.
+This page is preserved for historical reference. Its recommendations may still be useful, but they are not part of the current maintained scope.
 
 - [See all priority B rules](./rules-strongly-recommended)
 
-### Priority C: Recommended {#priority-c-recommended}
-
-Where multiple, equally good options exist, an arbitrary choice can be made to ensure consistency. In these rules, we describe each acceptable option and suggest a default choice. That means you can feel free to make a different choice in your own codebase, as long as you're consistent and have a good reason. Please do have a good reason though! By adapting to the community standard, you will:
+### Priority C: Recommended (Legacy Reference) {#priority-c-recommended}
```
bencodezen, line 35:
```suggestion
### Priority C: Recommended {#priority-c-recommended}
```

bencodezen (changes requested):
@serkodev Thanks for your work on this. If possible. I'd like to focus the PR on the typescript guide you wrote rather than deprecate the other rules at the moment.

There's still a lot of value in them for the community and the verbiage on the style-guide/index.md has had a lot of user testing so I'd prefer not to rewrite too much of it at this point. 

Let me know if you have any questions or want to discuss further in Discord!

src/style-guide/index.md
```diff
@@ -6,36 +6,40 @@ outline: deep
 
 This is the official style guide for Vue-specific code. If you use Vue in a project, it's a great reference to avoid errors, bikeshedding, and anti-patterns. However, we don't believe that any style guide is ideal for all teams or projects, so mindful deviations are encouraged based on past experience, the surrounding tech stack, and personal values.
```
bencodezen, line 7: As a general comment for the page, I'd like to avoid rewriting this one in terms of the guidance. There's a lot of value in helping guide developers towards developing their own style guide and I don't think that those rules have changed so drastically they need to be made legacy at this point. 

I'm happy to keep an eye on them to help maintain them, so hopefully that'll address any concerns there.

src/style-guide/rules-strongly-recommended.md
```diff
@@ -1,7 +1,43 @@
-# Priority B Rules: Strongly Recommended {#priority-b-rules-strongly-recommended}
+# Priority B Rules: Strongly Recommended (Legacy Reference) {#priority-b-rules-strongly-recommended}
 
 These rules have been found to improve readability and/or developer experience in most projects. Your code will still run if you violate them, but violations should be rare and well-justified.
 
+## Use multi-word component names {#use-multi-word-component-names}
```
bencodezen, line 5: While this wasn't a rule for Vue specifically, this is something we have seen cause errors on code bases where people would use single word component names and run into naming collisions with other libraries or even the HTML spec itself, so I think it still makes sense to keep it in the error preventing level in essentials.

PR #3404, opened 2026-06-08
fix: remove duplicate border-radius in VueSchoolLink component

Aaron-zon:
## Description of Problem

the .vueschool style block contains duplicate border-radius declarations. Since border-radius: 8px overrides the earlier border-radius: 2px, the previous declaration is ineffective.

## Proposed Solution

Remove the redundant `border-radius: 2px` and keep only `border-radius: 8px`.


bencodezen (approved):
Good catch. Thanks @Aaron-zon!

PR #3414, opened 2026-06-22
Fix: Move deutsch from progress to available lang

tisma95:
## Additional Information
This pull request makes a minor update to the translations list by removing the German translation entry from the `src/translations/index.md` file.

* Removed the German translation (`Deutsch / German`) and its source link from the list of available translations.


AloisSeckar:
Good catch, but this was already opened in #3412

tisma95:
> Good catch, but this was already opened in #3412

Thanks @AloisSeckar  we can close this.

PR #3423, opened 2026-06-30
docs: clarify `.once` listener attachment semantics

### LINKED ISSUE #3033: `v-on` modifier `.once` triggers handler at most once... per what?
https://github.com/vuejs/docs/blob/a961225c9dbfde52c510ce19549a74eb34993bea/src/api/built-in-directives.md?plain=1#L187

I found the statement above unclear, because this could imply the listener would be called once per: 

- an event,
- a mount,
- an update or
- a span in lifecycle (possibly specified somewhere).

This could lead readers to assume the semantics or simply avoid using this modifier not to cause confusion.

happysnaker:
## Summary
- clarify that `.once` applies once per listener attachment
- explain that the listener is removed after the first invocation
- note that recreating the element or child component attaches a fresh listener

Closes #3033

### DIFF
--- a/src/api/built-in-directives.md
+++ b/src/api/built-in-directives.md
@@ -184,7 +184,7 @@ Attach an event listener to the element.
   - `.capture` - add event listener in capture mode.
   - `.self` - only trigger handler if event was dispatched from this element.
   - `.{keyAlias}` - only trigger handler on certain keys.
-  - `.once` - trigger handler at most once.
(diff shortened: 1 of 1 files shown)


happysnaker:
Resubmitted from a restored public fork as #3425. Closing this thread in favor of the new PR so the contribution stays reviewable from a proper fork relationship.

PR #3425, opened 2026-06-30
docs: clarify `.once` listener attachment semantics

### LINKED ISSUE #3033: `v-on` modifier `.once` triggers handler at most once... per what?

happysnaker:
## Summary
- clarify that `.once` applies once per listener attachment
- explain that the listener is removed after the first invocation
- note that recreating the element or child component attaches a fresh listener

Closes #3033

## Note
This PR supersedes #3423, which I had to close after the original fork relationship was broken during repo visibility changes on my side. The content is being resubmitted unchanged from a proper public fork.

### DIFF
--- a/src/api/built-in-directives.md
+++ b/src/api/built-in-directives.md
@@ -184,7 +184,7 @@ Attach an event listener to the element.
   - `.capture` - add event listener in capture mode.
   - `.self` - only trigger handler if event was dispatched from this element.
   - `.{keyAlias}` - only trigger handler on certain keys.
-  - `.once` - trigger handler at most once.
+  - `.once` - trigger handler at most once for each listener attachment.
   - `.left` - only trigger handler for left button mouse events.
   - `.right` - only trigger handler for right button mouse events.
   - `.middle` - only trigger handler for middle button mouse events.
@@ -199,6 +199,8 @@ Attach an event listener to the element.
   When listening to native DOM events, the method receives the native event as the only argument. If using inline statement, the statement has access to the special `$event` property: `v-on:click="handle('ok', $event)"`.
 
   `v-on` also supports binding to an object of event / listener pairs without an argument. Note when using the object syntax, it does not support any modifiers.
+
+  The `.once` modifier removes the listener after its first invocation. If the element or child component is re-created later, a new listener is attached, so it can run once again.
 
 - **Example**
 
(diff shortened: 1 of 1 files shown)


src/api/built-in-directives.md
```diff
@@ -184,7 +184,7 @@ Attach an event listener to the element.
   - `.capture` - add event listener in capture mode.
   - `.self` - only trigger handler if event was dispatched from this element.
   - `.{keyAlias}` - only trigger handler on certain keys.
-  - `.once` - trigger handler at most once.
+  - `.once` - trigger handler at most once for each listener attachment.
```
bencodezen, line 187: "listener attachment" is not very clear to me. What about:


```suggestion
  - `.once` - trigger handler at most once for each event listener.
```

src/api/built-in-directives.md
```diff
@@ -200,6 +200,8 @@ Attach an event listener to the element.
 
   `v-on` also supports binding to an object of event / listener pairs without an argument. Note when using the object syntax, it does not support any modifiers.
 
+  The `.once` modifier removes the listener after its first invocation. If the element or child component is re-created later, a new listener is attached, so it can run once again.
```
bencodezen, line 203: This description is slightly inaccurate given `.once` has different behavior for native DOM elements vs components.

Rather than a dedicated paragraph here though, maybe it makes more sense for it to be described more in depth in the list?

See https://github.com/vuejs/docs/pull/3425/changes#r3521256033

src/api/built-in-directives.md
bencodezen, line 187:
```suggestion
  - `.once` - trigger handler at most once for each event listener. The event is removed in native DOM events and ignored (rather than removed) in a component emitted event instance.
```

bencodezen (changes requested):
Thanks for the PR @happysnaker! I left some comments on the PR. Let me know what you think or if you have any questions!

PR #3426, opened 2026-07-02
docs: clarify scoped CSS selector performance

ishaanlabs-gg:
## Description of Problem

The scoped CSS tips currently say an element selector such as `p { color: red }` will be "many times slower" when scoped. That warning is directionally useful, but it reads more absolute than current browser behavior warrants and does not tell readers when they should actually care.

Addresses vuejs/core#2608.

## Proposed Solution

This updates the tip to explain what scoped CSS compiles to, notes that selector cost depends on the browser, selector, and affected DOM size, and keeps the practical guidance to prefer classes or ids in large or frequently updated component trees.

## Additional Information

Validation:

- `git diff --check HEAD~1 HEAD`

This is a docs-only wording change, so I did not run the full VitePress site build.

### DIFF
--- a/src/api/sfc-css-features.md
+++ b/src/api/sfc-css-features.md
@@ -98,7 +98,7 @@ You can also include both scoped and non-scoped styles in the same component:
 
 ### Scoped Style Tips {#scoped-style-tips}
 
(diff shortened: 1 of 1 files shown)


bencodezen (changes requested):
I like the clarity around the browser impact to be less absolute, but I do think this rewrite hides the original message that using generic HTML selectors as scoped styles being slower performance wise. 

In addition, the scoped styles is explained earlier in the section, so we should avoid re-explaining it in the tip.

PR #3427, opened 2026-07-06
docs: clarify scoped slot examples

ubugeeei:
## Motivation

I have some feedback from my friend about scoped slot documentation.

Some readers find the scoped slots examples difficult to follow because the parent/child relationship and the usage/definition sides are implicit. 
This update makes the example flow more explicit by showing the parent component usage first, then the child component slot definition, matching the order used in the JavaScript function analogy below.

## Changes

- Show the parent component usage before the child slot definition in the scoped slots guide (align to JS image)
- Rename the received slot props binding to `receivedProps`
- Remove implicit local state from scoped slot examples
- Update the scoped slots diagram and Playground links to match the new examples

---

Note: `corepack pnpm type` currently fails on an existing `.vitepress/config.ts:795` plugin type cast issue unrelated to this change.


bencodezen (approved):
This is great. Thanks for this improvement @ubugeeei!

AloisSeckar:
A bit too late, but a question:

On line 375 of `src/guide/components/slots.md` there is a comment with link to Figma source for the image. It was added so local translations can possibly fork and make their own translated versions.

The link is for the old image. So my question is - would it be possible to make the new one also available via public Figma link? @ubugeeei ?

PR #3442, opened 2026-07-29
docs: fix broken playground links in scoped slots examples

### LINKED ISSUE #3441: Bug: `Try it in the Playground` links throw "Failed to load code from URL"
The two `Try in the Playground` links in the "Slots" article fail with a `Failed to load code from URL` alert when opened (see links and attached screenshot below).

https://github.com/vuejs/docs/blob/7681134fd8505e61a265d161d73d28acb3c74822/src/guide/components/slots.md?plain=1#L377-L386

Image

Mohith26:
Fixes #3441, reported by @illia-m-b.

The two "Try it in the Playground" links added to the scoped-slots section in #3427 embed hashes compressed with default-level zlib (`eJx...`). The playground's decoder (`atou` in @vue/repl) only recognizes level-9 zlib streams (`eNo/eNp/eNq/eNr`) and otherwise falls through to the legacy plain-base64 path, producing garbage that fails `JSON.parse`, hence "Failed to load code from URL". All 108 other playground links in the docs already use the level-9 format; these two were the only outliers (verified by a repo-wide prefix scan).

The fix re-encodes the byte-identical example content with level-9 zlib. Verified three ways: an exact Node port of the playground's decoder rejects the old hashes and decodes the new ones to byte-identical content, and both new URLs load the correct scoped-slot examples live on play.vuejs.org. Two-line diff, content unchanged.


bencodezen (approved):
Thanks for the PR @Mohith26! This fixes it!

PR #3446, opened 2026-07-31
docs: add short note about using multiple statements in inline handlers

whysopaul:
## Description of Problem

Vue allows passing multiple statements as inline handlers in v-on by separating them with a semicolon, but this isn't mentioned in the inline handlers [docs](https://vuejs.org/guide/essentials/event-handling.html#inline-handlers).

## Proposed Solution

Added a short note about using a semicolon to separate multiple statements in inline handlers.


NataliaTepluhina:
@whysopaul thank you for the proposed change! Ideally, we would encourage users to create a method for this instead of separating calls with the semicolon

PR #649, opened 2020-10-24
feat(migration): transition-group (close #648)

LinusBorg:
`<transition-group>` no longer renders a root element by default, since Vue 3 now supports fragments.

PR #3455, opened 2026-08-13
docs: clarify TransitionGroup wrapper element difference from Transition

### LINKED ISSUE #3454: Guide: TransitionGroup page falsely implies Transition renders a wrapper element
The ["TransitionGroup" page of the guide](https://vuejs.org/guide/built-ins/transition-group.html#differences-from-transition) includes the following text:

> ## Differences from `<Transition>`
>
> `<TransitionGroup>` supports the same props, CSS transition classes, and JavaScript hook listeners as `<Transition>`, with the following differences:
>
> - By default, it doesn't render a wrapper element. But you can specify an element to be rendered with the `tag` prop.
> - [...]

This implies that `<Transition>` _does_ render a wrapper element. But actually, in Vue 3, it appears that neither `<Transition>` nor `<TransitionGroup>` renders a wrapper element by default, so this isn't really a "difference".

(This language seems to have been introduced by #649.)

I'm new to Vue, so maybe there's something I'm not understanding here, but it seems like this should maybe be revised to say something like:

> - A `tag` prop can be passed to render a wrapper element. (By default, `<TransitionGroup>` doesn't render a wrapper element, similar to `<Transition>`.)

lazerg:
## Description of Problem

The "Differences from `<Transition>`" section of the TransitionGroup guide lists "By default, it doesn't render a wrapper element" as a difference, which reads as though `<Transition>` does render one. It doesn't. In Vue 3, `<TransitionGroup>` renders a fragment unless the `tag` prop is set, and `<Transition>` renders its single child directly.

## Proposed Solution

Reword the bullet so the actual difference is the `tag` prop, and state that neither component renders a wrapper by default.

## Additional Information

The wording comes from #649, which corrected the Vue 2 behavior (a `<span>` root) but left the bullet in the differences list.

Closes #3454


bencodezen (commented):
Thanks for the PR @lazerg! This is a simple but effective way to reduce confusion on the topic.

bencodezen (approved):
Thanks for the PR @lazerg! This is a simple but effective way to reduce confusion on the topic.

